PR #195, opened 2013-12-07
add  i18n chinese support

cnzin:
add  chinese and remove a duplicate function


SachaG:

PR #198, opened 2013-12-09
Add Spanish i18n

kevohagan:
Just a little pull request adding spanish (spain) translation for Telescope in es.js file


SachaG:

PR #205, opened 2013-12-18
Spanish revised

donflopez:
Some changes for improve spanish language.

Good job with Telescope, thanks!


SachaG:

Issue #248, opened 2014-02-11
Feature request: ability to downvote


multani7871:
Similar to Reddit

sth3000:
+1 for downvote option in settings

PR #315, opened 2014-07-06
Feature/downvote

jcsaaddupuy:
This enable the ability to downvote posts by adding a second button on the left of the post.
When downvoted, the icon icon-check will be displayed in red.

However, there is no setting as mentioned in https://github.com/TelescopeJS/Telescope/issues/248

### DIFF
--- a/client/sass/modules/_posts.scss
+++ b/client/sass/modules/_posts.scss
@@ -49,7 +49,6 @@
 		.upvote-link, .downvote-link{
 			height:50%;
 			display:table;
-			float:left;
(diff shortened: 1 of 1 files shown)


naping:
Hi @SachaG I just wanted to follow up to see if down vote will be supported in the near future. Thanks for everything!

SachaG:
It is supported, there's just no button for it in the default theme. If you fork the theme you can easily duplicate the upvoting code to create a downvote button. I might make it a user-defined setting in the future.

PR #458, opened 2014-10-03
German translation (de.js)

Helmi:
Some notes: there should be an option to use more than 2 characters for translation. For german you would at least need de_formal and de_informal. I did the informal translation as i found it more useful for myself and i think for most users.

Also i decided to translate "Posts" into "Links" as generally this are links and "Posts" doesn't have a german word that is general enough to not translate more into "Articles" which definitely are not the right phrase to use.

Also i found some not yet translated features but as i translated based on the actual master and not on a release i guess this is a known Problem.


SachaG:

PR #502, opened 2014-10-26
tr.js

tauslu:
This is the Turkish translation, I used German one as basis so on the first line, I changed i18n.translations.de to i18n.translations.tr and translated everything else


SachaG:

PR #509, opened 2014-11-04
Invite bugfixes

boingy:
1. Inviting a user via their profile page did not work because the call
to ‘inviteUser’ did not include the userId key. Also it used to
previously show 'Thanks, user has been invited.' regardless of success or
failure.
2. Meteor.users.update( ... ) in inviteUser method had a
double $inc modifier meaning that inviteCount would never actually be
decreased.
3. Added a uninviteUser method. Previously uninviting a user from the
all-users admin page would just update that user’s isInvited flag to
false. This was a problem because if that user was invited again
it wouldn’t work because a invite would still exist for them in the
Invites collection. This new method removes them both.

### DIFF
--- a/client/views/users/user_item.js
+++ b/client/views/users/user_item.js
@@ -36,10 +36,11 @@ Template[getTemplate('user_item')].events({
   },
   'click .uninvite-link': function(e, instance){
     e.preventDefault();
-    Meteor.users.update(instance.data._id,{
-      $set:{
-        isInvited: false
-      }
+    var userId = instance.data._id;
+    var userEmail = getEmail(instance.data);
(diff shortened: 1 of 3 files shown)


xavcz:
Hey, folks! Telescope has changed a lot in the past months, and there is a new React-based version, [Telescope Nova](https://github.com/TelescopeJS/Telescope) ([intro video](https://www.youtube.com/watch?v=_cLv6fH-PSQ&index=1&list=PLBoa_Q6hVeSzChsSi8X8x3FQ8NN2D73fB)). If you know Crater, it runs on the latest version.

The Legacy version is no longer supported. I'm closing this PR.

Happy hacking! 🚀

PR #620, opened 2014-12-12
Update SEO package for master, remove page titles

yourcelf:
Continuation from https://github.com/TelescopeJS/Telescope/pull/521 which catches up to latest master and removes title-setting as a responsibility of the package.


yourcelf:
I've updated this for latest master and simplified as discussed above.  I've left the ``seoOgImage`` field intact for the reasons @john-henry mentioned.

SachaG:
Awesome, merged on `devel`. The only thing I changed is renaming the `seoOgImage` to `siteImage` to make it more generic. 

One thing though, when I tested it it seemed like post titles didn't show up for some reason. Can you double check that?

Issue #452, opened 2014-09-30
[feature] Allow users to follow some posts


arunoda:
Hi,

I think it's great to follow some posts, without commenting.
Right now, only people who've commented would be able to get emails. (this is true for subcomments as well)

I really like to have a post wide feature, allow  users to follow the post and get notifications.

PR #650, opened 2014-12-21
Subscribe post

delgermurun:
Ref #452.

- Subscribe/Unsubscribe post
- When new comment added on user's subscribed posts, user receive notification.
- Show subscribed posts on user's profile
- Author can't subscribe own post (already sends notification)
- Change new comment's email notification subject, if post is subscribed post (not own post)
- Save subscriber count on post.

### DIFF
--- a/lib/publications.js
+++ b/lib/publications.js
@@ -22,6 +22,7 @@ privacyOptions = { // true means exposed
   'votes.downvotedPosts': true,
   'votes.upvotedComments': true,
   'votes.upvotedPosts': true,
+  'subscribedItems.Posts': true
 };
  
 // minimum required properties to display avatars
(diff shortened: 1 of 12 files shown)


SachaG:
I'm working on this. I might end up refactoring it a little bit :)

delgermurun:
:+1:

SachaG:
You can check out the results on the `devel` branch: https://github.com/TelescopeJS/Telescope/tree/devel

Main things I changed: 

- Rewrote the logic to decide which users to notify to be less confusing.
- Made `telescope-subscribe-to-posts` into its own package independent from the notification stuff. 
- A user's subscribed posts are now private and displayed on their user edit page, not their public user profile.
- The subscribe link is now at the bottom of the comment thread, not in the main post item.

Please review my changes and let me know what you think :)

I'm still testing out some stuff by the way, so it's probably not 100% bug-free yet…

delgermurun:
@SachaG notification says "has replied to your comment on", but I don't reply, I just write root comment.

PR #662, opened 2014-12-30
Adding nav client unit test

anthonymayer:
Using the jasmine package seemed like the best way to go because we'll want most of the testing modes. I added the travis file, but it doesn't quite work yet (see https://travis-ci.org/anthonymayer/Telescope). I don't think any other changes should be required of the travis file right now though because I'm controlling it all through https://github.com/anthonymayer/travis-ci-meteor-packages. Once I get that working I'll probably add a tag in the README too.

Mostly I'm looking for feedback on testing structure. I'm hoping that after we have a few more examples of these that we start making adding/updating tests a requirement of push requests. Getting feedback from either @sanjo and/or @samhatoum would be awesome too.

I also deleted some nav stuff that looked unused.

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -80,7 +80,6 @@ telescope-post-by-feed
 # Testing
 
 sanjo:jasmine
-velocity:core
 velocity:html-reporter
 
 # Custom Packages
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -77,4 +77,10 @@ telescope-singleday
 telescope-invites
 telescope-post-by-feed
 
+# Testing
+
+sanjo:jasmine
+velocity:core
+velocity:html-reporter
+
 # Custom Packages
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,0 @@
-language: node_js
-node_js:
-  - "0.10"
-before_install:
-  - "curl -L http://git.io/ejPSng | /bin/sh"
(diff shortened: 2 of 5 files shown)


SachaG:
Thanks! Just a heads up, there might be a few small tweaks to the navigation template in the next update.

anthonymayer:
Okay, well updating the tests should be pretty easy. Let me know if you need help.

SachaG:
I ended up disabling the testing packages in the latest release because I didn't have the time to get the tests to pass again. I also want to write some documentation about testing, if only to point people to the right place to get more information. Let's do all that for the next official release :)

anthonymayer:
Okay I can work on some documentation. Can you fix the tests? If not I can this time but it would be good if you did because writing/updating tests should be part of the development process. The tests were just failing, right, or were you having issues with the test runner itself? As soon as the todos on https://github.com/Sanjo/meteor-jasmine/pull/118 are done I plan on enabling travis for Telescope and probably adding a badge to the README so that people know the state of the tests. I'll make sure it's mentioned in the docs too. Also we shouldn't disable velocity if a test is failing, we can disable individual tests.

SachaG:
I took a look at the tests, but I'm not familiar with Velocity at all so I
wasn't able to fix them. For the future I agree though, I'll fix them as I
add new features.

PR #668, opened 2015-01-02
git ignoring .meteor/versions

anthonymayer:
@SachaG: What do you think of this? This file just keeps driving me insane. It seems like if we want to pin versions we should do it in .meteor/packages. I'm happy to do that too if you think it will be stabler for all packages. We shouldn't have 2 places to do this though. Also, if you're cool with this, I'll probably do the same thing with all the packages' versions.json files.

Also threw in random .editorconfig cleanup


SachaG:
I'm not sure, I thought the point of the `.versions` file was precisely to make sure everybody is using the same versions? I do agree that it's a little buggy right now. 

But how exactly is it driving you insane? I pretty much just ignore it except when I need to force a package to update and `meteor update` somehow doesn't work.

anthonymayer:
Yea, I guess, but it always auto bumps package versions anyway so it doesn't really seem to guarantee that everything's the same. It's annoying both when I do "git diff" and "git add -p" since it's always a lot of changes. It's also not really clear when/if we should be committing changes to it. If we moved to only pinning versions in `.meteor/packages` it seems like it would make it more consistent. I might just be completely misunderstanding how it's supposed to be used though.

anthonymayer:
Anymore opinions about this or any major objections to me merging it?

SachaG:
I don't think we should merge it. Let's ask about this on Crater first.

anthonymayer:
Posted on Crater: http://crater.io/posts/A3DFNuMAYdM99sAyP

Slava:
No, ignoring `versions` is not OK. The whole point of this file is to be able to reproduce the build as `versions` pins every package selected by the constraint solver for package constraints described in `packages` file. So the idea is to be able to reliably get the same build on any machine at any point of time past the initial build.

Every time you add or remove a package, the set of constraints changes, thus the solution can be different.

If you think the feature works incorrectly, please, report a bug with a reproduction to meteor/meteor GitHub issues.

Expect the relation to be the same as `Gemfile` and `Gemfile.lock` in ruby-land.

anthonymayer:
@Slava: So that means anytime I see updates in my `.meteor/versions` file I should be committing them? Wouldn't pinning every version in `packages` do the same thing? Sorry, not trying to come off dumb here. I really want to understand this and make sure I'm using it correctly.

Slava:
@anthonymayer as I explained above: packages file is only for defining constraints. For example `iron:router@0.9.3` doesn't pin the version to `0.9.3`. It means "give me 0.9.3 or anything that is higher and compatible in the semver sense, but prefer the latest version". There is an array of possible versions to pick depending on other packages.

anthonymayer:
@Slava: Ahh ok. Thanks. I thought it actually pins it. That was the big disconnect. Makes sense now. I also just found https://www.meteor.com/version-solver which clears it up a lot.

SachaG:
Glad we could figure it out, thanks again @Slava :)

PR #669, opened 2015-01-03
Add Bulgarian translation

toome123:
After changing i18n The structure, I had a new rewrite the new JSON format translation.


SachaG:

toome123:
I will make translation soon .. When i have some time :)

PR #672, opened 2015-01-05
Getting rid of redundant permissions functions.

anthonymayer:
Switched everything over to use can.* functions from telescope-lib instead of using can* functions from lib/permissions.js, deleted lib/permissions.js, added some tests for permissions, and some other random cleanup like deleting unused code.


SachaG:
I'm going to take a look, but I think it's best to keep the testing stuff in their own PRs for now. As things stand I wouldn't be able to answer any questions about testing or velocity, so I don't want to introduce it until I'm prepared to deal with that :)

client/helpers/handlebars.js
```diff
@@ -23,19 +23,13 @@ UI.registerHelper('isLoggedIn', function() {
...
+  return can.post(Meteor.user());
 });
 UI.registerHelper('canComment', function() {
-  return canComment(Meteor.user());
-});
-UI.registerHelper('canUpvote', function(collection) {
-  return canUpvote(Meteor.user(), collection);
-});
-UI.registerHelper('canDownvote', function(collection) {
-  return canDownvote(Meteor.user(), collection);
```
SachaG: Why did you delete this? Was it not being used?

SachaG:
To clarify about testing, it's not that I mind having it, it's just I'm worried that having the test runner automatically launch on startup will be confusing for users if we don't explain how it works beforehand. But if there's a way to make the test runner optional or manually triggered, that would work for me too. 

But in any case great work, a much needed clean-up. On the topic of permissions, I think we should keep moving away from role-based permissions (`isAdmin`) towards action-based permissions (`can.edit`). This will make it a lot easier to add new roles like moderators in the future. 

And by the way, I did also look into the meteor-roles package, but I couldn't really identify any areas where it would really make a difference. It seems more for giving permissions to individual users rather than defining broad roles like admin or mods.

client/helpers/handlebars.js
anthonymayer: Yea, as far as I could tell they were unused

anthonymayer:
Yea, moving to action-based permissions definitely seems more flexible. We might even end up with a permissions package. If we go that route we should probably even try to make it more general purpose than just Telescope if we can and just configured for Telescope. In this case I was mostly just working on deleting duplicate code. I've been playing around with https://codeclimate.com/github/anthonymayer/Telescope and found it.

As far as the testing goes, I turned it off again, because it doesn't look like there's a way (or at least not an obviously documented one) of running the tests without showing the test reporter in the app and it would still pop up an instance of Chrome that it actually runs the test in. The placement of the test reporter is also particularly poor for Telescope, because it blocks the "Post" button.  I totally agree that it's jarring for developers to suddenly see this and not know that it only shows up during development or what to do with it. I started working on [docs about how automated testing works in Telescope](https://github.com/TelescopeJS/Telescope/wiki/Testing) but I still have a fair bit to add.

The biggest problem is that we can't really start having more people write tests or having more tests until we have it turned on for everybody since the tests will otherwise rot and not do their job. Once travis is setup it will be less bad, since it should work for both pull requests and branches so we'll at least know after someone is done developing. Maybe that's enough for now to get awareness up. We can also start asking more developers to turn on testing until we have a stable group of developers who understand it. In the meantime I'll keep adding some tests. I definitely want to make sure we've got some more basic things covered. It will also give people more examples to look at.

We'll definitely want to have a way to move where the test reporter button pops up, which I'll look into on https://github.com/meteor-velocity/html-reporter. I might even try to get a help button added to it so that users can more easily figure out what's going on with it when they first see it. @sanjo, @samhatoum: if either of you have any thoughts on this problem, please chime in :)

Sanjo:
`VELOCITY=0 meteor` should disable the test running. But currently the html-reporter still displays with this set.

SachaG:
@sanjo @samhatoum Well I think Telescope is a bit of a special case, because the vast majority of end users should not be concerned with running tests. After all, you don't see tests run every time you launch a WordPress blog, even if you're running it locally over Apache. 

So I do think an option to *not* show tests by default and only enable them with `VELOCITY=1 meteor` is needed in our case. Or else we need to figure out how to have a different branch for tests, but I can foresee that getting pretty messy…

samhatoum:
all the test packages are `debugOnly` which means they'll only show up in development. The `debugOnly` flag means Velocity does not get bundled. 

As for testing without the reporter, it is possible, and it's also possible to stop jasmine from displaying the chrome window (I think @Sanjo is working on making phantomjs the default unit-testing browser).

Perhaps you can have a script that starts telescope in testing mode, where the script adds a line to the packages file and then starts meteor. This means you have to consciously want to run tests locally

SachaG:
@samhatoum we could do that. Although I always hate messing with the packages file, since that produces git conflicts. 

So a config option to not show the chrome window and the html reporter would probably still be best, if that's in the works?

samhatoum:
Eventually :)

anthonymayer:
Here's a temporary compromise https://github.com/TelescopeJS/Telescope/pull/706. Basically, we'll get Travis integration to at least make sure people aren't breaking tests and then encourage people to enable the testing packages during development. Should move us in the right direction.

samhatoum:
Good work!

PR #677, opened 2015-01-06
Update helpers.js

comerc:


lib/helpers.js
```diff
@@ -1,5 +1,5 @@
 cl = function(something){
```
anthonymayer: is this function actually used anywhere? as far as I can tell it's not and we should probably just delete it since it doesn't really seem worth having anyway. @comerc: Do you feel like updating this pr to just delete the function?

SachaG: Well I had this idea of a general log helper that you could toggle on and off through a `debug` flag, to make it easier to debug issues. But I don't know if that's a good idea, and anyway you're right that it's not used. So it might be best to remove it.

PR #678, opened 2015-01-06
fixed when is not possible get the description

Kikobeats:


SachaG:
Thanks! I actually ended up doing this in a slightly different way:

https://github.com/TelescopeJS/Telescope/commit/65f30864c6e457c379b1497a265bb541b4732493

PR #700, opened 2015-01-15
Humanize URL

Kikobeats:
The purpose of this PR is:

- Added global helper that ensure that a URL is correctly resolved in the worst of the case (the user doesn't put the protocol in the url).
- Added a global helper for delete the protocol for the URL (for better UX).
- Applied the two helpers in the profile URL of the user.


Kikobeats:
wtf digestController. I'm going to delete this commit!

Kikobeats:
arg, little problem with the session key.... any idea?

xavcz:
Hey @Kikobeats, thanks for this cool PR but we can't merge it because of Nova now. I hope you'll be able to give a try to Nova! 💯

PR #705, opened 2015-01-19
Making both Travis and CodeClimate integrations works

anthonymayer:


anthonymayer:
Discarding and redoing this on top of devel

PR #706, opened 2015-01-19
Making both Travis and CodeClimate integrations works

anthonymayer:


anthonymayer:
@SachaG: Any objections to the README badges?

SachaG:
Not at all :)

PR #712, opened 2015-01-20
Adding newsletter time setting

anthonymayer:
Adding newsletterTime setting.

Still todo
- [ ] Add tests
- [ ] Add some more translations

### DIFF
--- a/packages/telescope-newsletter/i18n/en.i18n.json
+++ b/packages/telescope-newsletter/i18n/en.i18n.json
@@ -9,6 +9,7 @@
   "mailChimpListId": "MailChimp List ID",
   "postsPerNewsletter": "Posts per Newsletter",
   "newsletterFrequency": "Newsletter Frequency",
+  "newsletterTime": "Newsletter Time",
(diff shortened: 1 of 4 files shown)


SachaG:
Should I merge it now, or wait until you add some tests?

anthonymayer:
If you think it's good you're more than welcome to merge it and I can add tests as a follow up. Part of the reason I wanted to add tests is that I don't actually use the newsletter so I've only tested the setting part manually, I don't know the best way to test that the newsletter actually gets sent.

SachaG:
Does the time use 24-hour or AM/PM notation? Also, can we make the default time 09:00 AM?

anthonymayer:
@SachaG: Changing the default to 9:00 AM isn't a problem. I used 00:00 because that's what it looked like the default is currently. The time notation is actually based on the user's browser settings, because it uses the HTML5 time input tag. It looks like it always returns 24 hr values though no matter what the user sees.

SachaG:
OK, sounds good :)

PR #716, opened 2015-01-21
Update _posts.scss

klamzo:
Added a "justify-content: center" so that no customization is needed to adjust the posts display for breakpoints.


SachaG:
Awesome :)

SachaG:
Actually I had to revert this because it centered all grid elements, which is not really what I wanted. But I did fix the mobile version :)

klamzo:
no problem.

PR #722, opened 2015-01-25
change sign in for register

Kikobeats:


i18n/en.i18n.json
```diff
@@ -106,8 +106,8 @@
 
   //Errors
   "you_are_already_logged_in": "You are already logged in",
-  "sorry_this_is_a_private_site_please_sign_up_first": "Sorry, this is a private site. Please sign up first.",
-  "thanks_for_signing_up": "Thanks for signing up!",
+  "sorry_this_is_a_private_site_please_sign_up_first": "Sorry, this is a private site. Please register first.",
+  "thanks_for_signing_up": "Thanks for register!",
```
splendido: should this be *registering*?

Kikobeats:
@splendido updated!

splendido:
tnx!

i18n/en.i18n.json
```diff
@@ -11,7 +11,7 @@
   "admin": "Admin",
   "post": "Post",
   "toolbox": "Toolbox",
-  "sign_up_sign_in": "Sign Up/Sign In",
+  "sign_up_sign_in": "Register/Sign In",
```
anthonymayer: we don't really have precedent for this as far as I'm aware, but would it be worth opening an issue to update the translations for this as well? It probably doesn't matter too much, but it seems like it would be good to be consistent

splendido: Yeah, you're right, it makes sense.
...althought the italian, i.e., is already using register from the beginning (there's no *sign up* equivalent...)

PR #724, opened 2015-01-26
Fixes #719. Allowing mobile nav to close if user clicks anywhere outside of it

### LINKED ISSUE #719: Mobile menu should close if you click anywhere outside the mobile menu
Should work if you try to swipe too

anthonymayer:
Also, adding test.


anthonymayer:
Uhh looks like there's some weirdness to how width is measured depending on how the test is run. This might be a velocity bug. I'll update this when the tests actually pass. Yay Travis catching test failures!

PR #726, opened 2015-01-27
Changed Sign-up to Register

kardotim:
When there's account set up yet, the navigation says Register instead of Sign-up. 

screen shot 2015-01-27 at 15 49 35

I've just updated the 'Read this First' steps accordingly.


splendido:
Tnx!

Issue #723, opened 2015-01-25
Whitespace not being trimmed from email/username when signing in


anthonymayer:
I thought this might have to do with http://github.com/meteor-useraccounts/core, but I tried on one of their demo sites and didn't run into the issue. I'll do some more digging to see where it's coming from.

splendido:
This was somehow discussed in #691 and is actually in phase of implementation on the [lowercase-usernames](https://github.com/TelescopeJS/Telescope/tree/lowercase-usernames) branch

splendido:
this should be fixed now since @SachaG [merged](https://github.com/TelescopeJS/Telescope/commit/31dffc58fb2e3085a703668a70c8aa51cb16334e) `lowercase-usernames` into `devel`

Issue #691, opened 2015-01-11
lowerCase/UpperCase in login


Kikobeats:
I detect that if you are registered with `kikobeats` and try to make login with `Kikobeats` the login basically failed. 

So I want to improve the login experience. For example, in mobile is very common auto-capitalize the first letter. Some popular services as Facebook use a effective UX login experience:

```
Password: aBcDeF123
Works:    AbCdEf123  // Flipped Case
Works:    ABcDeF123  // Caps First Char
Doesnt:   ABCDEF123  // Mixed Case
Doesnt:   AbCdEf!@#  // Shifted Numbers
```
more information [here](http://security.stackexchange.com/a/68014). I think that can be easy to implemente with [humanize](https://github.com/HubSpot/humanize) package for try combination or simply when mongo save the user do it in `lowerCase` and define a differently between the name in the database and the name in the website.

In other order of things, trim spaces left and right can be easy and useful of remove with [trim](https://www.npmjs.com/package/trim) package.

What is your opinion?

splendido:
Hi @Kikobeats,
I remember the discussion about case sensitive username input is an old one...
See for example #405.

Btw, Telescope uses [useraccounts](https://github.com/meteor-useraccounts/core) for the user accounts UI.
And with useraccounts you can setup the input field to be *lowercased* as well as *trimmed* (see the options for [form field configuration](https://github.com/meteor-useraccounts/core#form-fields-configuration)).
Lets have a look at the  [current configuration](https://github.com/TelescopeJS/Telescope/blob/master/lib/config/at_config.js)

So we can just change the current settings if there's enough agreement on this.
...or you might think to change them on your deployed fork if you wish.

While, about accepting different mixed case, I guess this is a more sensible discussion...
Should we consider this as an additional plug-in package for useraccounts@2.0? (...I'm going to start it soon)
This way anyone interested to enable it could simply add another package to his own fork of Telescope.

Thoughts?

Kikobeats:
oh, nice!

Maybe enable `lowercase` is more controversial, but why not enable `trim` in username?

Personally I will enable `lowercase` in my fork.

on the other hand the idea of account plugin package for mixed case sound amazing. If is possible to resolve like a plugin package you have my vote :-)

splendido:
Yeah, I also agree on the `trim`.
@SachaG are you with us?

splendido:
@Kikobeats, to enable the lower case on your fork this should be put on a file shared between the client and the server:

[...]

PR #728, opened 2015-01-28
Lowercase usernames and Email plus trim and space removal

splendido:
I've managed to update `useraccounts` to the latest version to enable trim/lowercase/remove_spaces on username, email and username_and_email fields.

...the only thing is I had to update iron:router to the latest version and meteor to 1.0.3.1.
To this aim I had to lock the version of other packages to prevent their update (it not me supposed to update packages I don't know how are used within Telescope... ;-) )

**Please give another try to this branch and do some test BEFORE mergin this!**

this should solve both #691 and #723.


splendido:
mmm, I had in mind to PR to `devel` not to `master` :(
do you want me to make another PR and close this one?

SachaG:
Sorry it's taken me so long to look into this. I tried your code and it seems to work :)

But I'm not sure if the UX is really what we want: right now it seems impossible to use *any* capitalization in usernames. So if I type "TechCrunch", it gets converted to "techcrunch" automatically. 

While I agree that we should ideally let people sign in with both "TechCrunch" and "techcrunch", I don't think forcing all usernames to lowercase is a great solution…

Sorry for not paying attention to that earlier :(

queso:
So you see "Travis" and "travis" as different usernames?

SachaG:
RIght now yes, you can sign up as both `Travis` or `travis`, and it'll create two user accounts.

queso:
This is a fundamental bug in Meteor, not a feature :)

This issue has been around a while: https://github.com/meteor/meteor/issues/550

splendido:
No worries @SachaG!
I understood you were comfortable with this, also fro the discussion about the migration process to convert username to lowercase and warning for possible conflicts ;-)
...btw, just let me know how you intended this and I'll be happy to update the code accordingly.

@queso would you be to keep it like this?

...or do you guys think leaving the username unchanged and considering it lowercased behind the scenes would provide for a better UX? (this might be a little tricker to implement, but not impossible by adding a few lines for this particula case...)

SachaG:
Yes, it's my fault, I didn't look at this closely enough. The solution I had in mind was what you suggested, leaving the username unchanged and testing behind the scenes. 

For example, you could have two username fields (say, `username` and `usernameLowercase`) and use `usernameLowercase` for all uniqueness checks?

splendido:
...mmm, what you're asking would require to change the Accounts package which is something I wouldn't  consider!

What about setting `username` always lowercased and keeping another  `profile.username` with the original case to be used on the client side to show it the way the user wanted it?

SachaG:
Yes, that would also be a good solution. In fact that's already what Telescope does, except with `profile.name` (but I'm happy to change it to `profile.username` if that's cleaner).

SachaG:
See: https://github.com/TelescopeJS/Telescope/blob/master/server/users.js#L29-L31

splendido:
oh, I didn't know: I'm not that into Telescope's code, sorry!

...well, in this case it would be enough to add set `profile.name` to the original *username* when `Accounts.createUser` is called and everything should work accordingly! :-)

...I'd say we might add a check for *already existing username* on the sign up form so to prevent bad surprises: this is actually already possible see [this](https://github.com/meteor-useraccounts/core/blob/master/Guide.md#custom-validation)
Thoughts?

SachaG:
OK, but with this PR you can't enter capital letters at all:



So to recap, I think a good solution would be:

1. Let user enter whatever they want (`Sacha`)
2. Copy that to `profile.username`.
3. Change username to lowercase version (`sacha`)

Then when somebody wants to sign in or sign up, you convert what they entered to lowercase and check it against `username`. And inside the app, you only ever display `profileusername`. Sounds good?

splendido:
Yeah, that's it!
...but if you're not keen to change all `profile.name` to `profile.username` through all the code, I can make it to set `profile.name` with the original `username` field content.
Actually I'll have to set one of the two in any case...

Btw, If you have one common helper to pick up `profile.name`, lets change that one! ;-)

I'll update useraccounts to have one more option to deal with this...

splendido:
I've just updated the `lowercase-usernames` branch with the `devel` one!
So perhaps this PR should be closed and a new one made to make it into `devel`...

In any case, now it LGTM. I've done some simple tests by hands but if you could revise it a little more it would be great!
I guess you've a better idea about the possible implications...

for example, should we update also the `user.slug` field within the migration code?

splendido:
Actual changes are better viewed with [this comparison](https://github.com/TelescopeJS/Telescope/compare/lowercase-usernames)

SachaG:
Awesome. And yes, there's a single helper for displaying usernames: 

https://github.com/TelescopeJS/Telescope/blob/devel/lib/users.js#L24-L35

A new PR would be great. And I think we probably don't need to modify `user.slug`, since it was lowercase to begin with?

splendido:
So [these lines](https://github.com/TelescopeJS/Telescope/blob/devel/lib/users.js#L26-27) should be changed to 

```javascript
    if (user.profile.username)
      return user.profile.username;
```

but then what's the reason for [getDisplayName](https://github.com/TelescopeJS/Telescope/blob/devel/lib/users.js#L36)?

splendido:
[This](https://github.com/TelescopeJS/Telescope/compare/lowercase-usernames#diff-93aa3cd6516e4575bf93e79539665635R37) is what I changed from `getDisplayName`...

SachaG:
Sorry you're right, we use `getDisplayName`, not `getUsername`. I should remove that one I guess.

splendido:
...another thing which is coming to my mind is:
since I'm changing the `profile.name` field for the `profile.username` one, is there any other migration which is expecting the `profile.name` to exist?

SachaG:
I don't think so. I'll test it out before merging anyway :)

SachaG:
Just to confirm, is the PR ready on your side? Or should I wait?

splendido:
It's ready for me, if you don't see any possible conflict with other stuff lets merge it ;-)

...if you prefer, lets make a manual  merge of `lowercase-usernames` into `devel`

SachaG:
One thing I forgot to mention is that the “display name” (which is what we store in `profile.username` can include spaces or any characters). 

So to recap, we could have people entering `The Next Web` or `TheNextWeb`, and they should both resolve to `thenextweb`. 

It seems to be working fine right now, but I just wanted to make that clear :)

SachaG:
Merged on `devel`.

SachaG:
Oh btw, the migration was failing on users that don't have a username (for example those that signed up via Twitter).

splendido:
yep, the `profile.username` is set to the original text entered at registration time!

...good call on the migration of users without a username! :(

splendido:
an optional step further might be checking the availability of the selected username before form submission.
Everything is already supported (I also have an example about how to achieve this in the [custom validation](https://github.com/meteor-useraccounts/core/blob/master/Guide.md#custom-validation) section): basically we can have a method called when the `username` input loses the focus (or whatever we prefer) and show a spinbox (or something...) next to the input to eventually let the user know whether the selected username is available or not before she clicks *Register*

Let me know if you think this could be good to have!
(it might open some *security* issues though, like automated username listing...)

SachaG:
This would be a nice touch but it doesn't seem critical right now. I'm already very grateful for all the work you've done on this :)

splendido:
My pleasure!

...next big step will be accounts melding ;-)

PR #733, opened 2015-01-29
*Added russian translation (not full, because some functions are fixed i...

fobo66:
...n English);

+Added semicolons in users.js' clog callbacks. Also, i find some unuseful functions in users.js. Can i delete it?


lib/locales/ru.js
```diff
@@ -0,0 +1,194 @@
+
+i18n.translations.ru = {
```
anthonymayer: this isn't the right place to do translations anymore and the latest version of master already has russian translations (see https://github.com/TelescopeJS/Telescope/blob/master/i18n/ru.i18n.json). Make sure you pull master and try again. Also, it's generally recommended to submit pull requests against `devel` instead of `master`

server/users.js
```diff
@@ -70,7 +70,7 @@ Meteor.methods({
     );
   },
   numberOfPostsToday: function(){
-    console.log(numberOfItemsInPast24Hours(Meteor.user(), Posts));
+    console.log(numberOfItemsInPast24Hours(Meteor.user(), Posts)); //does anyone know why is it here?!
```
anthonymayer: yea, this doesn't seem terribly useful. If this function still exists on master it's probably fine to delete it (make sure to delete it wherever it's called too).

server/users.js
```diff
@@ -39,14 +39,14 @@ Accounts.onCreateUser(function(options, user){
   // ------------------------------ Callbacks ------------------------------ //
 
   // run all post submit client callbacks on properties object successively
-  clog('// Start userCreatedCallbacks')
+  clog('// Start userCreatedCallbacks');
```
anthonymayer: ++

lib/locales/ru.js
fobo66: I'm sorry, i don't know about it. So, this PR becomes bad thing, please decline in and i'll make one another.

SachaG:

PR #740, opened 2015-01-31
Polish translation

juliancwirko:
Polish translation files (main and packages)


SachaG:
That's awesome, but that pull request looks a little strange… could you submit one that only has the language file changes?

juliancwirko:
Ok, sorry about that (devel merged into it..) there is another PR with only lang files: [#747](https://github.com/TelescopeJS/Telescope/pull/747)
Thanks.

PR #741, opened 2015-02-01
Fix Google Analytics

Sean-R:
I couldn’t get Google Analytics to work using the Tracking ID in the
admin section until I changed the code to this.


SachaG:
Interesting. Is that the code recommended by Google?

Sean-R:
Yeah, it was the only difference between my tracking code and what was in Telescope's function (other than my ID of course).

SachaG:
Thanks :+1:

PR #744, opened 2015-02-02
Brazilian portuguese translation

ibrahimcesar:
Hi @SachaG 

I've added core translation of Telescope for Brazilian Portuguese (pt-BR) and some packages!


SachaG:
Thanks! Not sure why it didn't pass the test though.

SachaG:

PR #747, opened 2015-02-03
polish translation - only lang files

juliancwirko:
Cleaned PR with only lang files


SachaG:
Awesome :)

SachaG:

PR #755, opened 2015-02-09
Added category assignment to feeds

Accentax:
I added category assignment by adding a field to the schema. You can assign multiple categories to an rss feed. It should work aslo without any category created or assigned. Also the feedId is now assigned to the post.


SachaG:
Awesome job, I'll test it out and merge it :+1: 

But wasn't the feed id already assigned to the post? Was that not working?

Accentax:
Great! The feedId didnt assign correctly.

SachaG:
By the way, for next time make sure you do the PR into its own branch, not on `master` :)

PR #757, opened 2015-02-11
Add some basic indexing for posts and comments. Not comprehensive.

queso:


SachaG:
Thanks again :) By the way, I'm not super-familiar with that part of the GitHub workflow, but is it possible to make PRs into `devel` instead of `master`?

Slava:
@SachaG make `devel` the default branch in repo's settings.

SachaG:
@slava done, thanks for the tip!

PR #765, opened 2015-02-13
rebuild user management page with reactive-table

jshimko:
This is a complete rebuild of the admin user management page (/all-users) that uses [aslagle:reactive-table][1].  

You can now sort any column by clicking the headers (with the exception of things that don't sort - avatars, user actions, etc.).  You can also filter/search users in realtime by typing in the input box in the top right (currently searches all data fields in the table by default).  Pagination functionality came for free with reactive-table and is MUCH nicer now.  The only trade-off is that there is no longer a way to build queries in the url to link to a specific table state (but it's entirely possible to build that).  That said, I left all of the old routing alone so that we could possibly use it as a starting point to build that functionality back into it.  However, I suspect nobody uses it anyway, so it wasn't a big priority when I put this together.  I think user searching/filtering/sorting was the biggest pain point and this update definitely covers all of that.

All of the logic for the table is [here][2] and very few [additional helpers][3] were needed thanks to the built in functionality of reactive-table.

Let me know if I missed anything!

[1]: https://github.com/aslagle/reactive-table
[2]: https://github.com/jshimko/Telescope/blob/users-list/client/views/users/list/users.js
[3]: https://github.com/jshimko/Telescope/blob/users-list/client/views/users/list/users_list_actions.js


SachaG:
Wow, this looks amazing! Great job :)

jshimko:
Oops.  I need to change the pub/sub stuff for this to work properly.  The reactive-table package takes care of pagination/filtering/sorting, so I need to just publish all users (or use [ReactiveTable.publish][1]) and let the package handle the rest.  Right now the allUsers subscription depends on the url to set the limit/filter/sort, so reactive-table only sees the default subscription of the collection (limit of 20, no filter, and sort by createdAt).  Didn't catch that until I used it in production with more than a few test users and couldn't access more than 20 users.

I'm off to bed now, but I'll sort it all out tomorrow and push the changes out.  

To be continued...

[1]: https://github.com/aslagle/reactive-table#server-side-pagination-and-filtering-beta

SachaG:
Well we shouldn't publish *all* users. But I guess that means the filtering and search won't really work then… If it can only filter on published data, that's not quite as impressive as I first thought :)

anthonymayer:
Nice! I was trying to clean it up in https://github.com/anthonymayer/Telescope/commit/fde9084d6d4064e2494545e5a51988679f995a5a too but this is way better.

jshimko:
Ok, I think this is sorted now.  I used [ReactiveTable.publish][1] to handle the publication [here][2].  It communicates directly with the reactive-table instance and does all of the limit/filter/sort pub/sub stuff right out of the box.  Pretty awesome actually.  If you're interested, check out the source for ReactiveTable.publish here: 

https://github.com/aslagle/reactive-table/blob/master/lib/server.js

Give [my users-list branch][3] a test drive and let me know if you can break it!

[1]: https://github.com/aslagle/reactive-table#server-side-pagination-and-filtering-beta
[2]: https://github.com/jshimko/Telescope/blob/users-list/server/publications/users_dashboard.js
[3]: https://github.com/jshimko/Telescope/tree/users-list

SachaG:
I'm finally finding some time to look at this, sorry for the delay! A few comments:

- It'd be nice to have prev/next page links. 
- Where is the data published to? `Meteor.users.find().count()` just returns 1. 
- It's a shame that Reactive Table uses a different i18n package. Maybe `anti:i18n` could become a weak dependency instead?

SachaG:
Merged on the `devel` branch.

PR #766, opened 2015-02-13
Various accessibility fixes, mainly hiding unnecessary elements from scr...

ndarilek:
...een readers, adding screen-reader-only content and using live regions for real-time content.

Telescope sites have a bad UX for screen reader users. First there are the use of icons for upvoting and commenting with no text equivalent. I added sr-only spans with text descriptions of these, and aria-hidden to the icons since some of them have spoken descriptions which don't make much sense or add clutter (interestingly enough, comments have "upvote"/"downvote" as words but posts don't, fixing that minor inconsistency would eliminate the need for the sr-only span for posts.)

Next I added aria-hidden to all the avatar links, hiding them from screen readers. They seem to be followed up with links to the profiles themselves so they seem a bit redundant, and screen readers spoke the images as a string ID.

Finally I added aria-live="polite" to the posts and comments lists, plus a few aria-live="off" attributes where it made sense. This makes screen readers automatically read new posts/comments as they arrive, but hides unnecessary changes like the recalculation of scores every 30 seconds.

I likely did one or two additional minor things that I'm forgetting. There's more work to do still, but I'll likely discover what needs done in time, particularly as this code rolls out to the various Telescope sites I read and I experience the UX first-hand.

### DIFF
--- a/client/views/comments/comment_item.html
+++ b/client/views/comments/comment_item.html
@@ -7,16 +7,16 @@
       <div class="comment-content">
         <div class="comment-actions {{#if upvoted}}upvoted{{else}}not-upvoted{{/if}} {{#if downvoted}}downvoted{{else}}not-downvoted{{/if}}">
           <a class="upvote" href="#">
-            <i class="icon-up"></i>
+            <i class="icon-up" aria-hidden="true"></i>
             <span>{{_ "upvote"}}</span>
           </a>
           <a class="downvote" href="#">
-            <i class="icon-down"></i>
+            <i class="icon-down" aria-hidden="true"></i>
             <span>{{_ "downvote"}}</span>
           </a>
         </div>
         <div class="comment-meta">
-          <div class="user-avatar avatar-medium">{{> avatar userId=userId shape="circle"}}</div>
+          <div class="user-avatar avatar-medium" aria-hidden="true">{{> avatar userId=userId shape="circle"}}</div>
           <a class="comment-username" href="{{profileUrl}}">{{authorName}}</a>
           <span class="comment-time">{{timeAgo ago}},</span>
           <span class="points">{{upvotes}}</span> <span class="unit">points </span>
--- a/client/views/comments/comment_list.html
+++ b/client/views/comments/comment_list.html
@@ -1,5 +1,5 @@
 <template name="comment_list">
-  <ul class="comments comment-list">
+  <ul class="comments comment-list" aria-live="polite">
     {{#each child_comments}}
       {{> UI.dynamic template=comment_item}}
     {{/each}}
--- a/client/views/comments/comment_reply.html
+++ b/client/views/comments/comment_reply.html
@@ -8,7 +8,7 @@
     {{/with}}
 
     {{#with comment}}
-      <ul class="selected-comment">
+      <ul class="selected-comment" aria-live="polite">
        {{> UI.dynamic template=comment_item}}
       </ul>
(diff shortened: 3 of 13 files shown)


SachaG:
Great work! Accessibility is definitely important. I just don't know that much about it myself…

SachaG:
Actually I approved this a bit too fast, it seems like it was missing a `display:none` on the `sr-only` class. I added it, I assume it won't mess up too much with screen readers?

ndarilek:
Actually, I think it will be a problem, since display: none hides the
element visually and the screen reader won't show it then, since
display: none is used to legitimately hide things that might later show,
whereas sr-only hides the element for all but screen readers. There are
legitimate instances where you want to hide text, and display: none
doesn't distinguish those from hiding from screen readers.

So is the sr-only text showing somewhere?

SachaG:
Well yes, since I didn't see any CSS rules to ensure it doesn't show? I'll leave `display:none` for now, but feel free to contribute a better, more accessible way to hide those elements. Or maybe we can remove them and use the `title` attribute on preceding elements instead? Not sure how screen readers work, sorry…

SachaG:
Oh also please make future PRs against the `devel` branch :)

ndarilek:
Here is the Bootstrap sr-only class:

.sr-only {
  position: absolute;
  width: 1px;
  height: 1px;
  padding: 0;
  margin: -1px;
  overflow: hidden;
  clip: rect(0, 0, 0, 0);
  border: 0;
}

Looks like that should have elements that make the text invisible--in
particular, setting it to 1px wide.

I'm pretty sure display:none won't work, since that will make it
disappear from screen readers as well.

I'd appreciate help with this, since I myself am blind and can't confirm
whether the sr-only class as documented on Bootstrap's site isn't hiding
text visually. If it isn't, that should be filed as a Bootstrap bug,
since as documented that text shouldn't appear visually.

Thanks.

SachaG:
Perfect, that worked. 

And sorry again for not being more familiar with accessibility. I assume being blind means browsing through a codebase can take you a bit longer, so I'm really grateful for the time you're spending contributing to Telescope :)

ndarilek:
Cool, no worries, not sure that it takes longer but I appreciate the
sentiment. :)

So to clarify, did the sr-only patch as shipped still leave the text
visible? I ask because I've submitted a similar patch to another project
and it is being rejected, but I think they may not get that I can't see
that the text is visible (assuming that it is.)

SachaG:
Yes, it did leave the text visible. Since Telescope doesn't use Bootstrap, there was no predefined `sr-only` CSS class to hide the elements you added. I wonder why that's not a standard browser feature by the way…

ndarilek:
Oh, that explains it. :) I saw some bootstrap packages in
.meteor/packages so assumed that it was being pulled in under one of the
many subpackages.

There are some additional sr-only definitions, here is the entire block
of code:

.sr-only {
  position: absolute;
  width: 1px;
  height: 1px;
  padding: 0;
  margin: -1px;
  overflow: hidden;
  clip: rect(0, 0, 0, 0);
  border: 0;
}
.sr-only-focusable:active,
.sr-only-focusable:focus {
  position: static;
  width: auto;
  height: auto;
  margin: 0;
  overflow: visible;
  clip: auto;
}

There's something else, but not sure if it's relevant:

.has-feedback label.sr-only ~ .form-control-feedback {
  top: 0;
}

PR #771, opened 2015-02-15
Calling analyticsRequest in onRun instead of onAfterAction

elie222:
onAfterAction can be called multiple times per route change, whereas onRun is only called once per page change

### DIFF
--- a/lib/router/filters.js
+++ b/lib/router/filters.js
@@ -166,7 +166,7 @@ Meteor.startup( function (){
 
     // Router.onAfterAction(filters.resetScroll, {except:['posts_top', 'posts_new', 'posts_best', 'posts_pending', 'posts_category', 'all-users']});
     Router.onAfterAction(analyticsInit); // will only run once thanks to _.once()
-    Router.onAfterAction(analyticsRequest); // log this request with mixpanel, etc
+    Router.onRun(analyticsRequest); // log this request with mixpanel, etc
     Router.onAfterAction(filters.setTitle);
 
     // Unload Hooks


jshimko:
onRun definitely seems like the right place to do this (according to IR [docs][1]), but I've found that the title/path/url data in that context is unreliable because the route doesn't appear to have finished running yet.  So you do get a single analytics call, but all the details are from the page you're leaving instead of where you're headed. (I use Segment.io and watch the events fire in their realtime debugger).

To get around this, I've been building the pageview details myself and adding them to the analytics call in my controllers.  However, that feels like an ugly band-aid on the problem and not a permanent solution.

For example, I do this in the PostPageController (the route for a single post - /posts/:_id )...
```javascript
  onRun: function() {
    var post = Posts.findOne(this.params._id);

    var title = getSetting('title', "Telescope") + ' - ' + post.title,  // calling this.getTitle() is never right here
        host = window.location.host,
        path = '/posts/' + this.params._id,  // can't use window.location.path because it's still the old page at this point
        url = host + path;

    analytics.pageview({
      title: title,
      path: path,
      url: url
    });

    this.next();
  }
```
This is the only option I've tried that actually sends the correct data reliably (and only once).  Otherwise, I'm always a page behind with the title, path, and url (which totally screws up the accuracy of my stats).

Not sure what the right answer is here, but I think this problem definitely could use a brain storming session (and perhaps some official advice from @cmather).

Thoughts?

[1]: https://github.com/iron-meteor/iron-router/blob/devel/Guide.md#available-hook-methods

splendido:
while waiting for a more official solution,

...would a flag `analyticsSent` set to false inside the `onRun` be enough to send analytics only once inside `onAfterAction`?

```
onRun: function() {
  this.analyticsSent = false;
},
onAfterAction: function() {
  if (!analyticsSent) {
    analytics.pageview({
      ...
    });
    this.analyticsSent = true;
  }
}
```

just my two cents ;-)

elie222:
Seems like a good idea. I'm surprised no official solution is out there
yet. Analytics packages exist but from what I've seen they all have these
problems.

splendido:
We might also think to write a `plugin` for IR which would be something like this:

```javascript
Iron.Router.plugins.sendAnalytics = function (router, options) {

  router.onRun(function(){
    // Marks analytics were not yet sent for this route!
    this.analyticsSent = false;
  });

  router.onBeforeAction(function(){
    if (!analyticsSent) {
      // Retrieves metadata for current route
      var routeMetadata = {};
      // ...

      // Actually Sends Analytics
      analytics.pageview(routeMetadata);

      // Remembers analytics were already sent!
      this.analyticsSent = true;
    }
  });
};
```

to be used like this:

```javascript
Router.plugin('sendAnalytics', {
  // except: ['routeOne', 'routeTwo']
  // or only: ['routeOne', 'routeTwo']
});
```

see also the [IG Plugin Guide](https://github.com/iron-meteor/iron-router/blob/devel/Guide.md#plugins)

elie222:
One thing I just remembered is that in the okgrow:analytics package,
onAfterAction is currently used to make analytics requests, and the info
there is inconsistent. Path, route name, url sometimes belong to the
previous route instead of the current one. And aren't always consistent. So
we'd have to make sure that whatever hook we use has consistent data.
Perhaps this is even an IR issue.

xavcz:
Hey, folks! Telescope has changed a lot in the past months, and there is a new React-based version, [Telescope Nova](https://github.com/TelescopeJS/Telescope) ([intro video](https://www.youtube.com/watch?v=_cLv6fH-PSQ&index=1&list=PLBoa_Q6hVeSzChsSi8X8x3FQ8NN2D73fB)). If you know Crater, it runs on the latest version.

The Legacy version is no longer supported. Thanks for your PR @elie222, but I'm closing it.

Happy hacking! 🚀

PR #781, opened 2015-02-20
More bulgarian translations

durrrr91:
Updated the main bg.18n.json and also translated some packages to keep everything consistent.


SachaG:
Thanks!

SachaG:

PR #783, opened 2015-02-24
Mailchimp limits email subject to 150 characters

thingista:
Campaign.js trims the newsletter's subject to 15 words which is enough in most cases. Yet, I have had the case where 15 words are more than 150 characters and the Mailchimp campaign fails.

### DIFF
--- a/packages/telescope-newsletter/lib/server/mailchimp.js
+++ b/packages/telescope-newsletter/lib/server/mailchimp.js
@@ -8,16 +8,23 @@ scheduleCampaign = function (campaign, isTest) {
 
   if(!!apiKey && !!listId){
 
+		var wordCount = 15;
+		var subject = campaign.subject;
+		while (subject.length >= 150){
+			subject = trimWords(subject, wordCount);
+			wordCount--;
+		}
+		
     try {
 
       var api = new MailChimp(apiKey);
       var text = htmlToText.fromString(campaign.html, {wordwrap: 130});
-      var defaultEmail = getSetting('defaultEmail');
+      var defaultEmail = getSetting('defaultEmail');			
       var campaignOptions = {
         type: 'regular',
         options: {
           list_id: listId,
-          subject: campaign.subject,
+          subject: subject,
           from_email: defaultEmail,
           from_name: getSetting('title')+ ' Top Posts',
         },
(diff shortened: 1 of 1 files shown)


jshimko:
On a related note, I wonder if it's possibly worth making a setting for the subject line that gives you the option to do the current behavior or use a custom field.  I ended up hard coding a subject for mine, but that's clearly a yucky solution.  

Thoughts?  Anyone want that?  I'd be happy to do it if other people would find it useful.

thingista:
@jshimko Personally I like the fact the subject is always "up to date" without the need for a manual update. If you want to preserve this feature and add the ability to manually tweak it, will it become a bit complicated (as in "do you want to prepend or append something", etc.)?

jshimko:
It's actually not bad at all.  It's just an input field that allows you the ability to switch between the current behavior or the value of that input field.  I can't see any reason why it would be much more than a dozen or so lines of code across a few files.

Pretty much just a settings schema addition...

``` javascript
var newsletterSubject = {
  propertyName: 'newsletterSubject',
  propertySchema: {
    type: String,
    optional: true,
    autoform: {
      group: 'newsletter',
      instructions: 'Leave blank for default subject line or type your own here',
      private: true
    }
  }
}
addToSettingsSchema.push(newsletterSubject);
```
 and then something like the following on [this line][1].
``` javascript
// uses value of input field or defaults to original behavior if empty
subject: getSetting('newsletterSubject', trimWords(subject, 15))
```

[1]: https://github.com/TelescopeJS/Telescope/blob/devel/packages/telescope-newsletter/lib/server/campaign.js#L65

thingista:
Sorry @jshimko my comment was not clear at all. What I meant is personally, I could potentially want to prepend a custom text, but I would still want to append a concatenation of the posts' titles. Something like (subject of the newsletter):

> New this week on xyz: How to...

I think this latter, dynamic part is important for my users. But maybe it's just me and other people would rather replace the title completely with a static string.
Thing is, it adds a bit of complexity to the UI since it can't just be a "toggle" in this case.

SachaG:
Thanks! And I agree, I don't think a setting for the subject is needed for now. Maybe later once we rethink the settings area to be a bit easier to navigate (maybe have a way to toggle "advanced" options)?

PR #784, opened 2015-02-24
extraCode helper in layout template was not defined

jshimko:
The `{{{extraCode}}}` helper [in the layout template][1] was not defined anywhere and was not outputting anything as a result.  This fixes that.

[1]: https://github.com/TelescopeJS/Telescope/blob/devel/client/views/common/layout.html#L18

### DIFF
--- a/History.md
+++ b/History.md
@@ -4,6 +4,10 @@
 * Added setting for external fonts.
 * Use site tagline as homepage title.
 * Make favicon customizable.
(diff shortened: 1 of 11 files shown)


jshimko:
Sorry, I was supposed to branch this from devel apparently.  It's trying to merge in other stuff I branched of of master with.  I'll close and resubmit.

PR #785, opened 2015-02-24
extraCode helper in layout template was not defined

jshimko:
The `{{{extraCode}}}` helper [in the layout template][1] was not defined anywhere and was not outputting anything as a result.  This fixes that.

[1]: https://github.com/TelescopeJS/Telescope/blob/devel/client/views/common/layout.html#L18


SachaG:
Thanks!

PR #804, opened 2015-03-11
Improvement on vote accesability through packages 2nd pass

Baxter900:
As per the conversation here: https://github.com/TelescopeJS/Telescope/pull/797
* Changed votePowerEq to getVotePower.
* getVotePower now accepts both Numbers and Functions.
* getVotePower now throws an error when set to an object type which is not a function and cannot be converted to a Number.
* The "Method" callbacks are now performed before votes are processed.

### DIFF
--- a/lib/vote.js
+++ b/lib/vote.js
@@ -47,13 +47,22 @@ var removeVote = function (userId, itemId, collection, upOrDown) {
 };
 
 upvoteItem = function (collection, item, user) {
-  user = typeof user === "undefined" ? Meteor.user() : user,
-      votePower = parseForVoteNumber(user),
-      collectionName = collection._name.slice(0,1).toUpperCase()+collection._name.slice(1);
+  user = typeof user === "undefined" ? Meteor.user() : user;
+  collectionName = collection._name.slice(0,1).toUpperCase()+collection._name.slice(1);
 
   // make sure user has rights to upvote first
   if (!user || !can.vote(user, true) || hasUpvotedItem(item, user))
     return false;
+
+  // ------------------------------ Callbacks ------------------------------ //
+
+  // run all post submit server callbacks on post object successively
+  result = upvoteMethodCallbacks.reduce(function(result, currentFunction) {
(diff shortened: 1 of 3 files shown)


Baxter900:
Followup pull request: https://github.com/TelescopeJS/Telescope/pull/804

SachaG:
Maybe I'm wrong, but it doesn't seem you're doing anything with the `result` of the callbacks? Shouldn't we pass it as `item` to the other operations?

And on second thought, I'm also not 100% sure how useful it is to accept both functions and numbers for `getVotePower`. It seems a bit of an edge case to be honest? Letting people customize the function if they want to is probably enough, imho.

SachaG:
If you're ok, I'll merge this in and then tweak it a little bit myself to simplify it.

Baxter900:
On the topic of the result of the callbacks, yea, possibly. To be honest, I'm not exactly sure why we were setting a value which is never used to the result of the callback, but it was done with the callbacks before I altered them, and on the other callbacks. I decided maybe I was missing something, and I left it as it was. If you have a better idea of what should be done with that, you should probably do it. I looked around for a reason and couldn't find one.

As for why I made it work with both a function, and a number, I figured that if one would ever intend to change the vote based off of something else, like maybe a separate option entered elsewhere. It would be far easier to type `getVotePower = calculatedPower;` rather than <pre><code>getVotePower = function (user) {
    return calculatedPower;
};</code></pre>
Thus I implemented it this way purely for the convenience of those who may create packages in the future. It's only a bit longer, but implementing it also only took a bit longer.

Anyway, feel free to merge it and make tweaks, I think it's at a pretty good state now functionality-wise, but you're probably right about the result of the callbacks not being used in a good way.

I think it might be best to leave the code for allowing it to be both a number or a function there, as it's a nice feature for future programmers, and it isn't that much longer. It does require a comment line explaining it, but explaining where getVotePower even existed requires several lines of comments less future programmers have to go through a wild goose chase to figure out that it exists in base.js in the base package, and that it might be being altered by a package.

PR #809, opened 2015-03-12
Fix broken twitter avatars

bengott:
- services.twitter.profile_image_url_https field wasn't being published

### DIFF
--- a/lib/publications.js
+++ b/lib/publications.js
@@ -15,6 +15,7 @@ privacyOptions = { // true means exposed
   'profile.site': true,
   'profile.twitter': true,
   'services.twitter.profile_image_url': true,
+  'services.twitter.profile_image_url_https': true,
   'services.facebook.id': true,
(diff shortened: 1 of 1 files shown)


SachaG:
Thanks!

SachaG:
Actually it still seems like something's broken? Looks to be a CSS issue. Did something change with that?

http://cl.ly/aCFo

bengott:
No CSS changes. It was just failing because that twitter avatar url wasn't accessible client-side, so the img element had no src, and it was just showing the alt="avatar" text inside the div. Are you still having problems?

SachaG:
Those are avatars of people who logged in with their email/password/username. So I don't know if the twitter thing has an impact on that?

bengott:
Hmm, it should be trying to use Gravatar in that case, then falling back to initials if the image fails to load.. I'll look into it more later.

PR #810, opened 2015-03-13
Fix broken twitter avatars (for real this time)

bengott:
- didn't notice avatarOptions object last time

Thanks @saimeunt!


SachaG:
Thanks! Seems to work :)

PR #811, opened 2015-03-14
Correct base score calculation in vote.js

Juaanma:
When we downvote an item or cancel one of its upvotes, we should decrese, not increase, the base score.


SachaG:
Good catch, thanks!

PR #814, opened 2015-03-14
Refactor postAfterEditMethodCallbacks execution on server

saimeunt:
postAfterEditMethodCallbacks execution behaviour was copied straight away from postEditMethodCallbacks execution : perform a reduction on the post modifier by executing every callbacks.
Since postAfterEditMethodCallbacks is executed after the actual DB update, we don't need to alter the modifier because it won't be used, so we can use a plain forEach instead of a reduction.
We also need to pass the post ID to the callback (as second argument so it won't break existing callbacks) because there's no way to identify the post that was modified in the callback if we don't.

Other modification on line 560 is due to trailing whitespace auto removal in my code editor.

### DIFF
--- a/collections/posts.js
+++ b/collections/posts.js
@@ -492,10 +492,10 @@ Meteor.methods({
 
     // ------------------------------ Callbacks ------------------------------ //
 
-    // run all post submit server callbacks on modifier object successively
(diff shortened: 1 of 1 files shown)


SachaG:
Makes sense, thanks! I don't think I've ever used `postAfterEditMethodCallbacks` for anything yet anyway.

PR #820, opened 2015-03-17
Assign feed item's categories to post.

delgermurun:
Assign feed item's categories to post.
Changed htmlparser2 to node-feedparser

### DIFF
--- a/packages/telescope-post-by-feed/.npm/package/npm-shrinkwrap.json
+++ b/packages/telescope-post-by-feed/.npm/package/npm-shrinkwrap.json
@@ -1,22 +1,19 @@
 {
   "dependencies": {
-    "he": {
-      "version": "0.5.0"
-    },
-    "htmlparser2": {
-      "version": "3.8.2",
+    "feedparser": {
+      "version": "1.0.0",
       "dependencies": {
-        "domhandler": {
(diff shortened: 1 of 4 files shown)


SachaG:
Awesome work. The only thing I would suggest is maybe comment the code a little bit more? Especially the categories stuff, since there's a lot of different variables with similar names.

SachaG:
Merged on `devel`.

SachaG:
Why is there an `if` here? 

https://github.com/delgermurun/Telescope/blob/post-by-feed-fix/packages/telescope-post-by-feed/lib/server/fetch_feeds.js#L33

Shouldn't we want to add feed categories even if the item has no RSS categories? Or maybe I don't understand that part?

SachaG:
Oh ok never mind, I get it. I think I would just do `itemCategories.concat(feedCategories)`, it's ok if `itemCategories` is empty.

SachaG:
Let me know what you think of my changes: https://github.com/TelescopeJS/Telescope/commit/7be7276980909a2d7b5770c9086ce12094fcd67c :)

delgermurun:
@SachaG nice refactor :+1: I forgot about duplicating :)

PR #821, opened 2015-03-17
Server-side rendering package ?

saimeunt:
Hi there,

Here is my WIP proposal for addressing SSR in Telescope, everything is put inside a new telescope-ssr package.
I built this POC to drastically improve server response time for social sharing spiders, especially twitterbot who has a low timeout (4 seconds), making twitter cards virtually impossible to achieve if we were to rely on standard spiderable solution.
This package creates a new server route /posts/:_id/share, in the future we might want to use the regular post URL but then we have to make sure we generate the same HTML as Blaze does on the client, or else this could be considered cloaking (generating different content for spiders and regular browsers) which is penalized by SEO.

Generating the same HTML output for both server and client environment is painful as you can see by reading the code of this package.
Maybe we can improve the spiderable solution by using PhantomJS 2 and caching the generated HTML (I wasn't able to make spiderable-cached work for some reason), then we might not need SSR.

This PR is not supposed to be merged, it's just to open a discussion about Telescope and SSR.
Waiting to hear your input on this.

There is a live implementation on this website : http://www.gadgethunt.club/

Regards,

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -88,3 +88,4 @@ telescope-subscribe-to-posts
 telescope-tagline-banner
 
 # Custom Packages
+telescope-ssr
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -72,6 +72,7 @@ meteorhacks:kadira@2.18.3
 meteorhacks:meteorx@1.3.1
 meteorhacks:npm@1.2.2
 meteorhacks:picker@1.0.1
+meteorhacks:ssr@2.1.1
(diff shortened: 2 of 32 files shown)


delgermurun:
+1

delgermurun:
How about using `_escaped_fragment_`? https://meteorhacks.com/meteor-server-sider-rendering-for-seo-purpose.html

xavcz:
Hey folks, SSR is now enabled out of the box with the react version of Telescope 👍

Issue #825, opened 2015-03-18
Clearing search keyword with backspace ( or Delete on mac ) doesn't clear last character


ywang28:
This issue is found in desktop version of this page in Chrome 41:
http://demo2.telescopeapp.org/

If search keyword is typed in slowly, focus will remain on search box. Then try to delete the keyword one character by one character with backspace. You will find the last character will pop up again after you delete it. Then if you delete it again, page will show "Sorry, we couldn't find any posts.". Notice search box is empty in this case.

ywang28:
Same issue happens when user tries to hightlight the search keyword and delete it all at once, the keyword will pop up again. My above pull request should fix this issue as well.

ywang28:
Fix is merged. Closing the issue.

tima101:
Hi All, I still the bug: when I delete keyword using "delete" (one character at the time) from the search input - I see the last deleted character reappear.

Does anyone else see the same small bug?

PR #833, opened 2015-03-20
fix bug where last character in search keyword couldn't be cleared

ywang28:
This fixes the bug in issue #825. User couldn't clear last character in search keyword because routes.js automatically sets it to be query string. We should check if user tries to clear last character before set searchQuery in Session.


ywang28:
Reposted request against devel.

SachaG:
Thanks!

PR #837, opened 2015-03-21
Added missing translations in Brazilian Portuguese

alanmeira:


SachaG:

fr0zen:
For sure, @SachaG. I'm already there with nickname frzn.

PR #839, opened 2015-03-21
Added image key to the categories scheme

dtsepelev:
So it will be possible to use images in the templates for any category


SachaG:
Merged on `devel`.

Issue #834, opened 2015-03-20
Issue for non-admin users


awrelll:
When users try to edit their own post they get this error  "get the error message Disallowed property detected: categories.0."

From the JS console:

//Disallowed property detected: categories.0
0e99b08212fd4d624ea0bf9a089bd85ed70899b0.js:3
Exception while simulating the effect of invoking 'editPost'

yourcelf:
It looks like this is related to https://github.com/aldeed/meteor-autoform/issues/765.  AutoForm is now creating modifier docs like

    {$set: {"categories.0": "id1", "categories.1": "id2"}}

instead of

    {$set: {"categories": ["id1", "id2"]}}

This impacts both the permissions check for updating categories and the value that gets set if categories is null (mongo will interpret the set as creating an object if there isn't a "categories" array present).

I'm working on a fix for both right now.

SachaG:
Oh you're right. I forgot about this issue as well (see also https://github.com/TelescopeJS/Telescope/issues/828)

awrelll:
Any available fix so far ? Thanks!

SachaG:
This should be fixed on the `devel` branch.

awrelll:
Cant you please point me to the fix ? I updated to aldeed:autoform  5.1.0 and still got the issue.
Thanks!

SachaG:
If you're running the latest AutoForm, then saving a form should overwrite categories with a new array. Is that not the case?

awrelll:
Yes. But i still get the error 
"//Disallowed property detected: categories.0"

SachaG:
That's strange. Does that happen on new posts as well?

awrelll:
no, just non-admins when trying to edit their own post. here's the repo https://github.com/ethereumbuilders/eth.hangouts
Thanks for your reply!

awrelll:
can i get some hint for my issue ? i can't figure it out. please!
Thank you !

SachaG:
I tried running the repo but I ran into a couple errors:

```
=> Errors prevented startup:

   While building package cb-theme:
   error: File not found: lib/client/stylesheets/plexus.css

   While building package cb-theme-pages:
   error: File not found: lib/client/templates/plexus.html
   error: File not found: lib/client/templates/plexus.js
   error: File not found: lib/client/templates/plexus_link.html
```

[...]

PR #845, opened 2015-03-24
Workaround categories permission and data type bug

yourcelf:
Workaround for https://github.com/aldeed/meteor-autoform/issues/765

Fixes issue #834.

### DIFF
--- a/collections/posts.js
+++ b/collections/posts.js
@@ -471,6 +471,12 @@ Meteor.methods({
         // loop over each property being operated on
         _.keys(operation).forEach(function (propertyName) {
           var property = postSchemaObject[propertyName];
+          // If we're handed an array index instead of a top-level property,
+          // try looking up by the parent array key.
+          if (!property) {
+            var parentKey = propertyName.split(/\.\d+/)[0];
+            property = postSchemaObject[parentKey];
+          }
           if (!property || !property.autoform || !property.autoform.editable) {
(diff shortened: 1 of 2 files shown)


SachaG:
Thanks! Ideally this kind of thing should be fixed at the AutoForm/SimpleSchema level, but I guess for now we can do it that way.

We just need to make sure we don't forget to remove this once the bug gets fixed.

SachaG:
This does seem a bit hacky though, let me ask @aldeed if he has a timeline to fixing that bug first before merging this :)

aldeed:
I think I am going to go back to old behavior as default and allow people to opt in to the new behavior when they need to update specific array items rather than an entire array, and they know the array already exists in the database. It's not a perfect solution because you can't do both in the same form, but oh well.

I thought people would be OK with just working around the mongo issue themselves, ensuring that all array props are never null or undefined in the db, but it's causing a lot of consternation, so it's probably better for AF to continue masking the issue.

Hoping to release this change today.

SachaG:
Thanks! I think that's the right decision for now. I don't mind working around the issue at the app level, but I think introducing that kind of breaking change is always a bit tricky, especially when it's not obvious what's going on (I wasn't familiar with the Mongo issue myself for example).

SachaG:
Closing this PR since it shouldn't be necessary anymore :)

PR #846, opened 2015-03-24
Implement getTitle and getDescription for search

yourcelf:


SachaG:
Thanks! (btw, if you can make sure you submit PRs to the `devel` branch in the future).

PR #847, opened 2015-03-24
Increse specificity of `keyup` search event

yourcelf:
It's hard to override the behavior of .search-field in a theme when the
event listener is listening to all ``keyup`` events in the template.
Make it more specific as ``keyup .search-field`` so a theme that doesn't
want the default event listener can override the input class.


SachaG:
Good idea.

PR #849, opened 2015-03-25
Create ar.i18n.json

nwabdou85:
ar.i18n.json  is the translation of  en.i18n.json


SachaG:
Thanks! Merged into `devel`.

nwabdou85:
you're welcome!  there are some items (words and phrases) in the application that I don't find in the en.i18n.json file!!

SachaG:
Yeah, sorry about that. We need to take some time to clean up the whole i18n process, and maybe start using Transifex to keep track of everything.

nwabdou85:
I understand you;
Is there à way to use only one language in Télescope and remove thé others??

SachaG:
Are you worried about loading the extra files? Because you don't have to worry about that, only files for the current language are loaded to the client.

nwabdou85:
No, what bothers me is when I refresh the page, it render the English version for a few seconds and after that it render the Arabic version; I think it's not a good user experience !! that's why I thought to leave only one language. What do you think !!

SachaG:
Oh ok. In that case it's a bug, maybe you can open a new issue about it?

nwabdou85:
yeh, I'll

nwabdou85:
It's ok, now. Thank's a lot Sacha

SachaG:

nwabdou85:
it' was a  pleasure  to join you;)  it's done

PR #854, opened 2015-03-26
Add userId param to changeEmail method

### LINKED ISSUE #852: userAccount submit handler calls `changeEmail` for logged in user instead of user being edited
The [user account form handler](https://github.com/TelescopeJS/Telescope/blob/11dd2cd49c55e10305841ed2f991973c0ab6a75a/client/views/users/account/user_account.js#L92) calls ``Meteor.call("changeEmail", email)``, which will attempt to update the email of the currently logged-in user to the value from the form.  However, the user whose account is being edited might not be the currently logged in one (for example, an admin editing another user's account).

It looks like the [``changeEmail`` method](https://github.com/TelescopeJS/Telescope/blob/11dd2cd49c55e10305841ed2f991973c0ab6a75a/server/users.js#L60) may need to be modified to take a user id in addition to the email as arguments.  As this is the only place the ``changeEmail`` method is called from, I'll assume that this is OK.  Am working on a fix.

yourcelf:
Fixes #852.

### DIFF
--- a/client/views/users/account/user_account.js
+++ b/client/views/users/account/user_account.js
@@ -89,7 +89,7 @@ Template[getTemplate('userAccount')].events({
       });
     });
 
-    Meteor.call('changeEmail', email);
+    Meteor.call('changeEmail', user._id, email);
 
   }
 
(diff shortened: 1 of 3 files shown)


SachaG:
Thanks!

PR #862, opened 2015-03-28
added Dutch translation

reneras:


SachaG:
Thanks! Merged on `devel`.

SachaG:

PR #875, opened 2015-03-31
Unify posts hooks

saimeunt:
This PR unifies posts after submit/edit hooks.

"After" hooks should both be executed only on the server and asynchronously (to avoid holding up the client).
They don't need to modify either the post or the modifier because they happen after insert/update, the after submit hook gets passed the post and the after edit hook gets passed both the modifer and the modified post (this allows fine diffing between the two to take actions only if a particular field was modified).

Shouldn't break anything because it extends the existing API without modifying expected behaviour.


SachaG:
Good idea. I would've thought `postAfterEditMethodCallbacks` was in a `defer` already. It's certainly more consistent this way :)

PR #882, opened 2015-04-02
post-by-feed: Normalize encoding to utf-8

### LINKED ISSUE #729: Wrong parsing non-utf-8 feeds rss in feed packages
For example rss - http://f1life.ru/system/services/rss.php?service=News&site=1- in windows-1251 charset. Yes, I'm also very sad that such sites are still there.

Screenshot -
20150129032459mgw7h

delgermurun:
closes #729
Meteor's HTTP.get return already decoded content.
So we needed to use nodejs's request module.
See: https://github.com/ashtuchkin/iconv-lite/wiki/Use-Buffers-when-decoding


bengott:
There's also aldeed:http. Really easy to use -- extends Meteor's HTTP methods.
https://atmospherejs.com/aldeed/http

delgermurun:
@bengott Thank you very much. I updated PR.

bengott:
Cool, glad to help. And thanks to Eric Dobbertin (@aldeed)!

aldeed:
FYI, Meteor 1.0.4+ now has [npmRequestOptions](https://github.com/meteor/meteor/blob/devel/History.md#utilities) available in the core http package, if you don't want the extra dependency and you don't need to support older Meteor versions. (I personally think the API in my pkg is a better isomorphic solution, but you take what you can get.)

delgermurun:
@aldeed Thanks. I'll take a look at it.

delgermurun:
`{npmRequestOptions: {encoding: null}}` restarts server, when response statusCode is 404.

@aldeed do you know why is it :)?

aldeed:
No I'm not sure. I haven't actually tried `npmRequestOptions` yet so maybe it's buggy.

PR #883, opened 2015-04-02
Fix a bug where the post submit autoform hook wasn't called

saimeunt:
The autoform docs specify that you should name your hook after the form type, not after the name of the method.

https://github.com/aldeed/meteor-autoform#callbackshooks


SachaG:
oh thanks!

PR #893, opened 2015-04-06
Fix #892 Meteor Method was assigned to a variable

### LINKED ISSUE #892: Cron Job?
I'm using an rss feed to post to telescope but it doesn't seem to pull the latest data from the rss feed automatically.  I keep needing to call Meteor.call('fetchFeeds');.

Do I need to set up a corn job that calls a url?

Thanks.

azizur:
Fix #892 Meteor Method was assigned to a variable

### DIFF
--- a/packages/telescope-post-by-feed/lib/server/cron.js
+++ b/packages/telescope-post-by-feed/lib/server/cron.js
@@ -1,24 +1,24 @@
 SyncedCron.options = {
   log: false,
   collectionName: 'cronHistory',
-  utc: false, 
+  utc: false,
   collectionTTL: 172800
-}
+};
 
 var addJob = function () {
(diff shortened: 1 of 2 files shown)


jshimko:
The original state is the recommended approach.  Removing the `var` makes that function a global variable.  Since the function olny gets used in the same file (right below it), a local `var` is the right choice to keep the namespace cleaner.

You can still always call that method globally anyway with `Meteor.method('fetchFeeds')`.

SachaG:
Well it's in a package so it wouldn't be global anyway. But yeah, I feel like having `var` is a bit cleaner? At least it indicates that this function is not supposed to be used anywhere else.

azizur:
Will update and raise another PR

PR #908, opened 2015-04-13
Change color names in email package

saimeunt:
Color names changed in latest Telescope version but email package was not updated to reflect the change.


SachaG:
Oh good catch, thanks!

PR #917, opened 2015-04-14
update of french translation for general app and for two packages 

13pass:
update of french translation for general app and for packages telescope-newsletter and telescope-tags


13pass:
I got wrong some translations

PR #930, opened 2015-04-24
fix external services not checking for existing email

jgdovin:
When creating an account with oauth (meteor-developer, github, google, etc) the unique identifier emails.$.address is not set anywhere. This is used by accounts-base to handle unique emails. This PR sets the emails.$.address to the email address provided during profile completion to prevent duplicate accounts in the future.

### DIFF
--- a/client/views/users/user_email.js
+++ b/client/views/users/user_email.js
@@ -16,6 +16,7 @@ Template[getTemplate('user_email')].events({
     var user=Session.get('selectedUserId')? Meteor.users.findOne(Session.get('selectedUserId')) : Meteor.user();
     var update = {
       "profile.email": $target.find('[name=email]').val(),
+      "emails": [ { address : $target.find('[name=email]').val() } ],
       "username": $target.find('[name=username]').val(),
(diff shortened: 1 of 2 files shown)


splendido:
what if someone enters an email address which is not actually owing?
...I'd say this wold prevent the real owner to sign up.

What about doing this once the user has verified her email?

queso:
@splendido: in the meantime, I have people trying to join Crater and they can't because they signed up on a main email and then forget and try to rejoin using their Meteor account. It then errors out when they try to put in their email address.

splendido:
we could try to add `splendido:accounts-meld` while waiting for `useraccounts@2.0`
it proved to be working good...

I'm available to help with this, but I'd need help from someone knowing the collections structure to deal documents transfers during *user melding* operations

splendido:
Btw, up to my best knowledge, there's no way right now to get an email address from Twitter...

xavcz:
We are using some required fields to ask the user to fill for instance their email now as a workaround 👍 

The Legacy version is no longer supported. I'm closing this PR.

Happy hacking! 🚀

PR #936, opened 2015-05-03
Remove telescope-avatar submodule (prevents errors on git pull)

mikewheaton:
The telescope-avatar package was added as a submodule but did not have a corresponding .gitmodules to checkout properly. This was resulting in "no submodule mapping" errors when pulling the repo. Sacha confirmed on Slack that the directory was there by mistake. This pull request removes it.


SachaG:
Thanks for the PR! But I just merged the `namespace2` branch that takes care of this issue, so this is probably not needed anymore.

PR #943, opened 2015-05-05
Add topLevelCommentId field to comments.

cerupcat:
This allows childComments to know what the top most parent comment is by adding a topLevelCommentId to each childComment.

### DIFF
--- a/packages/telescope-comments/lib/client/templates/comment_submit.js
+++ b/packages/telescope-comments/lib/client/templates/comment_submit.js
@@ -23,6 +23,12 @@ AutoForm.hooks({
           var parentComment = parent.comment;
           comment.parentCommentId = parentComment._id;
           comment.postId = parentComment.postId;
+
+          if(!parentComment.topLevelCommentId) { // root comment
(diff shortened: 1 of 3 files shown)


SachaG:
Awesome, thanks!

cerupcat:
Great - thanks SachaG. You're fast! :)

SachaG:
Taking another look at this, I don't think we can keep that migration. There's three nested `forEach` loops in there, and I'm not sure if that's a good idea for sites like Crater that already have a lot of comments…

Maybe we just need to leave current data alone and assume the `topLevelCommentId` property might not always exist?

cerupcat:
Yeah, I mean, if people don't need this for old comments threads, then it shouldn't be a big deal to not have a migration. I was't sure how to do the migration any simpler.

SachaG:
Yeah I don't know how you could do it simpler either. It's more about doing it with https://github.com/percolatestudio/meteor-migrations or some similar system that can account for how long it'll take. Telescope's current migrations system is pretty bare-bones, so it can mess up with the deploy process.

PR #945, opened 2015-05-06
fix for nearly all of the getting started package issues

jgdovin:
When starting with a fresh database a few items were out of date on the getting started package, this fixes all except for the default users do not have a gravatar image.


SachaG:
Great work, thanks!

PR #954, opened 2015-05-09
minor fixes

capensisma:
on post submit and setting scroll to top when user click submit and shows a message

### DIFF
--- a/packages/telescope-posts/lib/posts.js
+++ b/packages/telescope-posts/lib/posts.js
@@ -111,7 +111,9 @@ Telescope.schemas.posts = new SimpleSchema({
     },
     autoform: {
       noselect: true,
-      options: Telescope.config.postStatuses,
+      options: function () {//Telescope.config.postStatuses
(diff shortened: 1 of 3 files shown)


SachaG:
Thanks, the settings thing is a good idea. But we can't change `Telescope.config.postStatuses` like this, the idea is that statuses should be only defined once in the app. 

Also, when internationalizing any string make sure you use a key (so `i18n.t('settings_saved')`, not `i18n.t('Settings Saved')`, and then add the translation to `i18n.en.json`.

capensisma:
You're welcome.

i will work on more on the settings. i didn't know where the i18n.json is located that's why, but i know i should have used a key string for the i18n.

and for the Status i don't but if it should be defined once, so the field should be omitted for all user since i see it as an number input field.
it should be a hidden field. omit:true

The the pending post issue was solved but the still when you click a pending post 
it doesn't show post page, just a blank page.

capensisma:
Please add the settings code so it will be easier to get to the top of the page after saving it 
thanks

SachaG:
OK, I'll add it :)

capensisma:
Thanks :dancer:

Issue #959, opened 2015-05-12
Comment count unit is hardcoded


james-gibson:
The unit value of comment upvotes is hardcoded to 'points', ignoring any translation settings.

Relevant source: https://github.com/TelescopeJS/Telescope/blob/master/client/views/comments/comment_item.html#L22

Expected behavior: Unit value should reflect specified translation value.

PR #960, opened 2015-05-12
#959 - Fixed plural hardcoding issue by adding pointsUnitDisplayText …

jknisley:
…similar to the logic from post_info, where singular/plural is already handled correctly.


SachaG:
Thanks!

PR #962, opened 2015-05-13
Add Docker deployment support

arunoda:
Hi,

I've added a Dockerfile to telescope. With we can deploy Telescope with docker very easily, officially. There are few more steps for to do this.

* Create an account an Docker Hub: https://hub.docker.com/account/signup/ (May be an organization called telescope)
* Then click on the automated build as shown below to setup the automated build for every git-push

screen shot 2015-05-13 at 6 14 33 am

In that, you can target different branch to tag version. I prefer this way

* telescope/telescope:latest - master branch
* telescope/telescope:devel - devel branch

May be not everyone want to deploy telescope as a docker image, since telescope needs some customization before deploy. (Themes, etc)
But this would be great for quick deployment or showcasing things around.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,1 @@
+FROM meteorhacks/meteord:onbuild


SachaG:
Very nice, thanks! I'll get it set up :)

arunoda:
Nice.

SachaG:
Would the Dockerfile also work for https://github.com/TelescopeJS/sample-project ? Using the main Telescope repo will only be recommended for developers moving forward.

jshimko:
Yes, it'll work the same way in the root of any Meteor app.

arunoda:
Yep.

PR #965, opened 2015-05-14
Posts category filter

cerupcat:
Allow filtering on category by passing in a category id on postsList subscription.


cerupcat:
looks like structure changed. I'll do new pull after looking at latest code.

PR #969, opened 2015-05-14
split telescope:users into two packages

alexcorre:
Moved some code from `telescope:users` into `telescope:users-base` which users now depends on. This is the first step toward a telescope that can be used with an existing users db within an existing app. 

To do so they'd be able to either:
1. Fork `telescope:users-base` and add in additional schema fields, change the way roles/permissions are implemented behind telescope's api.
2. Drop in a different `users-base` package that exposes the same api to telescope

`telescope:users-base`
- namespace
- roles
- schema
- permissions
- public field objects
- user bootstrap on `Accounts.onCreateUser`

`telescope:users`
- telescope specific UI templates
- modules and menus registration
- user routes
- publications
- methods

Thinking about possibly pulling out publications and methods


SachaG:
I like the idea, but I'm not quite sure where to do the split. Is the idea they'd keep only `telescope:users` or only `telescope:users-base`? They both seem to contain pretty important stuff.

alexcorre:
telescope only knows about `telescope:users`.  `telescope:users` depends on `telescope:users-base`. The idea is that a 3rd party could drop in an alternative to `telescope:users-base` with their custom schema, and roles/permissions implementations, as well as custom `Accounts.onCreateUser` implementation.

capensisma:
That's a good idea essentially for user roles and permissions

---
This email has been checked for viruses by Avast antivirus software.
http://www.avast.com

alexcorre:
trying to get to a place where can drop in and customize user schema + permissions + roles. is this the right direction? what do you think @SachaG

SachaG:
Well I feel like you probably don't want to reimplement permissions (things like `can.editField`, etc.), so that should go in `users`, not `users-base`. 

So I guess you'd just drop in your own `users-base` in `/packages`, and it'd automatically take priority over the one from Atmosphere?

That being said, I think there's also ways you could make it work without replacing the package. For example, if you were using a `user.admin` flag instead of `user.isAdmin`, you could redefine a new `Users.is.admin()` function to check that instead. 

Or we could make the location of the "admin" property into a constant, and let people redefine just that. 

Basically I think the more of the code you can keep, the better. I don't think many people will end up adding Telescope to their app if it requires rewriting part of the users logic (and making sure it works, too).

PR #971, opened 2015-05-16
Seach within a category feature

capensisma:
You can search on the same page of a category


SachaG:
Cool that you got it working :) But basically, to do this properly I think we need to come up with a way where the `search`, `categories`, `top`, etc. post views can all be combined at will. So in the future, if someone adds, say, a `city` field to each post, they should be able to add that criteria to any search seamlessly.

capensisma:
Thanks.

we can do that and check all the posts fields then add them to search but this is a basic search functionality.

There will be an advanced search package link in the bottom of the search input where some clicks it and it shows all post fields and views : like search in titles, date from to, score, votes, by user, sort by !!

Many other options in the advanced search package.

Issue #850, opened 2015-03-25
[Feature] Ctrl+Enter to submit post


dandv:
Nice shortcut, adopted by various web apps.

SachaG:
It's on the roadmap. PRs welcome!

xavcz:
Hey @jgdovin @dandv, this feature was on Telescope Legacy (blaze) but doesn't exist anymore with the React-based version: I opened a new issue with a step by step guide to do it, if you are interested : #1471 👍

PR #1008, opened 2015-06-08
fix for #850

jgdovin:


andmitsch:
Missing statements: `(event.keyCode == 10 || event.keyCode == 13) && e.ctrlKey)` for different browser support.

Issue #986, opened 2015-05-26
Key up down on single day view


Accentax:
On single day view the selected day changes back one day on any arrow key press, effectively making it impossible to scroll using arrow keys. Encountered on chrome on windows.

PR #1013, opened 2015-06-10
Fix Single Day Arrow Nav

johnthepink:
This should fix the arrow navigation problem #986. 

- The key event was triggering for basically every key. This should limit to the left and right arrows
- The `currentDate` variable was on getting set `onCreated` and never updating. This would cause `undefined` to be passed to the Router when trying to navigate right from today's view. Now, it is updated after key press.
- I also noticed the key bindings were carrying over to other pages causes `undefined` to get passed to the Router. So, I cleaned it out on template destroyed.

Let me know if you have any thoughts!


SachaG:
Thanks! Just a small note, please make future PRs to the `devel` branch, not `master` :)

SachaG:
Merged on `devel`.

johnthepink:
@SachaG Ok deal. Thanks!

PR #1017, opened 2015-06-11
slugified title in URL

tima101:
Is it better to have posts/slugified-title/ID URL for UX?


SachaG:
Thanks for the PR! This is on the roadmap, we've been discussing it in the #slug channel on Slack. We can't break existing URLs though, so I can't accept the PR in its current state. Also check out: https://trello.com/c/dDmLKxb7/16-slugs

xavcz:
It's working well on Nova now :)

Issue #949, opened 2015-05-08
Custom CSS option


sweetppro:
An option to add custom CSS in the Settings panel would be awesome.
it currently works by enclosing css in a <style></style> tag and entering into the" Extra Code" panel.

but a dedicated CSS option, much like the Background CSS option would be much appreciated and useful.

sweetppro:
cool!

PR #1019, opened 2015-06-11
Add Extra CSS

johnthepink:
This adds an "Extra CSS" settings field that adds that CSS to every page mentioned in #949. Any thoughts? Thanks!

screenshot 2015-06-11 18 40 42


SachaG:
Yeah, this is a good idea. Thanks :)

PR #1023, opened 2015-06-15
fix settings publication to hide private fields

jshimko:
The publication was supposed to be removing private fields from the published data, but the schema object was undefined and all properties were being published as a result.


jshimko:
Sorry, wrong version.  Be right back...  :)

PR #1027, opened 2015-06-18
Added .startOf('day'); to `today` variable

jamiewilson:
I think it needs this to avoid the timezone affecting the diff. It was displaying the Today label for tomorrow.


SachaG:
Thanks :)

PR #1029, opened 2015-06-19
Fix newsletter author name issue

johnthepink:
This should fix the author not being shown in newsletters. Looks like it was just the wrong variable being passed to `Users.getDisplayName` :smile_cat:


SachaG:
Thanks! I'll merge it in manually.

PR #1030, opened 2015-06-19
Handle Deleted User In Email

johnthepink:
This fixes the issue of posts from deleted users showing up in an email campaign. It does two things:

1. Returns blank if trying to get the profile URL of a user that does not exist
2. Just displays "Submitted on" instead of "Submitted by user on" if the user does not exists

Here's an example:
screenshot 2015-06-18 21 54 59

Thanks!

### DIFF
--- a/packages/telescope-newsletter/lib/server/templates/emailPostItem.handlebars
+++ b/packages/telescope-newsletter/lib/server/templates/emailPostItem.handlebars
@@ -14,7 +14,13 @@
     <a class="post-domain" href="">{{domain}}</a>
     | 
   {{/if}}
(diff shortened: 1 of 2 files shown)


SachaG:
I'm actually just refactoring this part of the app, so I'll just merge this is manually. Thanks!

SachaG:
Although, note that the author name is denormalized on the post, so we should always be able to show a name even if the user has deleted their account.

SachaG:
Here's what I ended up doing: https://github.com/TelescopeJS/Telescope/commit/f437153d63a33d7e146346094b3c003b686d864e

johnthepink:
@SachaG Nice! That is much nicer. Thanks for the heads up!

Issue #1047, opened 2015-06-30
Creating wrong slug when editing post


delgermurun:
http://meta.telescopeapp.org/posts/DEGEkPH8sKvJJemeP/less-than-p-greater-than-v0-21rc-slugscope-less-than-p

https://github.com/TelescopeJS/Telescope/blob/devel/packages/telescope-posts/lib/callbacks.js#L30

jshimko:
Fixed in this PR:  https://github.com/TelescopeJS/Telescope/pull/1049

PR #1049, opened 2015-07-01
fix slug generation on post edit

jshimko:
Fixes https://github.com/TelescopeJS/Telescope/issues/1047

Removed the unnecessary `marked()` call that was adding HTML to the slug string.  

Also realized that the `modifier.$set.title` in the `if` statement below was redundant because the `modifier.$set` object always contains all fields regardless of which is being edited.  

So instead of:
```js
Posts.before.update(function (userId, doc, fieldNames, modifier) {
  if (Meteor.isServer && modifier.$set && modifier.$set.title) { 
  ...
```
It can be more concise like:
```js
Posts.before.update(function (userId, doc, fieldNames, modifier) {
  if (Meteor.isServer && modifier.$set) { 
 ...
```

### DIFF
--- a/packages/telescope-posts/lib/callbacks.js
+++ b/packages/telescope-posts/lib/callbacks.js
@@ -25,9 +25,9 @@ Posts.before.update(function (userId, doc, fieldNames, modifier) {
  * Generate slug when post title is updated
  */
 Posts.before.update(function (userId, doc, fieldNames, modifier) {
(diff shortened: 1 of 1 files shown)


SachaG:
>  the modifier.$set object always contains all fields regardless of which is being edited.

I assume you mean this in the case of a document being edited through Autoform? 

I'd rather keep the check just to make this a bit more generic though.

Issue #1048, opened 2015-06-30
Move useraccounts to the theme, please


queso:
I want to use useraccounts:bootstrap instead of unstyled, but the new refactored telescope forces useraccounts:unstyled as part of telescope-libs... Is there any way we can move to to the hubble theme instead, so each package can decide what they want to use?

/cc @splendido

jshimko:
I suppose we could just make that an optional package.  That'd probably be the most flexible approach (and the least amount of work!).

queso:
I would probably argue against an optional package because 100% of the first time users will want it :)

I am betting even custom themes are generally ok with it, but useraccounts-bootstrap is better for me and bad things happen when both are installed (you get template duplicate naming issues for things like errors)

jshimko:
I just mean it's installed by default, but users can remove it if they want.  Like the Embedly plugin, for example.

jshimko:
Anything in the `.meteor/packages` file is technically optional and anything in the `telescope:lib` package.js is required.

jshimko:
The only issue I can think of with having it as a dependency of the theme is someone could potentially want to use the theme, but not the `useraccounts:unstyled` package.  So they'd be in the same position as you are.  Having it in the `.meteor/packages` instead solves that for everyone.

queso:
Looks like we are back to the drawing board here? If I send a PR to move this package requirement to telescope-theme-base, will you accept that as a solution?

SachaG:
@queso yeah let's do it this way for now.

xavcz:
You can do this easily by extending the components on Nova now 😃

PR #1050, opened 2015-07-01
make useraccounts:unstyled optional

jshimko:
Fixes https://github.com/TelescopeJS/Telescope/issues/1048

This makes it so `useraccounts:unstyled` can be removed and replaced with one of the other styled options.

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -38,6 +38,11 @@ telescope:theme-base
 telescope:theme-hubble
 telescope:update-prompt
 
+
+############ Atmosphere Packages (Optional) ############
+useraccounts:unstyled
+
+
 ############ Your Custom Packages ############
 
 # my-custom-package
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -154,7 +154,7 @@ twitter@1.1.4
 ui@1.0.6
 underscore@1.0.3
 url@1.0.4
-useraccounts:core@1.8.1
+useraccounts:core@1.11.1
 useraccounts:unstyled@1.8.1
 utilities:avatar@0.7.12
 webapp@1.2.0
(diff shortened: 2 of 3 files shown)


SachaG:
Are there any downsides to having non-Telescope packages in there? What happens if someone removes `useraccounts:unstyled`?

jshimko:
It would surely break stuff, but I don't see any other way to allow users to choose one of the other useraccounts styling packages.  And if someone is doing that manual removal step, it's probably safe to assume it was intentional and they know what they're doing, right?  (or at least know how to fix it)

We could tie it to one of the themes like Josh suggested, but then people would still be stuck with the package if they wanted to use the theme, but not the unstyled useraccounts templates.

SachaG:
Well, let's think about it. Down the road one of our goals is to make it possible to drop in Telescope into any existing Meteor app. So maybe what we should do instead is separate `telescope:users` into `telescope:users` and `telescope:accounts`, and `useraccounts:unstyled` would be a dependency of `accounts`, not `users`? This way you'd only have to customize `accounts`. 

Doing it your way is easier of course, but I liked the idea of only having `telescope:xyz` packages…

jshimko:
What would go in `telescope:accounts`?

SachaG:
Basically all the `useraccounts` stuff.

SachaG:
Actually we can't use your solution, because it would break people who update using `meteor update`… But that's true for my solution, too. I guess for now we have to keep the same package list.

But we could make `'useraccounts:unstyled` a dependency of `telescope:users` and not `telescope:lib`, that would already be an improvement.

queso:
I am not sure I follow, what breaks on meteor update? You might get a newer version of useraccounts that might break things?

SachaG:
@queso what will break is that people won't be able to update their app with just `meteor update`, since they would have to manually add `useraccounts:unstyled` to their package file.

queso:
Why not move it to theme-base? It seems like I am totally going off reservation and not using that, but most will at least rely on theme-base for their custom theme, right?

SachaG:
Yes, that's also an option.

PR #1053, opened 2015-07-01
fixing user invite

badibouteraa:


SachaG:
Good catch! Thanks :)

SachaG:
By the way, please make future PRs to the `devel` branch :)

PR #1067, opened 2015-07-11
Title Links on Avatars

mpauly:
Added a title attribute to the avatar links to give an overview who has posted on a topic even if no avatars are available. This should display the Displayname of a user on hovering over an avatar.


SachaG:
Great idea. Actually, maybe we could even have small tooltips like on Product Hunt?

PR #1076, opened 2015-07-18
Improve newsletter

sungwoncho:
Trim HTML-stripped version of the post body, because trimming `post.body` gives unexpected results.

e.g.

If post.body was `### Title\n\n * list item 1 * list item 2`, it appears in the newsletter as  `<h3 id="title-list-item-">Title * list item…</h3>`

### DIFF
--- a/packages/telescope-email/lib/server/templates/emailWrapper.handlebars
+++ b/packages/telescope-email/lib/server/templates/emailWrapper.handlebars
@@ -91,6 +91,22 @@
           border-bottom: 1px solid #ddd;
           padding: 10px 0;
         }
+        .post-comment-excerpt {
+          padding: 10px;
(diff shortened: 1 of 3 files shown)


sungwoncho:
Cannot figure out how to display avatars because I cannot use `avatar` template in the server side rendering (see https://github.com/meteor-utilities/avatar/issues/3). 

Don't want to duplicate it. Any suggestions?

SachaG:
Thanks, sorry it took me so long to merge this! I ended up modifying the styling a bit, I hope you don't mind :)

sungwoncho:
Not at all. This PR does not include displaying of avatar for commenters.

SachaG:
Yeah. We can add that once the avatar package supports it.

PR #1091, opened 2015-08-01
Show Share button on desktop version

valgalin:
Via CSS, I've showed the share button and hide the discuss icon on the post's page. Basically, the share button replaces the real estate of the discuss icon because we are already on the post's page, and we still have the comment link/count in the post's meta.

### DIFF
--- a/packages/telescope-theme-hubble/lib/client/scss/modules/_posts.scss
+++ b/packages/telescope-theme-hubble/lib/client/scss/modules/_posts.scss
@@ -41,6 +41,17 @@
 	font-size: 22px;
 }
 
+.single-post {
+	@include medium-large {
+		.post-discuss {
+  			display: none;
+		}
+		.post-share {
+  			display: block;
+		}
+	}
+}
+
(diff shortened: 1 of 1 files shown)


SachaG:
Actually after thinking about it I'm still not 100% happy with that solution… I feel like it's weird to reuse almost the same UI element, but then change only one component of it.

valgalin:
Okay, I understand :)

On Wed, Aug 12, 2015 at 2:23 PM, Sacha Greif <notifications@github.com>
wrote:

> Actually after thinking about it I'm still not 100% happy with that
> solution… I feel like it's weird to reuse almost the same UI element, but
> then change only one component of it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/TelescopeJS/Telescope/pull/1091#issuecomment-130184549>
> .
>



-- 
*Regards,*
Val Galin

*Let's be friends! :)*
t: @valgalin <http://twitter.com/valgalin>
g+: +ValGalin <https://plus.google.com/+ValGalin>

PR #1093, opened 2015-08-01
Improve search behavior

saimeunt:


saimeunt:
Quick improvement of the search behavior : when clearing the search field, redirect to the previous route.
Needs to be code reviewed and quality controlled but otherwise looks good to me.

PR #1098, opened 2015-08-07
Correctly get url for sitemap using slug

srkirkland:
This fixes the current error when using the telescope-sitemap package by using the Posts.getLink(post) function that all of the routers use, plus getting the slug field for all posts makes the url's pretty & SEO friendly.


SachaG:
Thanks!

PR #1100, opened 2015-08-07
Added a RSS route that returns posts filtered by category

lewisnyman:
Hey!

I'm pushing posts from Telescope into Hipchat via Zapier RSS. We wanted to have categories post into specific Hipchat rooms instead of all into the same room. We've also had comments that people would like to select categories to subscribe to in the newsletter.

Hope this helps! Let me know if you have any feedback. I'm not familiar with Meteor.

### DIFF
--- a/packages/telescope-rss/lib/server/routes.js
+++ b/packages/telescope-rss/lib/server/routes.js
@@ -20,6 +20,16 @@ Meteor.startup(function () {
       where: 'server'
     });
 
+    // Categories RSS
+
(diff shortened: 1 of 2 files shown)


SachaG:
Very cool! We should do the same for the API. 

And yeah, per-category newsletter subscription would be nice, but it would add a lot of complexity to the MailChimp integration. Basically for n categories, we'd need to create n! MailChimp email groups for all the permutations…

PR #1102, opened 2015-08-10
make sure the postView is a function

srkirkland:
otherwise it'll hard crash on an unknown value, like "daily"


SachaG:
Thanks! By the way, the sitemap package is actually disabled by default because it had a problem, and also because sitemaps are apparently not that useful anymore? So I might take it out of the core eventually.

srkirkland:
Interesting, I like sitemaps and my boss definitely does because of the SEO.  I noticed it was disabled by default but I'm glad it's at least an option to add.

SachaG:
Well, I heard they weren't that useful from a SEO perspective. But that may be wrong.

PR #1106, opened 2015-08-14
don't show future posts in daily view

srkirkland:
currently posts scheduled for the future still show up in the daily view for regular users, which is undesirable. This checks to see if the current time is before the end of the day in question, and if so restrains the query to cap itself to only posts before the current moment.

Not the prettiest but it works and was the best I could think of on short notice :)

### DIFF
--- a/packages/telescope-daily/lib/client/templates/posts_daily.js
+++ b/packages/telescope-daily/lib/client/templates/posts_daily.js
@@ -12,11 +12,6 @@ Meteor.startup(function () {
       return daysArray;
     },
     context: function () {
-      //if the post is in the future and we aren't showing future, cap at current moment. Else use usual end of day
-      var beforeDate = !Posts.views.baseParameters.showFuture && moment().isBefore(moment(this.date).endOf('day')) ?
-              moment().toDate()
-            : moment(this.date).endOf('day').toDate();
-            
       var context = {
         terms: {
           view: "singleday",
--- a/packages/telescope-daily/lib/client/templates/posts_daily.js
+++ b/packages/telescope-daily/lib/client/templates/posts_daily.js
@@ -12,6 +12,11 @@ Meteor.startup(function () {
       return daysArray;
     },
     context: function () {
+      //if the post is in the future and we aren't showing future, cap at current moment. Else use usual end of day
+      var beforeDate = !Posts.views.baseParameters.showFuture && moment().isBefore(moment(this.date).endOf('day')) ?
+              moment().toDate()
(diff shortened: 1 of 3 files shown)


SachaG:
Oh, is the problem that it shows future posts for the current day? 

I feel like that's probably not the best place to fix this though. Maybe it should go in https://github.com/TelescopeJS/Telescope/blob/master/packages/telescope-posts/lib/parameters.js instead? This way it'll be fixed on both the client and server.

srkirkland:
Yup that's the problem. I'll take a look in the parameters.js file and fix the PR if I can make it work there instead.

SachaG:
Here's the issue: 

https://github.com/TelescopeJS/Telescope/blob/master/packages/telescope-posts/lib/parameters.js#L40-L42

There's already a check to prevent future posts form showing up, but it's overridden if `postedAt` is already specified (which it is in this case). 

So we need to check even when `postedAt` is specified, and cap its `after` to the current time, or something.

srkirkland:
@SachaG ok I fixed it up & this works better now since it also handles singleday. The trick was the "scheduled" view was not sending along the showFuture param so I added it to that view.

Otherwise I just updated the parameters.js code to only hide future posts if showFuture is not set, and I'm careful not to replace any existing postedAt params that might exist like $gte by using _.extend.

SachaG:
Nice! But if you don't mind, I might wait until this next version is released to merge it in. I feel like it's the kind of change that probably should be tested :)

srkirkland:
Sounds good to me, thanks!

SachaG:
Sorry for the delay, but I'm finally taking another look at this. It seems to me though that with the current code, `postedAt.$lte` would always end up being equal to the current timestamp? How does that work with showing posts from say, two days ago?

SachaG:
Seems like what we want to do is something like this?

```js
  if (!parameters.showFuture) {

    if (!!parameters.find.postedAt) {
    
      if (!!parameters.find.postedAt.$lt) {

        // if postedAt.$lt is defined, use it or current date, whichever is earlier in time
        var lt = parameters.find.postedAt.$lt;
        lowerThan = lt < now ? lt : now;
      
      } else {

        // if postedAt.$lt doesn't exist, use current date
       lowerThan = now;

      }

    } else {

      // if postedAt doesn't exist at all, set it to {$lt: now}
      parameters.find.postedAt = { $lt: now };

    }

  }
```

SachaG:
OK, I'll use this for now, let's see if it works :)

PR #1107, opened 2015-08-14
fix a duplicated source file: i18n/en.i18n.json

acidsound:
please check this out :)


SachaG:
Oh, good catch, but that code is changing anyway. To avoid submitting PRs on old code please always make them to the `devel` branch :)

PR #1121, opened 2015-09-03
Use abstraction of adminUsers consistently

yourcelf:
A method to fetch users who are admins is defined in [telescope-users/lib/helpers.js](https://github.com/TelescopeJS/Telescope/blob/88b5afc65047300999ddb961492c095be3d886ac/packages/telescope-users/lib/helpers.js#L277-L279), but it wasn't used consistently.

Update ``telescope-notifications`` and ``telescope-post-by-feed`` to use the abstraction rather than querying for isAdmin directly.  To support their use, add an ``options`` parameter to ``Users.adminUsers`` that can be used to provide field/limit options to the lookup.

### DIFF
--- a/packages/telescope-notifications/lib/callbacks.js
+++ b/packages/telescope-notifications/lib/callbacks.js
@@ -5,7 +5,7 @@
 // add new post notification callback on post submit
 function postSubmitNotification (post) {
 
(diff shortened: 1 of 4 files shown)


SachaG:
Good idea, thanks for the PR!

Issue #1079, opened 2015-07-24
Removing deleted user from Mailchimp list


agnitin:
If a user is deleted, it is not automatically un-subscribed from the mailchimp list. If the user tries to register again with the same email, there is a login error that user is already subscribed. Think it would make sense to un-subscribe the user automatically on deletion

jms301:
The pull request does still leave the situation where an e-mail address is changed, leaving the mailchimp newsletter with the old one. Presumably in that situation the old address should be removed and new one then added. 

You'd need a 'onUserChange' hook to do this cleanly though, wasn't sure if that was warranted.

capensisma:
> The pull request does still leave the situation where an e-mail address is changed, leaving the mailchimp newsletter with the old one

No look the commented code at the end, It takes all cases when the user updated the email and even unsub

xavcz:
Hey, folks! Telescope has changed a lot in the past months, and there is a new React-based version, [Telescope Nova](https://github.com/TelescopeJS/Telescope) ([intro video](https://www.youtube.com/watch?v=_cLv6fH-PSQ&index=1&list=PLBoa_Q6hVeSzChsSi8X8x3FQ8NN2D73fB)). If you know Crater, it runs on the latest version.

The Legacy version is no longer supported. I'm closing this issue.

I invite you to check [the new version](https://github.com/TelescopeJS/Telescope), it's awesome!

Happy hacking! 🚀

Ps: by the way, don't hesitate to join the [Slack](http://slack.telescopeapp.org) to ask any question, you are welcome! 👍

PR #1140, opened 2015-09-13
Mailchimp newsletter enhancement - remove deleted users, edit setting in user profile. 

jms301:
As per [issue](https://github.com/TelescopeJS/Telescope/issues/1079).

This change crate a new hook "onRemoveUserAsync".


jms301:
Ok have now added in user config settings and a hook for when the user config changes. New hooks are: 

onRemoveUserAsync
and
onUpdateUserAsync

capensisma:
Oh nice but can please keep the names format as used in telescope docs

userAccountRemoved

and

profileUpdateSync

Thanks

if we keep it like that we can fix and improve plenty of things

jms301:
I don't see these hooks in the docs. I can only find [this](http://docs.telescopeapp.org/docs/callback-hooks) list of hooks, are there more somewhere else?

SachaG:
Thanks for the great work. I'll probably wait until the next release to merge this in though, just to avoid having even more stuff change at the same time. Hope that's ok?

And yeah, we should have a standard syntax for defining hooks names. `onRemoveUserAsync` is not a bad name, but it's true it doesn't really follow the same pattern as the other ones. If we wanted to be consistent, it would be `userDeleteAsync`. I'm happy to discuss this though.

jms301:
Ha sure I was just blatting issues. 

cool yea that is a better name for it. I don't mind what the name is just nice if there is a scheme. I'll update to your suggested names.

What for the update hook? userUpdateAsync?

SachaG:
maybe `userEditAsync`?

PR #1141, opened 2015-09-14
Update/Remove/Add Email in Mailchimp if profile updated

capensisma:
Update/Remove/Add Email in Mailchimp when the user profile is updated(email is updated or user sub/unsub from the newsletter

it will delete removed users from mailchimp list as well!

though two hooks need to be defined


SachaG:
@jms301's PR seems to be adressing the same issue and is more complete, so I'll probably use that one?

https://github.com/TelescopeJS/Telescope/pull/1140

PR #1142, opened 2015-09-15
fix showing user_downvoted_posts if not enableDownvotes

raduchiriac:
If `enableDownvotes` are enabled, the Downvotes block should not be shown on the user's profile page.


SachaG:
I think you're missing a `getSetting` in there? Anyway I'll add this feature to the next release :)

PR #1143, opened 2015-09-15
Fix phone home performance

queso:
Only gather up stats for phoning home when we are actually going to do it.


SachaG:
Good idea! Btw, please make sure your PRs are to the `devel` branch if you can :)

PR #1145, opened 2015-09-19
Allow admins to set the author of comments

yourcelf:


SachaG:
I'll wait a little before merging this because I don't think we want the user selection control appear in the comment submission form as well? Not sure how to best deal with that yet. Maybe we need an option to specify that a form control should only appear in the submit and/or edit form?

PR #1155, opened 2015-10-01
replaced getUserName with getDisplayName for comments

Scott-Hutcheson:
This is to ensure that in notifications, those users who do not have a username,
are still displayed using their display name.

This is required because certain users are missing user names (caused by the usage of
the social sign-up options which does not create a username)


SachaG:
Thanks!

PR #1159, opened 2015-10-05
Fix e-mail template overrides by adding the "custom" prefix server-side

garyttierney:
Separate template replacement from adding the initial "custom" prefix,
so that server-side templates (namely e-mails) can also have overrides.


garyttierney:
Bit of context to this: the `custom_` prefix is only ever registered on the client. Since `telescope-email` manages templace replacements itself it never finds the prefix.

To reproduce, create and add a new "custom_emailWrapper.handlebars" to a package, any changes made won't be reflected on your server.

The solution is to move `Telescope.config.addCustomPrefix("custom_");` to a JS file required on both the client and server.

SachaG:
Makes sense, thanks!

PR #1164, opened 2015-10-11
i18n.t bg translation + adding missing ones

durrrr91:


SachaG:
Thanks a lot! BTW, please use LingoHub to submit translations in the future :) See http://docs.telescopeapp.org/docs/internationalization

PR #1167, opened 2015-10-14
Fixing `(error.error === 603)` always results false

hoihei:
`if (error.error === 603) {...}` is always resulting false even error.error is "603" and the codes insides is never executed. 
It is because error.error is string "603".
It is need double quote to be  `if (error.error === "603") {...}`


SachaG:
Thanks :)

Issue #1170, opened 2015-10-20
Empty post-thumbnail-preview is shown broken image and the position is unpleasent


hoihei:
Environment : 
Firefox 41 for Linux (Ubuntu 14.04)

Steps:
1. Go to Submit Page

Expected Result : 
Clear empty post-thumbnail-preview placement .

Result :  
Empty post-thumbnail-preview is shown broken image.
it is lay out closing its below fields when I set 858px for thumnail width

![screenshot from 2015-10-21 01 16 01](https://cloud.githubusercontent.com/assets/327342/10615488/b0946820-7792-11e5-8de3-487a20547499.png)

SachaG:
This doesn't seem like a critical bug, so I probably won't have time to address it right away. PRs are always welcome though :)

bharatjilledumudi:
Just add an empty alt attribute for <img tag. That will fix it.

xavcz:
Hey, folks! Telescope has changed a lot in the past months, and there is a new React-based version, [Telescope Nova](https://github.com/TelescopeJS/Telescope) ([intro video](https://www.youtube.com/watch?v=_cLv6fH-PSQ&index=1&list=PLBoa_Q6hVeSzChsSi8X8x3FQ8NN2D73fB)). If you know Crater, it runs on the latest version.

The Legacy version is no longer supported. I'm closing this issue.

I invite you to check [the new version](https://github.com/TelescopeJS/Telescope), it's awesome!

Happy hacking! 🚀

Ps: by the way, don't hesitate to join the [Slack](http://slack.telescopeapp.org) to ask any question, you are welcome! 👍

PR #1187, opened 2015-10-31
Fixing issue #1170

hoihei:
Add an empty alt attribute for <img tag , to fixing broken empty image placement in firefox
as suggested by @bharatjilledumudi


SachaG:
Thanks!

PR #1204, opened 2015-11-19
Cached Embedly Thumbnail template

capensisma:
It checks if there is a cached version in the file if yes it shows it if not it takes the one from embedly


SachaG:
Actually I think we'll just change `thumbnailUrl` and keep the original non-cached image in `originalThumbnailUrl`. This way we don't need to change the template.

PR #1207, opened 2015-12-09
upvote/downvote callbacks include user as second argument

AlexeyMK:
Should be backwards compatible due to javascript's hand-wavy treatment of argument counts in functions.


SachaG:
Actually it's better if you make PRs to the `devel` branch. So I'll close this one and reproduce the change manually :)

AlexeyMK:
cool! I was trying to find a contributing.md or equivalent and could not.
might be worth having those instructions explicit somewhere.

SachaG:
Good point.

PR #1210, opened 2015-12-12
allow operations on editable blackbox schema fields

Mostafa-Samir:
**The Issue**

When there's an editable blackbox field in the schema, it's contained in the [allowedFields](https://github.com/TelescopeJS/Telescope/blob/master/packages/telescope-lib/lib/collections.js#L51) array. However, any property of that field, which can be anything and not defined in the schema, will not be listed in that array. So if a field in the modifier is a property of that blackbox field, the operation will be rejected by the <code>allowCheck</code> function as that field will still be there after executing <code>_.difference(fields, allowedFields)</code>.

**The Solution**

Before checking against the <code>allowedFields</code>, the <code>fields</code> are filtered out of any field whose name has a prefix that matches the name of one of the editable blackbox fields that exist in the schema.

### DIFF
--- a/packages/telescope-lib/lib/collections.js
+++ b/packages/telescope-lib/lib/collections.js
@@ -49,13 +49,27 @@ Telescope.allowCheck = function (collection, userId, document, fieldNames, modif
   var schema = collection.simpleSchema();
   var user = Meteor.users.findOne(userId);
   var allowedFields = schema.getEditableFields(user);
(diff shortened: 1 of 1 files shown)


Mostafa-Samir:
created a new pr #1241 where the base branch is `devel` instead of `master` as per the new contribution guidelines in 85e9d00

PR #1217, opened 2015-12-27
Fix Prerender https host issue

nobutakaoshiro:
I found that the `host` become incorrect value If including https in `siteUrl`.

### DIFF
--- a/packages/telescope-prerender/lib/server/prerender.js
+++ b/packages/telescope-prerender/lib/server/prerender.js
@@ -1,11 +1,12 @@
 if (Settings.get('prerenderIOToken')) {
-  
+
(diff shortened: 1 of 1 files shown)


SachaG:
Thanks! By the way, next time if you can try to make the PR to the `devel` branch :)

SachaG:
I added your fix here: https://github.com/TelescopeJS/Telescope/commit/a2c77429e05fe558ff38e966e5489de3a50b0483 :)

nobutakaoshiro:
@SachaG Thanks! 

> By the way, next time if you can try to make the PR to the devel branch :)
Okey! I'll try to make the PR the devel branch next time :smiley:

PR #1220, opened 2015-12-28
Attach data to post for spam detection package.

queso:


SachaG:
The underscores look weird… is there a specific reason why the property names are not in camelCase?

PR #1221, opened 2015-12-28
Unblock publications

queso:
Publications are blocking and there is really no reason to, making the post publications synchronous doesn't help anything. I used the `meteorhacks:unblock` package and it seems to be helping load times on Crater so far.

Hattip to @arunoda for his help on this.

### DIFF
--- a/packages/telescope-posts/lib/methods.js
+++ b/packages/telescope-posts/lib/methods.js
@@ -184,6 +184,9 @@ Meteor.methods({
       post.userId = user._id;
     }
 
+    post.user_ip = this.connection.clientAddress;
+    post.user_agent = this.connection.httpHeaders["user-agent"];
+
     return Posts.submit(post);
   },
 
--- a/packages/telescope-posts/lib/posts.js
+++ b/packages/telescope-posts/lib/posts.js
@@ -202,6 +202,21 @@ Posts.schema = new SimpleSchema({
     optional: true
   },
   /**
+    Save info for later spam checking on a post. We will use this for the akismet package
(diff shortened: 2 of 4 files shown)


SachaG:
Oh cool! So I guess we don't need the other PR then?

queso:
Oops, I guess not, if you want to accept both. My branching skills lacked for this, haha.

SachaG:
What about the underscore thing?

SachaG:
I'm changing this to `userIP` and `userAgent`.

SachaG:
I'll wait a little before merging this to `master`, I was talking with @tmeasday about unblock and he pointed me to this thread: https://github.com/meteor/meteor/issues/853

But I don't have the time to consider the pros and cons right now. Hope having it on devel works for you for now?

queso:
Devel works fine. Honestly, I am not sure how much longer I will keep up with telescope going forward... Might just be easier to tweak the site here and there, the all encompassing architecture sometimes makes it hard to do things. Like the upvote stuff is slow because all the callbacks that do DB lookups, most of them looking up the same data over and over.

I guess I saw a few small use cases @tmeasday, but it looks like @arunoda responded to them... Personally, I've seen a big first time render performance increase and don't plan to go back to blocked pubs...

tmeasday:
For the record, as I posted to that thread, I don't think I have any concerns about unblocking publications any more.

arunoda:
As I said in the other post, I will publish a new version of unblock. Which
unblock all publications by default.

PR #1223, opened 2015-12-31
supply default email based on 3rd party login, if possible

jjwon0:
Most sites, when you log in with a 3rd party OAuth provider, will fill in whatever possible fields when making a new account.  Mirroring the capability here.

I was unable to test with Weibo due to not having a Chinese phone number, but I tested with the other external services mentioned [here](http://docs.meteor.com/#/full/meteor_loginwithexternalservice).  The ones I didn't add support for do not provide the email address back by default.


SachaG:
Oh, great idea, thanks!

PR #1229, opened 2016-01-08
Set counter name to category id instead of category slug

rizafahmi:
If using slug, when we edit the category, the slug changes and the counter
become 0 until we refresh the browser. I think it's better if we stick with
_id because _id is not changing.


SachaG:
Good idea :)

PR #1236, opened 2016-01-23
fixed <a> href in post_author template

christabella:
Link to author profile wasn't working (refer to meta.telescopeapp.org to see broken link). Just corrected it to match with the post_avatars template, which *was* working.

P.S. If anyone can explain what was going on I'd like it because I still don't really understand


SachaG:
Thanks, I'll merge it in manually :)

PR #1241, opened 2016-02-08
allow operations on editable blackbox schema fields

Mostafa-Samir:
**The Issue**

When there's an editable blackbox field in the schema, it's contained in the [allowedFields](https://github.com/TelescopeJS/Telescope/blob/master/packages/telescope-lib/lib/collections.js#L51) array. However, any property of that field, which can be anything and not defined in the schema, will not be listed in that array. So if a field in the modifier is a property of that blackbox field, the operation will be rejected by the <code>allowCheck</code> function as that field will still be there after executing <code>_.difference(fields, allowedFields)</code>.

**The Solution**

Before checking against the <code>allowedFields</code>, the <code>fields</code> are filtered out of any field whose name has a prefix that matches the name of one of the editable blackbox fields that exist in the schema.

### DIFF
--- a/packages/telescope-lib/lib/collections.js
+++ b/packages/telescope-lib/lib/collections.js
@@ -49,13 +49,27 @@ Telescope.allowCheck = function (collection, userId, document, fieldNames, modif
   var schema = collection.simpleSchema();
   var user = Meteor.users.findOne(userId);
   var allowedFields = schema.getEditableFields(user);
+  var blackboxFields = schema.getBlackboxFields();
+  var allowedBlackboxes = _.intersection(blackboxFields, allowedFields);
   var fields = [];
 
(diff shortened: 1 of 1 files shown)


xavcz:
Hey @Mostafa-Samir, Telescope has changed a lot in the past months, and there is a new React-based version, [Telescope Nova](https://github.com/TelescopeJS/Telescope) ([intro video](https://www.youtube.com/watch?v=_cLv6fH-PSQ&index=1&list=PLBoa_Q6hVeSzChsSi8X8x3FQ8NN2D73fB)). If you know Crater, it runs on the latest version.

The Legacy version is no longer supported. Thanks for your PR, however I'm closing it.

Happy hacking! 🚀

PR #1242, opened 2016-02-09
Add siteUrl in front of action link. The link is wrong

rizafahmi:
Current invite link is broken.

airmail 2016-02-09 18-36-54

Maybe it's because without `siteUrl`. So I just add it at the beginning of the link.


pbmarcano:
I'd very much prefer this PR go through and fix this... please and thanks!

SachaG:
Thanks, merged. But make sure you make future PRs to the Nova branch :)

PR #1257, opened 2016-03-02
Add PostViews to loading & no posts scenarios; fix paths on Links generation …

paulmolluzzo:
I added the `PostViews` component to the loading and not found scenarios for the `PostList` because it's more visually consistent across slow loading, and if you don't have it when there are no posts you're sort of stuck without a way out.

Also, the links were broken so I fixed that. I'm guessing that FlowRouter method was deprecated...?

### DIFF
--- a/packages/base-components/lib/posts/list/PostList.jsx
+++ b/packages/base-components/lib/posts/list/PostList.jsx
@@ -15,13 +15,19 @@ const PostList = props => {
   } else if (!props.ready) {
     return (
       <div className="postList">
-        <PostsLoading/>
+        <PostViews />
+        <div className="post-list-content">
+          <PostsLoading/>
+        </div>
       </div>
     )
   } else {
     return (
       <div className="postList">
-        <NoPosts/>
+        <PostViews />
+        <div className="post-list-content">
+          <NoPosts/>
+        </div>
       </div>
     )  
   }
--- a/packages/base-components/lib/posts/list/PostViews.jsx
+++ b/packages/base-components/lib/posts/list/PostViews.jsx
@@ -12,7 +12,7 @@ const PostViews = props => {
       <h4>Sort By:</h4>
       <ul>
         {views.map(view => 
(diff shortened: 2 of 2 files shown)


SachaG:
Weird, `FlowRouter.extendPathWithQueryParams` should work. The reason I don't just use the `/` route is that if you're already sorting by category, I don't want to undo the category sort when you also filter the list by view. So this method extends the query params without reseting them.

SachaG:
I'll merge in the other change manually though.

paulmolluzzo:
Instead of manually merging I can just use the current path and append the params. Not including that part breaks the links once you've clicked on any of the sorting links. Should I do that or is it too late?

paulmolluzzo:
Also, I can't find anything about `FlowRouter.extendPathWithQueryParams`. Am I overlooking it in the docs?

SachaG:
`FlowRouter.extendPathWithQueryParams` is defined here: 

https://github.com/TelescopeJS/Telescope/blob/nova/packages/nova-core/lib/router.jsx#L3-L8

paulmolluzzo:
Well, even still this extends and not replaces, which seems more appropriate. And if you do want to extend and then apply multiple params there's something still wrong.

There's a new PR more specific to the URLs: https://github.com/TelescopeJS/Telescope/pull/1259

SachaG:
Well the intended behavior is definitely to extend and not replace. So if that's not working properly maybe we need to fix the `extendPathWithQueryParams` function? That being said this is part of the components anyway, so people are supposed to customize it to their needs. So we can probably use either behavior as long as it works properly.

paulmolluzzo:
So you want to be able to have both "best" and "rejected" applied at the same time? Using extend allows two or more to be applied at the same time and some seem to conflict internally.

SachaG:
Not best and rejected since they're both views, but best and a category, or a category and dates, etc.

SachaG:
But anyway I see the issue, I'll take a look and either fix my function or change the behavior :)

SachaG:
merged this manually.

PR #1258, opened 2016-03-04
update alt:react-accounts for full SSR

emmostrom:
remove conflicting blaze accounts-ui
support NPM React with Meteor packages requiring react-runtime


SachaG:
Thanks! btw where do you use the `react-runtime` package from?

emmostrom:
the alt:react-accounts* packages use the meteor react package
the meteor react package uses react-runtime
and react-runtime includes the React js files.

when in 1.3 you use React from npm instead of the meteor react package you now have two copies of React and it doesn't like that.

The react-runtime package I included acts as an override to the meteor react-runtime and instead of including a second copy of React is just exposes the NPM one from your app.

SachaG:
Makes sense. But I wonder, could we get rid of the Meteor react package altogether then?

emmostrom:
alt:react-accounts* needs react-runtime to be backwards compatible with 1.2

I think I read somewhere that the final 1.3 will do something like this automatically - so this react-runtime patch may just be temporary until that happens.

PR #1260, opened 2016-03-04
[Nova] Update share package

paulmolluzzo:
This is an update of the `nova:share` package to use React, be a little more abstract, and fix some related dependencies. I've also added this package and it's being used on a single Post page so you can see it in action. It's pretty much the same behavior as before, I think. (Toggles open and closed, opens share in new window), but the styles are a little different to meet the use case in this repo.

There are a couple changes I would make if this goes through OK:
* Pass in an array of networks to decide what you can share on. If you don't want LinkedIn you could omit it, if you want Reddit you can add it.
* Add more social networks. HN, Reddit, Pinterest, Buffer, Tumblr, etc. See another repo I have here: http://paulmolluzzo.github.io/simple_share_popups/

LMK what you think.


SachaG:
Very nice :) Maybe the next step is adding some meta tags back in so that pages shared have the proper metadata?

paulmolluzzo:
I'm wondering if that should be a separate package or not. I think FB/Twitter meta tags should be here, but titles etc no. There should be a common mechanism for adding tags though I think, so something should be for "tag creation" and something in this package would be "use tag creation to create these tags". Does that sound right or am I over complicating it?

SachaG:
Oh yeah I didn't mean as part of the share package, I think it should be independent. As for the mechanism, how about a hook to add code to a page and/or route's loading trigger?

paulmolluzzo:
https://arianasiresearch.files.wordpress.com/2015/04/7453018.gif?w=660

PR #1264, opened 2016-03-05
update to version of alt:react-accounts* that doesn't depend on react-runtime

emmostrom:
New version doesn't have a dependency on the Meteor react-runtime package so works well with the NPM react version.

There are other packages (utilities:react-list-container, utilities:react-form-containers) in Telescope that do use the react-runtime so until those are updated the local react-runtime package is still needed as an override to remove the duplicate React issue.

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -40,3 +40,4 @@ base-components
 base-styles
 
 nova-demo
+alt:react-npm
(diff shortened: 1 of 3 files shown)


SachaG:
Great, I'll work on my end to remove the other React dependencies then :)

PR #1270, opened 2016-03-08
HEAD Tags component

xavcz:
I used [react-helmet](https://github.com/nfl/react-helmet) in order to add HEAD tags in an easy way.

Currently, it's static and placed in the Layout component

A good thing for SEO would be to change dynamically the metadata.

React-helmet's says that duplicate tags are overwritten, so a solution might be to have the generic one in the Layout, and specific one in Post? It would imply to duplicate code, it may not be the best solution.

Else, there is the solution of a hook triggered on page loading / route changing (cf #1260).

Which way would be the best, what do you think? 
Ping @SachaG @paulmullozzo who may have thoughts on this subject :dart:


SachaG:
I think a hook system would work, but we can probably use something like this in the meantime. 

I'd probably not define yet another `settings` object though, I feel like it's a bit confusing to have that extra layer.

SachaG:
Apparently having the `<HeadTags>` component in there messes with FlowRouter. To reproduce, go to `?view=best` and click any post edit link. It will redirect you to the homepage :(

PR #1279, opened 2016-03-17
fixing typo in newCommentSubscribed notification

AlexeyMK:
Without this, the following bug would happen:

When a user adds a comment to a post with a subscriber that is not otherwise notified, the notification doesn't get sent and the server repots the following error: "Exception in defer callback: Error: [Can't start Herald.escalate: TypeError: undefined is not a function]" when trying to notify a subscriber of a comment.


SachaG:
Thanks :)

PR #1280, opened 2016-03-21
Nova excerpt : short summary of the htmlBody

capensisma:
A short version of the htmlbody that will help the performance when loading all items in the listing and also will be useful for meta description and we could even allows users to fill the excerpt when they are posting ...

### DIFF
--- a/packages/base-components/lib/routes.jsx
+++ b/packages/base-components/lib/routes.jsx
@@ -40,6 +40,22 @@ FlowRouter.route('/posts/:_id', {
   }
 });
 
+FlowRouter.route('/p/:slug', {//short link for posts
+  name: 'posts.single',
+  action(params, queryParams) {
+    ({AppContainer, DocumentContainer, Post} = Telescope.components);
+    mount(AppContainer, {content: 
+      <DocumentContainer 
+        collection={Posts} 
+        publication="posts.single" 
+        selector={params}
+        terms={params}
+        joins={Posts.getJoins()}
+        component={Post}
+      />});
+  }
+});
+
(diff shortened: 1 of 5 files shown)


SachaG:
We can't do it that way because it might cut halfway through an HTML tag (i.e. `<strong>hello world</stro`).

Not that easy… ;)

capensisma:
fixed with the Telescope.utils.trimHTML

SachaG:
Nice! One last suggestion: maybe we should name the field `htmlExcerpt` or `htmlBodyExcerpt` so it's clearer what it's an excerpt from?

Also you can add the new field to `Posts.publishedFields.list` and remove `htmlBody` :)

capensisma:
Thanks :)

I think we should keep just excerpt so there will be no confusing since excerpt now doesn't' contain any html code at all! and is just a short version of the body! so it's just post.excerpt ...

I will swap it with htmlBody in the publisheFields...

also after we could check and see if the user has filled excerpt before filling after the post submit/edit ....

SachaG:
Right. I forgot that `trimHTML` removes HTML tags. BTW, I think you're using it wrong? It should be `trimHTML(HTMLstring, numberOfWords)`.

capensisma:
Oh, thanks

I didn't take a good look, changed it so we take only 30 words only from the text of the htmlBody as a short description...

SachaG:
Oh, I hadn't seen the extra changes. Please don't mix two unrelated features in the same PR. I'll merge the excpert stuff manually this time.

capensisma:
thanks :)

PR #1281, opened 2016-03-29
[nova] Test email + fix sample_settings.json

xavcz:
Two ways of making nova:email working.

Which way shall we take?


SachaG:
A few notes:

- Is there a specific reason to move `Email.addTemplates(emailTemplates)` from `template.js` to `routes.js`?
- It feels weird to me to have an `email/test` route for testing instead of calling a method. I feel like a route should display content, not trigger a function.
- I completely forgot about Meteor's own `Email` function when I changed email from `Telescope.email` to just `Email`, thanks for noticing this! I'm not sure if we want to monkey-patch it though, it seems easier to just have two different objects?

xavcz:
No need no more ! :+1:

PR #1284, opened 2016-03-31
Fixed issue with UserEdit.jsx and undefined EditComponent

MHerszak:
EditComponent wasn't available through Telescope.components. The import from nova:forms for UserEdit.jsx fixed it. I had the following issue
`Warning: React.createElement: 
type should not be null, undefined, boolean, or number. 
It should be a string (for DOM elements) or a 
ReactClass (for composite components). 
Check the render method of `UserEdit`.`


SachaG:
Thanks for catching that! Turns out it's enough to just add

```
import ReactForms from "meteor/nova:forms";
const EditDocument = ReactForms.EditDocument;
```

to the file, so I'll just do that for now :)

PR #1286, opened 2016-04-02
Bugfixes for std:accounts-ui

timbrandin:


packages/nova-base-components/package.js
```diff
@@ -24,8 +24,8 @@ Package.onUse(function (api) {
     'tmeasday:check-npm-versions@0.1.1',
     // 'alt:react-accounts-ui@1.2.1',
     // 'alt:react-accounts-unstyled@1.2.1',
-    // 'studiointeract:react-accounts-ui-basic@1.0.1',
-    'studiointeract:react-accounts-ui@1.0.7',
+    // 'std:accounts-ui-basic@1.0.1',
+    'std:accounts-ui@1.0.7',
```
queso: shouldn't this be 1.1.2, @timbrandin?

timbrandin:
Oh yeah, you're right

SachaG:
Thanks so much for this! Great work on the accounts stuff :)

PR #1290, opened 2016-04-05
Responsive Modals

capensisma:
it works 

https://www.dropbox.com/s/pbitniksr0pzhya/Screenshot%202016-04-05%2001.40.47.png?dl=0


SachaG:
Actually, it turns out react-bootstrap has a modal component already! So we can just use that :)

PR #1291, opened 2016-04-06
HeadTags: dochead instead of react-helmet

xavcz:
Well, seems to work better with dochead than react-helmet @SachaG :)

I wrapped the rendering in a component as recommended by the doc in order to have SSR.

The component accept props, which by default are set to the general information of the Telescope.

I removed the old one from `Header` and placed it in 3 places :
1. `PostPage` container to have access to post props.
2. `PostListHeader` which is displayed on "basic view".
3. `UserProfile` component to have access to a user data.

What do you think of the "placement"?

I'm thinking about using new utils/helpers like `Telescope.utils.getLogoUrl()` instead of `Telescope.utils.getSiteUrl() + Telescope.settings.get("logoUrl"); ` and `Posts.getPostThumbnailUrl()` instead of `post.thumbnailUrl`. It would give a shorthand for access absolute url to logo and postThumbnail (maybe get it from embedly if active?).

### DIFF
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -26,6 +26,7 @@ check@1.1.2
 chuangbo:cookie@1.1.0
 chuangbo:marked@0.3.5_1
 coffeescript@1.0.15
+cosmos:browserify@0.8.4
 dburles:collection-helpers@1.0.4
 dburles:spacebars-tohtml@1.0.1
 ddp@1.2.3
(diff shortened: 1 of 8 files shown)


SachaG:
Great! And yeah good idea about creating new helpers :)

SachaG:
Actually the current placement of the `<HeadTags/>` component adds an extra `div, which means an extra space in the Flexbox layout:



Is there any way we could do it without rendering an extra element? Or maybe move it somewhere else at least?

xavcz:
Yep, I do another PR with the headtags rendered in another place + the helpers.

PR #1292, opened 2016-04-07
Fix HeadTags <-> Flexbox + add 2 helpers for images

xavcz:
I put HeadTags outside the `post-header-list` div in order to not mess with Flexbox :+1: 

And I created the two helpers `Telescope.utils.getLogoUrl()` which returns the logo url & `Posts.getThumbnailUrl(post)` which returns the thumbnail url.

### DIFF
--- a/packages/nova-base-components/lib/posts/list/PostListHeader.jsx
+++ b/packages/nova-base-components/lib/posts/list/PostListHeader.jsx
@@ -8,13 +8,15 @@ const PostListHeader = () => {
   ({PostViews, SearchForm, CategoriesList, HeadTags} = Telescope.components)
 
   return (
-    <div className="post-list-header">
+    <div>
       <HeadTags />
-      <div className="post-list-categories">
-        <ListContainer collection={Categories} limit={0} resultsPropName="categories" component={CategoriesList}/>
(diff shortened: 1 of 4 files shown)


SachaG:
I guess that'll work if there's no better solution. Still seems a bit hacky though, I'll have to think about it.

SachaG:
I guess the current placement means pages that are not the post list won't have any meta tags though? Shouldn't it be in `Layout` instead?

SachaG:
Actually the thumbnail URL helpers break Embedly thumbnails, since their URL is like:

`"//i.embed.ly/1/image/crop?url=https%3A%2F%2Fi.ytimg.com%2Fvi%2FfJh-oaPsXnk%2Fhqdefault.jpg&key=9cc27064a8ca49798dd07a193475498c&width=200&height=125"`

SachaG:
OK, so I think I would like to have the following behavior:

- Head tags are set for every page by default. 
- They can optionally be overridden at the component level. Doing this at the route level would be more logical, but then there's probably a lot of cases where we need to subscribe first, and we can't do that in the route. 

I'd like to know how other React apps handle this?

Issue #1250, opened 2016-02-28
Exception while simulating the effect of invoking 'submitPost' TypeError: Cannot read property 'clientAddress' of undefined


Jarskih:
I am having this problem with submitPost method.
Running Windows 8.1 and Telescope 25.7 master

methods.js:
```
187: post.userIP = this.connection.clientAddress;
188: post.userAgent = this.connection.httpHeaders["user-agent"];
```

`Exception while simulating the effect of invoking 'submitPost' TypeError: Cannot read property 'clientAddress' of undefined
    at Meteor.methods.submitPost (http://localhost:3000/packages/telescope_posts.js?1669b07ebc5bde5e4e7d3997406b7df923d1d855:1226:34)
    at http://localhost:3000/packages/ddp-client.js?250b63e6c919c5383a0511ee4efbf42bb70a650f:3911:25
    at _.extend.withValue (http://localhost:3000/packages/meteor.js?9730f4ff059088b3f7f14c0672d155218a1802d4:971:17)
    at _.extend.apply (http://localhost:3000/packages/ddp-client.js?250b63e6c919c5383a0511ee4efbf42bb70a650f:3902:54)
    at _.extend.call (http://localhost:3000/packages/ddp-client.js?250b63e6c919c5383a0511ee4efbf42bb70a650f:3780:17)
    at http://localhost:3000/packages/aldeed_autoform.js?b3aa1a45a8c57d1da1124520538d2cb9fa458509:3292:13
    at runHook (http://localhost:3000/packages/aldeed_autoform.js?b3aa1a45a8c57d1da1124520538d2cb9fa458509:7734:11)
    at Object.cb (http://localhost:3000/packages/aldeed_autoform.js?b3aa1a45a8c57d1da1124520538d2cb9fa458509:7746:13)
    at Object.result (http://localhost:3000/packages/underscore.js?46eaedbdeb6e71c82af1b16f51c7da4127d6f285:794:19)
    at runHook (http://localhost:3000/packages/aldeed_autoform.js?b3aa1a45a8c57d1da1124520538d2cb9fa458509:7760:15)`

Jarskih:
Sometimes submitting new post causes this as well:
```
\AppData\Local\.meteor\packages\meteor-tool\1.1.10\mt-os.windows.x86_32\tools\utils\fiber-helpers.js:168
    }).run();
       ^
FATAL ERROR: JS Allocation failed - process out of memory
```
Not sure if related.

[...]

PR #1294, opened 2016-04-08
Quick fix for #1250

meteorplus:
Commented userIP and userAgent for now

until we find a reliable solution : since the ip can't be empty (vpn, routers...) or using other ones Forward address ... like that


xavcz:
We wrapped it around with `Meteor.isServer && this.connection` 👍  Thanks for the PR youss!

PR #1301, opened 2016-04-15
Completed profile hook

xavcz:
Reactively show a modal if the profile required fields are not completed. 

The only way to close this modal is to complete the fields or log out with the button provided.

I have used `nova:forms` + `users.edit` method and add a prop `requiredFieldsOnly` to show only `telescope.email`.

### DIFF
--- a/packages/nova-base-components/lib/common/Layout.jsx
+++ b/packages/nova-base-components/lib/common/Layout.jsx
@@ -5,12 +5,14 @@ const FlashContainer = Core.FlashContainer;
 
 const Layout = props => {
 
-  ({Header, Footer, FlashMessages, NewsletterForm, HeadTags} = Telescope.components);
+  ({Header, Footer, FlashMessages, NewsletterForm, HeadTags, UserProfileCheck} = Telescope.components);
 
   return (
     <div className="wrapper" id="wrapper">
 
       <HeadTags />
+
+      <UserProfileCheck {...props} />
 
       <Header {...props}/>
     
--- a/packages/nova-base-components/lib/components.js
+++ b/packages/nova-base-components/lib/components.js
@@ -65,6 +65,7 @@ Telescope.registerComponent("CanEditUser",          require('./permissions/CanEd
 
 Telescope.registerComponent("UserEdit",             require('./users/UserEdit.jsx'));
 Telescope.registerComponent("UserProfile",          require('./users/UserProfile.jsx'));
+Telescope.registerComponent("UserProfileCheck",     require('./users/UserProfileCheck.jsx'));
 Telescope.registerComponent("UserAvatar",           require('./users/UserAvatar.jsx'));
 Telescope.registerComponent("UserName",             require('./users/UserName.jsx'));
 Telescope.registerComponent("UserMenu",             require('./users/UserMenu.jsx'));
--- a/packages/nova-base-components/lib/users/UserProfileCheck.jsx
+++ b/packages/nova-base-components/lib/users/UserProfileCheck.jsx
@@ -17,7 +17,7 @@ const UserProfileCheck = ({ currentUser }) => {
           methodName="users.edit"
           labelFunction={ (fieldName)=>Telescope.utils.getFieldLabel(fieldName, Meteor.users) }
           successCallback={ (user)=> Telescope.callbacks.runAsync("profileCompletedAsync", user) }
-          requiredFieldsOnly={ true }
+          fields={ ["telescope.email"] }
         />
       </Modal.Body>
       <Button bsStyle="primary" className="complete-profile-logout" onClick={ () => Meteor.logout() }>Log Out</Button>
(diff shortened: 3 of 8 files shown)


SachaG:
Nice work :) Just one suggestion: instead of adding special logic in NovaForm for `requiredFieldsOnly`, maybe it could support a `fields` property that lets you specify which subset of fields you want to display? So if the edit form would normally show fields `foo`, `bar`, and `baz`, and you pass `fields=["foo", "baz", "buzz"]` it would only show `foo` and `baz` (that's how Autoform does it at least). 

Also `devel` was a bit behind, sorry about that!

xavcz:
Ooow, I like this! Yep, I'll modify following this way!

xavcz:
Don't be afraid by this huge comment, it's my changelog as a little story: this is a description for you to better understand what I did, and also for me to take a step back 👍 

c2d1a1a 
Be in accordance with `devel`, 😄  ok that was the easy part.

2baf40a
Then, I ran into several bugs due to undefined settings (I run with Nova with `settings.json`). I found that it was due to a missing `typeof` in the `Telescope.settings.get` method. Even if it's not related to this PR, I thought it wasn't useful to do a new one.

5c424da
I switched from my `requiredFieldsOnly` prop to a `fields` prop which accepts an array of fields and will filter on it the way you described me, it does the job.

💥 
The story could have be done by this time, but I realized that in some cases after submitting the modal or logging out, I got warnings displayed in my browser console.

d878b84
The first warning was related to `nova:forms` and I guess it was due to the update to React 15:
`Warning: Unknown DOM property dataType. Did you mean datatype?`. Quick typo change: 

72ea1ab
The way I handled the modal was in fact an anti-pattern! Ok it worked, but I got crazy warnings telling me I was doing shit with the state of my component! 🐑 

```Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the undefined component.```

```Warning: setState(...): Cannot update during an existing state transition (such as within `render` or another component's constructor). Render methods should be a pure function of props and state; constructor side-effects are an anti-pattern, but can be moved to `componentWillMount`.```

These warnings were caused by a really bad way of closing the modal (show prop true/false) on logout or submitting a new email. After struggling and trying a lot of different patterns, I finally wrapped the modal in a container passing to the modal to manage three different cases: 
- there is not a current user -> render null
- there is a current user who has a complete profile -> render the modal but don't show it.
- there is a current user who has no email / erased it from his profile -> render the modal and show it!

In addition to that, I added a go home callback on logout to move the user from the account route if he logs out with the modal.

**Conclusion: I just f$#§ing leveled up in Meteor and React!!!** :metal: 

It's now time to spread love and drink beers, have a good weekend! 🍻

**Comment by SachaG** [comment]
Wow that's awesome! Regarding the `datatype` thing, that's just a name I gave the prop, but maybe React treats `data*` properties differently? I don't really understand that error tbh. I'm also surprised that you needed to use React Komposer, if I have time I'll test out both ways as well. I could use some levelling up in React myself, too ;)

**Comment by SachaG** [comment]
Oh ok I understand the problem with your first solution now. As soon as you submit the form, the user profile becomes complete and the modal component was being unmounted. So that's why the modal needs to stick around wether the profile is complete or not :)

**Comment by SachaG** [comment]
So as far as I can tell, the important part is keeping the modal around, Komposer is not really required here I think.

**Comment by SachaG** [comment]
OK I simplified it a bit, let me know what you think:

https://github.com/TelescopeJS/Telescope/commit/60f29bbfd189eed2fdb355c3758053de48f5258f

Basically the requirements are: 

- If there is no user object, do not mount the form.
- If there is a user object, always mount the form.
- If the profile is incomplete, show the modal.

**Comment by SachaG** [comment]
Oh also it shouldn't be `fields={ ["telescope.email"] }`, but instead:

```
// return fields that are required by the schema but haven't been filled out yet
    var schema = Users.simpleSchema()._schema;
    var requiredFields = _.filter(_.keys(schema), function (fieldName) {
      var field = schema[fieldName];
      return !!field.required && !Telescope.getNestedProperty(Meteor.user(), fieldName);
    });
```

**Comment by SachaG** [comment]
Also if I'm not mistaken, the way you did `getFieldNames` would *only* get fields passed as `this.props.fields`:
```
const schema = collection.simpleSchema()._schema;
relevantFields = _.filter(_.keys(schema), function (fieldName) {
   return _.contains(fields, fieldName);
});
```

Instead we want the *intersection* of `this.props.fields` and whichever fields are editable/insertable: 

```
relevantFields = _.intersection(relevantFields, fields)
```

xavcz:
Yeah this way to handle the modal does the same things in a much simpler way: it is contained and control by another component without the need to compose with tracker 👌

For the fields, I guess I didn't understand the specs well, I see now. I'll update it 😅

Still gaining XP, thanks for the comments!!

PR #1303, opened 2016-04-16
fix for peerDependencies requirements!

meteorplus:
downgraded to 0.14.8
formsy is making some bad trouble here better use the nice bootstrap forms? :) please?

thanks :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,14 +11,15 @@
     "formsy-react-components": "^0.6.6",
     "js-cookie": "^2.1.0",
     "load-script": "^1.0.0",
-    "react": "^15.0.1",
-    "react-addons-pure-render-mixin": "^15.0.0",
+    "react": "^0.14.8",
+    "react-addons-pure-render-mixin": "^0.14.8",
     "react-bootstrap": "^0.28.5",
-    "react-dom": "^15.0.0",
+    "react-dom": "^0.14.8",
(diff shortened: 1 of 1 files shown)


justintime4tea:
peerDependencies warning has already been addressed in latest devel branch. 
No need to downgrade React, formsy version 0.18 will now use React@^15.0.1

`git clone https://github.com/TelescopeJS/Telescope`
`cd Telescope/`
`git checkout devel`
`npm install`
`meteor'

If you prefer the bootstrap forms feel free to create a Telescope package which replaces the functionality of NovaForm component using bootstrap forms and issue a pull request.

This would be an awesome contribution and allow users to choose between either formsy or bootstrap.

SachaG:
@justintime4tea well Formsy itself uses Bootstrap, so I don't know if we need another Bootstrap version. Also I'm not sure if React-Bootstrap by itself provides the same kind of form wrapping logic that things like Formsy or Redux-Forms do. Although I'll admit I'm still pretty new to React forms so I could be wrong.

justintime4tea:
Aw, makes sense. I wasn't familiar at all with the Bootstrap Forms the person above us was talking about. Good to know.

capensisma:
me too I don't know this person above lol :D

xavcz:
The deps issue got fixed earlier 🎉 🎋

Issue #1314, opened 2016-04-29
Add soft delete for comments


justintime4tea:
# Feature Request

Add the ability to ***`soft delete`*** comments.

**Soft Delete:**  
```
Replace body of comment with "Delete" and flag as isDeleted.
Remove ability to edit and "soft delete" once comment is flagged as isDeleted.
Reduce the posts and users comment count.
Subsequent replies to "soft deleted" comment will remain un-touched.
```

justintime4tea:
I created this so I could reference the issue ID when creating the feature branch and subsequent pull request.
Feature branch will be labeled:  feature-1314 and can be referenced back to this issue.

CollinGraves:
Can this be closed?

PR #1315, opened 2016-04-29
Complete soft delete feature for comments

justintime4tea:
# List of changes
- Add new CommentsDelete component to nova-base-components
- Add new props and functions to CommentsItem for delete feature
- Hide edit, delete and reply UI elements in CommentsItem if comment isDeleted
- RegisterCommentsDelete with Telescope namespace in components.js
- Add style for comment-delete and comments-delete-form elements in _comments.scss
- Add style for novaform-cancel element in _other.scss
- Add isDeleted check before decrementing user and post comment count
- Change cancel element from <a> to <Button> in NovaForm
- Add submitButtonText prop to NovaForm to allow changing the Submit button text
- Add check for isDeleted to Users.can.edit permissions

Commit: [c922645](https://github.com/justintime4tea/Telescope/commit/c922645350fc544ecff6b9990837a7a6a44b9303)

Closes issue: #1314

### DIFF
--- a/packages/nova-base-components/lib/comments/CommentsDelete.jsx
+++ b/packages/nova-base-components/lib/comments/CommentsDelete.jsx
@@ -1,0 +1,37 @@
+import React, { PropTypes, Component } from 'react';
+import Actions from '../actions.js';
+import NovaForm from "meteor/nova:forms";
+
+class CommentsDelete extends Component {
+
+  render() {
+    return (
+      <div className="comments-delete-form">
+        <NovaForm
+          collection={Comments}
+          document={this.props.comment}
+          currentUser={this.context.currentUser}
+          methodName="comments.deleteById"
+          submitButtonText="Delete"
+          successCallback={this.props.successCallback}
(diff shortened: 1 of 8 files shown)


SachaG:
Sorry it took me a while to look at this! It all looks great, but I'm just wondering why you added a `CommentsDelete` component using NovaForms? Why not just call the method directly?

(See how posts are deleted here: https://github.com/TelescopeJS/Telescope/blob/devel/packages/nova-base-components/lib/posts/PostsEditForm.jsx#L19-L27 )

justintime4tea:
That's a very good question Sacha :) I guess I hadn't taken a look at how Posts are deleted and instead jumped the gun on modifying the CommentsEdit. I'll take a look at the direct method call of posts delete and see if I can refactor this so it is more efficient/better.

justintime4tea:
Refactored to use direct call, making new branch and new pull request.

SachaG:
Sounds good, I'll check out the other PR then. Thanks again!

PR #1318, opened 2016-05-03
Only admin see post stats

kienvn:


SachaG:
Good idea! Although note that the data itself (`viewCount`, `clickCount`, `score`, etc.) is currently still published to the client. So anybody can still access it through the browser console.

kienvn:
Maybe we can handle this in PostsStats by passing user var

SachaG:
Well whatever we do in PostsStats won't change what gets published or not.

PR #1320, opened 2016-05-05
add user's posts to UserProfile

auggernaut:
Show user's posts on their profile page.

### DIFF
--- a/packages/nova-base-components/lib/users/UserProfile.jsx
+++ b/packages/nova-base-components/lib/users/UserProfile.jsx
@@ -1,8 +1,10 @@
 import React, { PropTypes, Component } from 'react';
+import SmartContainers from "meteor/utilities:react-list-container";
+const ListContainer = SmartContainers.ListContainer;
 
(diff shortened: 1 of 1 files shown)


SachaG:
Good work! I think it might be a good idea to also specify the `selector` and `options` arguments though, because the way it is now it would just look for `Posts.find()` and display every post it finds. Which is not a problem currently, but in the future you might want to also show a list of the post's upvoted posts on the same page, which would cause problems (let me know if that sounds confusing, it can be a little tough to explain ;). 

In the routes.jsx file, I just do:

```
const {selector, options} = Posts.parameters.get(queryParams);

      mount(App, {content: 
        <ListContainer 
          collection={Posts} 
          publication="posts.list"
          selector={selector}
          options={options}
          terms={queryParams} 
          joins={Posts.getJoins()}
          component={PostsList}
          cacheSubscription={false}
        />})
```

SachaG:
I'm going to merge this in manually.

PR #1321, opened 2016-05-05
Post by feed package

xavcz:
That was a lot of fun to work on it! I had to dive in NovaForms and ListContainer to see how they are built 💯  And well, also the legacy `telescope:post-by-feed` 👍 

So what's up? I'd like to have your feedback on what has been built so far 😄 

As an admin, you can add a feed, see the existing feeds and remove them. I didn't implement an edit method yet.

To performs these actions, an admin can open a modal via the UserMenu dropdown. The modal's display  is likely as the same as the settings form (I tweaked the `modalOpen` state).

Here is a screenshot of the modal displaying most of the available states, in the case of opening the modal over a normal user profile right after a feed has been added, and looking at which users can be set for a new feed:
modal nova post-by-feed

Inside this modal: 
- a `ListContainer`, subscribing to the admins/owner only, shows the `NovaForm` which will trigger a `feed.new` method (= legacy `insertFeed`). The `Feeds.schema` provides a `prefill` property on the currentUser, so it prevents an error if the user selector was not clicked (without that, it shows the username, but don't pass any `userId` value to the method). This schema also maps and filters on `Meteor.users` to display in the select *only* the admins: the scene takes place in a modal & without this option, it would add to the selector input all the others users available (all available users in posts list, normal user profile).

- a `ListContainer` renders all the feeds imported. It's based on `PostLists` structure, and even reuse the same classNames 😅 . I extended the legacy schema with a title, which is set when the feed is fetched.

Server-side, not much has changed other than modernizing with npm modules imports, and updating the feed doc with a title when reading its meta.

I decided to display this package with modal style to stay consistent with the Nova structure. I thought about editing a feed via a link/button rendering a nova form + doc container on the place of feed item. Maybe the way I play with the data can be improved, I tried the most to not overpublish and have access to only what we need. I also thought about displaying the count of posts by feeds, but `Counts` is commented everywhere in publications, I remember you saying that it didn't work some time ago and I didn't give it (dunno if it provides value).
What's in your mind? 🎱


SachaG:
Really nice work! But what would you think about making this a separate package and keeping it out of core? Feed parsing is a really cool feature but I don't know if it's something every single Nova install needs out of the box. 

So I'm thinking maybe this belongs more as a separate repo, with a separate maintainer (you ;). Of course this is more work for you, so maybe you could charge for it or have a pro version or something?

SachaG:
Also another small thing, it'd be great if people also had the option to configure their feeds via their `settings.json` file. Of course, handling both methods at the same time can be a little tricky so there's pros and cons…

xavcz:
Yeah, it's not a core feature! 

I thought about having it as an additional feature, and so extend the components UserMenu for example and have the Feeds Components in the separate package. I go for it :)

Considering the `settings.json`, I tried at first to implement the feeds directly inside the settings collection (haha), but now I see how I can make them both cohabits. I'll go for it too, I like the idea and see more or less how I can do!

I close the PR 💯

PR #1323, opened 2016-05-06
Complete soft delete feature for comments (Revision 2)

justintime4tea:
This follows the same functionality and style of deleting posts, no use of NovaForm, direct call.

- Add deleteComment function to CommentsItem.jsx
- Add Delete link next to Edit link in CommentsItem.jsx
- Hide reply link if comment.isDeleted
- Add styling for Delete link element in _comments.scss
- Add check for document.isDeleted in Users.can.edit in permissions.js


SachaG:
Thanks, sorry it took me a while to merge this in :)

PR #1324, opened 2016-05-07
Load categories at startup in load_categories.js

hotello:
Loading categories inside Meteor.startup() allows inserting categories from settings.json with an extended schema (extended with Categories.addField()).


SachaG:
Good point! Thanks :)

PR #1332, opened 2016-05-22
Newsletter + Mailchimp subscription enhancement

### LINKED ISSUE #1327: Subscribing to newsletter (already signed in): Error invoking Method 'users.setSetting': Internal server error [500]
I'm having an issue when the user subscribes (already signed in).

It's not saving the cookie/settings to hide the newsletter box once the user subscribes because the newsletter box still shows after a refresh.

The errors in the console are:
`Exception while simulating the effect of invoking 'users.setSetting' Error: After filtering out keys not in the schema, your modifier is now empty(…)
debug.js:41` 

`Error invoking Method 'users.setSetting': Internal server error [500]`

This is the error in my terminal:

`Exception while invoking method 'users.setSetting' { stack: 'Error: After filtering out keys not in the schema, your modifier is now empty\n at [object Object].doValidate (packages/aldeed_collection2-core/lib/collection2.js:374:1)\n at [object Object].Mongo.Collection.(anonymous function) [as update] (packages/aldeed_collection2-core/lib/collection2.js:173:1)\n    at [object Object].usersSetSetting (packages/nova:users/lib/methods.js:161:11)\n    at [object Object].methodMap.(anonymous function) (packages/meteorhacks_kadira/lib/hijack/wrap_session.js:164:1)\n    at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1704:12)\n    at packages/ddp-server/livedata_server.js:711:19\n    at [object Object]._.extend.withValue (packages/meteor/dynamics_nodejs.js:56:1)\n    at packages/ddp-server/livedata_server.js:709:40\n    at [object Object]._.extend.withValue (packages/meteor/dynamics_nodejs.js:56:1)\n    at packages/ddp-server/livedata_server.js:707:46', source: 'method' }`

Any help would be really appreciated!

xavcz:
I ran in the same issue that @dannyvaughton encountered, and I realized that it was filtering because the field `telescope.newsletter_showBanner` was not defined! 

It led to another bug which was on the `Users.getSetting`, at the level of `Users.getProperty` : using false as a defaultValue in the `Newsletter` component, the `_.isEmpty(settingValue)` returned true instead of false 😱  -> I switched to check if it's typeof undefined instead of checking null.


SachaG:
Actually I think I wanted to get rid of `newsletter_showBanner` altogether. The only reason it was there in the first place was so that users have a way to subscribe to the newsletter if they ever dismiss the banner, but a better way would be to add a subscribe/unsubscribe control hooked to the Mailchimp API to the user account page. 

So I would suggest removing `newsletter_showBanner` and fixing whatever was depending on it to work without it :)

xavcz:
Haha ok 👍

packages/nova-base-components/lib/common/NewsletterButton.jsx
```diff
@@ -0,0 +1,58 @@
...
+    };
+  }
+
+  render() {
+    const isSubscribed = Users.getSetting(this.props.user, 'newsletter_subscribeToNewsletter', false);
+
+    return (
+      <Button
+        className="newsletter-button"
+        onClick={this.subscriptionAction(isSubscribed ? "removeUserFromMailChimpList" : "addUserToMailChimpList")}
```
xavcz: maybe this could be moved directly in the `subscriptionAction` method, this would repeat `Users.getSetting` again but maybe clearer?

packages/nova-base-components/lib/common/NewsletterButton.jsx
```diff
@@ -0,0 +1,58 @@
...
+  constructor(props) {
+    super(props);
+    this.subscriptionAction = this.subscriptionAction.bind(this);
+  }
+
+  /**
+   * @summary subscribe or unsubcribe
+   * @params action: string, name of the method for subscribing or unsubscribing user
+   */
+  subscriptionAction(action) {
```
xavcz: handle two types of action, depending on the string passed in (see render and comment below)

packages/nova-base-styles/lib/stylesheets/_users.scss
```diff
@@ -18,4 +18,9 @@
     display: block;
     line-height: 1;
   }
+}
+
+.users-newsletter-settings {
```
xavcz: learned saas variables & include there. mind blown 💯

packages/nova-newsletter/lib/server/campaign.js
```diff
@@ -1,4 +1,4 @@
-import Avatar from 'meteor-avatar-core';
+import { Avatar } from 'meteor/nova:core';
```
xavcz: the missing one ! 🎯

packages/nova-users/lib/helpers.js
```diff
@@ -232,8 +232,8 @@ Users.getProperty = function (object, property) {
   var array = property.split('.');
   if(array.length > 1){
     var parent = array.shift();
-    // if our property is not at this level, call function again one level deeper if we can go deeper, else return null
-    return (typeof object[parent] === "undefined") ? null : this.getProperty(object[parent], array.join('.'));
+    // if our property is not at this level, call function again one level deeper if we can go deeper, else return undefined
+    return (typeof object[parent] === "undefined") ? undefined : this.getProperty(object[parent], array.join('.'));
```
xavcz: If my setting was at false, the old way was returning `null`: the chaining of `getProperty` -> `getSetting` with the `null` + `_.isEmpty` doesn't return the good value.

That's why I switched to `undefined`

packages/nova-base-components/lib/common/Newsletter.jsx
```diff
@@ -1,70 +1,63 @@
...
+    const showBanner =
+      Cookie.load('showBanner') !== "no" &&
       !Users.getSetting(context.currentUser, 'newsletter_subscribeToNewsletter', false);
 
     this.state = {
       showBanner: showBanner
     };
   }
 
+  componentWillReceiveProps(nextProps, nextContext) {
```
xavcz: update the banner reactively! my first "context question" on the slackroom was about that... 💃  but in fact my problem was just related to handling the component lifecycle, not about the context!!! 😄

SachaG:
Great work :)

PR #1333, opened 2016-05-23
bug fix: select nova package, not meteor-avatar-core

Falieson:
updated selection of modules package
 so fresh clone will install and boot


SachaG:
Thanks! That was already fixed on the `devel` branch though :)

Also the reason the docs are not linked is that they are not up to date for Nova.

PR #1342, opened 2016-06-10
Hook up siteImage in settings to the open graph meta tags.

BenTristem:
Adjusted Open Graph image sharing using custom settings, and verified with Facebook's Open Graph Debugger.


xavcz:
Nice one! However it seems that a "s" is missing on `this.prop` L13!

sampattuzzi:
Thanks, @xavcz, Hopefully that fixes it?

SachaG:
Thanks!

PR #1348, opened 2016-06-14
Fixes

xavcz:
Hi 👋 

Some fixes on context passing which were blocking actions like editing / adding new categories and upvoting (3 commits):
* default export for Messages and pass context messages, etc. to child components
* pass messages, etc. to child components on ContextPasser
* intl strings + context on settings edit form

And a fix about a bug indicated on slack about mailchimp.


SachaG:
awesome work! thanks!

PR #1351, opened 2016-06-15
Add polyfill for Intl.

sampattuzzi:
Safari and other older browsers don't support the new JS Intl API. So
react-intl will not work on these browsers and prevents navigation. The
fix is to use the included polyfill.


SachaG:
Oh I think it's the same polyfill I'm already using actually. Let me figure out why it's not working properly then.

SachaG:
OK fixed it :)

PR #1353, opened 2016-06-16
Fix post link when viewing post

justintime4tea:
When trailing forward slash is missing from post link, the link href doubles up when viewing the post. Not sure why there is a post link when viewing the post since you are already viewing the post but regardless it should work.

To reproduce the issue, create new clone of Telescope branch devel and open a post then click on it's title again.


SachaG:
Oh you're right that's actually wrong, it should be the post URL.

SachaG:
Fixed it.

PR #1356, opened 2016-06-17
🐙 Fix custom package

xavcz:
I was so into custom packages and understanding react-router + `Telescope.routes` that when I saw that the custom package had used FlowRouter... I couldn't resist 😂 

🎢  Replacement `FlowRouter.route(...)` by `Telescope.routes.add(...)`. 
⌛  Import `moment` where needed.
✉️  `this.renderButton()` wasn't defined anymore (https://github.com/TelescopeJS/Telescope/blob/master/packages/my-custom-package/lib/components/CustomNewsletter.jsx#L24) after the refactoring on the button -> instead of copy/pasting the code of `Telescope.components.NewsletterButton` and its props + context, I recreated the method `.renderButton`.

### DIFF
--- a/packages/my-custom-package/lib/components/CustomPostsItem.jsx
+++ b/packages/my-custom-package/lib/components/CustomPostsItem.jsx
@@ -1,4 +1,5 @@
 import React, { PropTypes, Component } from 'react';
+import moment from 'moment';
 
 class CustomPostsItem extends Telescope.components.PostsItem {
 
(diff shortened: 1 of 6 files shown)


SachaG:
Thanks! But I don't think we really need `transformArrayInObject`? Isn't it just confusing if there are too many ways to do things?

SachaG:
oh disregard my previous message, for some reason GitHub was showing me an old version of the changes.

xavcz:
👍

PR #1362, opened 2016-06-20
use original PostsItem component

xavcz:
Completing your last commit on `master`, just a fix on `PostsItem` which didn't render anymore because it couldn't access to its methods `renderCategories` et al. 😅


SachaG:
Oh good catch, thanks!

PR #1368, opened 2016-06-23
Use defined language (fallbacks to "en")

beeva-franciscocalle:
Settings defined language is not been used (forced to "en")

### DIFF
--- a/packages/nova-base-components/lib/common/App.jsx
+++ b/packages/nova-base-components/lib/common/App.jsx
@@ -5,7 +5,7 @@ import { AppComposer } from "meteor/nova:core";
 class App extends Component {
 
   getLocale() {
-    return Telescope.settings.get("locale", "en");
+    return Telescope.settings.get("locale", Meteor.settings.public.language || "en");
   }
 
   getChildContext() {


SachaG:
What's wrong with using `locale` as a setting?

beeva-franciscocalle:
Ok, my bad. I misunderstood how the language should be set. It's a bit confusing since `sample_settings.json` has this config attribute called `language`. Maybe the sample file should include a default `"locale": "en"`

beeva-franciscocalle:
Is this behaviour by design? I mean, having to add the locale through the db?

SachaG:
Sorry I don't undertsand, why would you use the db?

beeva-franciscocalle:
Maybe I've missed something. This is how I did the setup.

* I've removed everything from the db
* Launched Telescope and check it was in English. Shutdown.
* Remove again everything from the db
* Added this settings file named local.settings.json:
```json
{
  "public": {
    ...
    "language": "es",
    "locale": "es",
    ...
  },
  ...
  "locale": "es",
  ...
}
```
* Restart using `--settings local.settings.json`
* Check the language and It's still in English. Console log for `Telescope.settings.get('locale');` returns `en`

This is the list of packages I'm using:
```
############ Nova Core ############

nova:core # do not remove!

############ Optional Packages ############

nova:settings
nova:users
nova:posts
nova:comments
nova:newsletter
nova:search
nova:notifications
nova:categories
nova:share
nova:voting
nova:forms
nova:embedly
nova:api
nova:rss
nova:kadira
nova:email

############ Customizable Packages ############

nova:base-components
nova:base-routes
nova:email-templates
nova:i18n-en-us

accounts-password
accounts-google
# accounts-twitter
# accounts-facebook

############ Your Packages ############
nova:i18n-es-es
semantic:ui-css@2.1.2
```

To solve this, I had to change the line in the App.jsx or add it manually to the db.

SachaG:
If you set `locale` to `es` in `settings.json` then `Telescope.settings.get("locale")` should definitely output `es`. The database doesn't really have anything to do with it, so I'm not sure what's going on…

PR #1369, opened 2016-06-23
New i18n package - es_ES

beeva-franciscocalle:
Settings defined language is not been used (forced to "en")


SachaG:
Thanks for the translation! But actually I've decided that translations should be separate repos (see https://github.com/TelescopeJS/nova-i18n-fr-fr). So if you host and publish the repo yourself, I'll just link to it from the docs :)

beeva-franciscocalle:
Great, here is the repo: https://github.com/beeva-franciscocalle/nova-i18n-es-es

SachaG:
Awesome. You'll also need to publish it to Meteor's package repo (`meteor publish --create`) before people can use it.

beeva-franciscocalle:
Done, here is the link. I've changed to my personal account: https://atmospherejs.com/fcallem/nova-i18n-es-es

PR #1373, opened 2016-06-27
Different fixes

xavcz:
Howdy! 👋 

Here are some stuffs I fixed while playing with the app!

1. 8cb454d : `function()` didn't let `this` `context` go through. (💭  this keyword, context, this context, mmmh. there surely a pun to make but I'm not good english speaker enough)

2. d57cfcf : we couldn't create new category anymore, collection imported in `CategoriesNewForm`

3. c67c04f : Clean a little `HeadTags` and fix a bug: 
=> cleaning: by default the image meta is the logo url defined in settings (force 0). It can also be the siteImage defined in settings (force 1). Or last but not least a props passed to the HeadTags component (force 2). 
=> bug fix: if the image specified is a complete url http, ftp or whatever (contains `//`), don't do anything. Else this would mean it's stored locally, add telescope site url.

4. bfcc395 : I found that user profile and edit account was in perpetual loading state -> imported `Users` where it was needed for them, publication & components! (4bf0a15 console.log removal)

Considering this last part about `Users`, in roles: `Users.is.adminById = Users.is.admin`. Is this `adminById` a tweaky remainder of good ol' Legacy? Shall we change it by `.admin` when we encounter it or leave at it is?

Cheers! 🚀

### DIFF
--- a/packages/nova-base-components/lib/categories/CategoriesNewForm.jsx
+++ b/packages/nova-base-components/lib/categories/CategoriesNewForm.jsx
@@ -1,5 +1,6 @@
 import React, { PropTypes, Component } from 'react';
 //import { Messages } from "meteor/nova:core";
+import Categories from "meteor/nova:categories";
 import NovaForm from "meteor/nova:forms";
 
 const CategoriesNewForm = (props, context) => {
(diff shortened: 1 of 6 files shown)


SachaG:
Thanks! And yeah we should change it to `Users.is.admin` I guess.

PR #1374, opened 2016-06-28
add post._id as a childContextType for access by custom forms, and placeholder support for custom forms

sweetppro:


SachaG:
Thanks, I'll check out the placeholder stuff! Not sure about the `_id` thing though. Maybe a more generic way to do it would be to pass NovaForm's `getDocument()` down through the context?

sweetppro:
sounds good, i didn't want to pass to much down, so figured just the _id would be enough for customisations

SachaG:
Well, if you just pass the method you're not actually passing anything down. So I think it's a good compromise.

SachaG:
I'm going to do the changes manually. Btw, for future PRs make sure they're made to the `devel` branch :)

PR #1376, opened 2016-06-29
Meta SSR with react-helmet

xavcz:
Got it working! 🎸 

You can view the diff by curling my instance & sidebar :
- Nova with helmet: `curl http://meteor.startupfrance.co` -> has meta in the head
- Nova with dochead: `curl http://nova.sidebar.io` -> no meta in the head

ps: I advice to test with `curl` and not fb debugger because I have a nginx issue with the fb debugger on my instance 😓  I do believe that the scraper would see the meta if nginx is well configured (not my case).


SachaG:
Nice, merging that right away :)

PR #1381, opened 2016-07-04
Fix Facebook settings error in sample_settings.json and README.md

abizermk:
Within the Facebook social logins settings `clientId` should be `appId`


SachaG:
Thanks!

PR #1387, opened 2016-07-06
remove double inclusion of base styles

IlyaSemenov:
The base styles were included twice, first in `package.js` then imported in `main.scss`
This patch gets rid of the first inclusion.


SachaG:
Good point! By the way, please make future PRs to the `devel` branch if you can. I'll merge this one in manually :)

PR #1391, opened 2016-07-09
set locale in settings

fortunto2:
not work if only language, need set locale


SachaG:
You're right, it should be `locale` not `language`.

fortunto2:
thanks!
i start work with russian for telescope
https://atmospherejs.com/lifefilm/nova-i18n-ru-ru

SachaG:
Awesome :)

PR #1393, opened 2016-07-11
fixed types comparison in Posts.isApproved helper

schabluk:
`Posts.isApproved` helper return 'false' when admin user is adding new post with the Status = "approved" - the `post.status` is a `String` compared with `Integer` value of `Posts.config.STATUS_APPROVED`.
As a result, the `postedAt` value is not stored in the Posts collection, which leads to ugly error whenever `<FormattedRelative value={post.postedAt}/>` is used, i.e. in the `PostsItem`.

I was able to fix this simply by parsing `post.status` with `parseInt()`.


SachaG:
Thanks!

PR #1403, opened 2016-07-21
NovaForm: custom control has access to document as a props

xavcz:
Hey!

It all started by fixing a bug when editing another user as an admin, if you were to click on the user's newsletter setting, it would un/subscribe yourself.

I needed my `NewsletterSubscribe` to have access to the user document. By passing `getDocument` function in the context, I got errors (`this.props.document is undefined`). 

So I have added a new piece of logic when looping through a fieldSchema : if it's a function (React Component), give it a prop `document`. https://github.com/TelescopeJS/Telescope/commit/91cc0e3f1bdc76fa0cf6fe409cee83377d95f4d5

What do you think of that? I believe it can help to create more flexible/what-ever-you-need components in NovaForms.

(so in this PR, I added this logic and fixed the behavior of the newsletter button)

### DIFF
--- a/packages/nova-base-components/lib/common/Newsletter.jsx
+++ b/packages/nova-base-components/lib/common/Newsletter.jsx
@@ -62,7 +62,7 @@ class Newsletter extends Component {
 
   renderButton() {
     return <Telescope.components.NewsletterButton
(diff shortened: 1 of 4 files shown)


SachaG:
Nice work!

PR #1405, opened 2016-07-22
Added category sorting, by order

dannyvaughton:
I noticed categories were not being sorted by their order.


SachaG:
Thanks, but `$query` and `$orderBy` look weird. Maybe double-check that? Also ideally the same sort would be applied client side, too.

PR #1410, opened 2016-07-27
Refactored subscribe-to-posts

schabluk:
Here is the basic functionality, with one UI component available by import.

```
import { SubscribeButton } from "meteor/subscribe-to-posts";
...
<SubscribeButton post={post} />
```

Like in original code, the `subscribeItem` and `unsubscribeItem` functions are also exported from the package.

Notifications are being send by callback function CommentsNewNotifications from nova-comments/lib/callbacks.js.

ToDo:
- UI component for `user_subscribed_posts` template
- Moving notification from nova-comments/lib/callbacks.js to subscribe-to-posts.
- Cleanup old code


SachaG:
Really awesome work! Thanks so much :)

PR #1412, opened 2016-07-28
Extendable Subscribe component & locale

schabluk:
What was done:
- New `Subscribe` component from which `SubscribeButton` extends.
- Updated dependencies (fontawesome)
- Updated `nova-i18n-en-us` package with locale for `Subscribe` component
- Fixed `Subscribe` onClick event propagation
- Moved  subscribers notification from nova-comments as a new `SubscribedCommentsNotifications` callback function.


MHerszak:
@schabluk this is a nice component. Could you provide an example of usage with it?

schabluk:
@MHerszak sure. The best way is to put it into your `CustomPostsItem`. You can use `Subscribe`, which is a simple link, or fancier SubscribeButton.

```
import { Subscribe, SubscribeButton } from "meteor/subscribe-to-posts";

class CustomPostsItem extends Telescope.components.PostsItem {
  render() {
    // ...
    return (
      // ...
        {this.renderCommenters()}

        <SubscribeButton post={post} />
      </div>
    )
  }
};
```
Looks like this:
observe

MHerszak:
@schabluk: Thanks! I will check it once in a while.

SachaG:
Nice work as always! A few remarks:

- Do we really need both `Subscribe` and `SubscribeButton`? Why not just `Subscribe`?
- I'd rather not have a dependency on `fontawesome` here. Plug-ins should provide additional back-end features, but if possible they should avoid overloading you with dependencies you might not need.

So basically a plugin should give you extra tools, but it's up to you how you actually use them in your app (i.e. which icon you use, how you set up your button component, etc.)

schabluk:
Good point! Removing `SubscribeButton` and 3rd party deps. 

What about UI component for user_subscribed_posts template? Shall it be obsolete?

SachaG:
Maybe we can still have the list of a user's subscribed posts as an exported component? I don't think that would add too many dependencies?

schabluk:
Finally I had time to make it work :) The `SubscribedPosts` is based on `PostsHome` component, and the `userSubscribedPosts` view is as it was in the old package, only with small modification.

PS. Accidentally, I pushed bootstrap.css update (to 4.0.0-alpha.3)...

SachaG:
Nice! So is this ready for merging?

schabluk:
Yes!

SachaG:
@schabluk I've looked through the code and cleaned up/reformatted a couple things. Do you think you could check out [the commit](https://github.com/TelescopeJS/Telescope/commit/99a70a326233b3cbdf251aaebbeab24e7a8d2b9f) and make sure everything still works/makes sense?

Also I've started fleshing out the Readme, if you have time maybe you could add more details?

(for some reason a weird `.build.telescope-blank` folder got commited too, sorry if it makes things a bit messy. Not sure what that's about…)

schabluk:
@SachaG I did small changes and it works like charm :) Pull request will follow.

I will update the Readme in a days. Don't know what the `.build.telescope-blank`, I think it's not needed.

PR #1417, opened 2016-08-04
Proposal for a CanDo Higher-Order Component

xavcz:
**Goal:** have a generic component to handle permission adapted to the new groups API

**Trigger:** When updating post-by-feed to use the new group API, I realized that I didn't have any check on permission to show or not my components.

**Problem:** Looking at the base components available in `/permissions`, they all do the same kind of things and we repeat the code over and over with just some small variants in 5 components :
- check if the user is logged in or not
- check if the user can perform the action
- display a no permission message

Moreover, in many components, we check permissions like `{Users.canEdit(this.context.currentUser, this.props.post) ? component : ""}` inside render methods.

**Hypothetic solution:** In order to ease the writing of components and handle permissions with more flexibility, I created this `CanDo` component that follows a similar logic.

You supply it an action, a document or not, and a "no permission" message or not, and it takes care of the job for you. I've tried it on the post new form, on the edit action on posts list and on user edit

By now, it's just a draft and a lot of boilerplate. If we go farther in this logic, I think we could use it as a HOC wrapping an exported component : 

```
// MyComponent code
// ...

const permissionsOptions = {
  action: "posts.view.pending.all",
  // ...
};

export default canDo(permissionsOptions)(MyComponent)
```

Feedbacks & discussions really welcomed!

### DIFF
--- a/packages/nova-base-components/lib/components.js
+++ b/packages/nova-base-components/lib/components.js
@@ -60,6 +60,7 @@ Telescope.registerComponent("CategoriesNewForm",    require('./categories/Catego
 
 // permissions
 
+Telescope.registerComponent("CanDo",                require('./permissions/CanDo.jsx'));
 Telescope.registerComponent("CanCreatePost",        require('./permissions/CanCreatePost.jsx'));
 Telescope.registerComponent("CanEditPost",          require('./permissions/CanEditPost.jsx'));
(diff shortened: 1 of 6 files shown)


SachaG:
Seems like a good idea! I'm not sure about using a HoC versus just having a normal `CanDo` component. But in any case this seems like a good first step, I think you can merge it :)

PR #1418, opened 2016-08-04
Change username to authorName in email template

jeffreywyman:
Social logins might not have a username, but they will always have an
authorName.  This fix changes {{username}} to {{authorName}} in email
templates, avoiding the "new user account: undefined" bug.


SachaG:
Are you sure? `authorName` refers to a comment or post author's name, I'm not sure why a user would have an `authorName`? I think it makes more sense to figure out why some users don't have `usernames`.

SachaG:
What if we replace `username` with `displayName` here? https://github.com/TelescopeJS/Telescope/blob/master/packages/nova-users/lib/notifications.js#L6

jeffreywyman:
Thanks, great suggestion!  I will try this out.

SachaG:
Should I close this?

jeffreywyman:
Yes, thanks :) haven't had time to play with this

xavcz:
Support added on `devel` 👍

PR #1420, opened 2016-08-05
Corrected imports for debug convenience globals

pcoutin:
When debugging, I wondered why Posts.find().fetch() returned the same as Users.find().fetch() (with both console.table() on the client and meteor shell)

It's because the imports for the Comments, Users and Posts globals were all importing from nova:posts which exports Posts as default.. so they were equal


SachaG:
Oh wow I can't believe I never noticed that! Thanks!

PR #1425, opened 2016-08-12
nova:subscribe all the things

xavcz:
Long PR to explain all these abstraction I wrote and documentation draft for the package 

# In short: What this PR provides
Availability to subscribe to `Posts`, `Users` & `Categories` out of the box, and can be extended to whatever collection.

### Via :
Generic methods `collection.subscribe` & `collection.unsubscribe`.
Reusable component `<Telescope.component.SubscribeTo document={} documentType='' />`
Email notifications when something you are subscribed to has a new post/comment

...

# Long desc
## Goal
subscribe all the things 😝

## Trigger
Customizing Nova for a client, he asked me to make it possible for a user to subscribe to posts, other users, and categories. He gave the agreement to publish it. @schabluk told me I could continue its first adaptation of the legacy package, thanks a lot & high five man ✋

## Problem
How to make `nova:subscribe` extensible to almost any domain (posts, users, ..) of Nova? 

And more important, how to make it easy for everyone to use it on their project?

## Story-lution
Starting by adapting the existing methods for the `Users` collection, I remarked that the same pattern would be used over and over for both methods and components.

If we consider a `collection` we have :
1. Two methods: `collection.subscribe(documentId)` & `collection.unsubscribe(documentId)`
2. One component to subscribe/unsubscribe **~=** `<CollectionSubscribe doc={document} />`
3. One callback `SubscribedCollectionNotifications` to notify the subscribers that what they waited for happened.
4. Custom fields `subscribers` & `subscriberCount` (missing on the picture below)

So I ended with a generic pattern to be able to *subscribe all the things* (and simplify my life :octocat: )

[Draft picture](https://files.slack.com/files-pri/T02L990A7-F20Q3V9HR/slack_for_ios_upload.jpg?pub_secret=93d58e78a9)

### Methods structure ( ... & method generator 🚀 )
The structure of a method un/subscribe is always the same:
1. Run checks on the permissions on the current user, or if this is an admin that want to subscribe another user and checks like "is the user subscribing to himself/one of his posts", abort if necessary
2. Prepare the action : subscription/unsubscription, fields, modifier, user collection or not (see one p below)
3. Add to set or pull the subscriber from an array of subscribers in the relevant collection (the one which is "subscribed")
4. Update the subscriber to log her action in its fields

There is an edge case, the `Users` collection stored the subscriptions on `telescope.subscribers` and not just `subscribers`. So check are done every time we have to deal with the relevant field. 

After solving this edge cases, I happened to repeat myself a lot ([screenshot](https://www.dropbox.com/s/rcul1tjn71yc7ua/Capture%20d%27%C3%A9cran%202016-08-12%2014.30.19.png?dl=0)). 

Few things apart some letters changed, so I created a generator of methods based on collection ([another schema](https://slack-files.com/T02L990A7-F20QAMFUK-73ea6ce18c)): from an array of collections `[Posts, Users, Categories]`, it generates the two methods I need with the approriate checks, etc. 👍 

It follows the previous code from @SachaG & @schabluk, abstracted and more in depth, and can be adapted to any collection, as long as it has custom fields `subscribers` and `subscriberCount`.

### Reusable Component
Same kind of stuff happened for `<PostsSubscribe />`, `<UsersSubscribe />` and friends, so I adapted their schema to one component:

```jsx
<Telescope.components.SubscribeTo document={post} documentType={"posts"} />
<Telescope.components.SubscribeTo document={user} documentType={"users"} />
<Telescope.components.SubscribeTo document={category} documentType={"categories"} />
```

### Callbacks
Well, I didn't abstract it, but it's always the same thing, check if there are subscribers & create notifications if relevant

### Custom fields
`subscribers` & `subscriberCount` needed on the collection to be subscribe-able.

## Conclusion
It's working well 💃 
There still maybe some work on email templates, but I'd like nova-hacker's feedback on the usefulness / usability of this feature first 😅


xavcz:
Subscribe all the things

SachaG:
Wow, that's so awesome! I'm blown away :)

The only thing I wonder about, is if it wouldn't be possible to remove the dependencies on `nova:posts`, `nova:users`, `nova:categories`, etc. inside the `nova:subscribe` package, and split out `Meteor.methods(subscribeMethodsGenerator([Posts, Users, Categories]));` into each fo the three packages?

Of course then you would need a dependency to `nova:subscribe` inside `nova:posts`. But maybe that's better?

xavcz:
In terms of flexibility, that would make more sense!

xavcz:
>Of course then you would need a dependency to nova:subscribe inside nova:posts. But maybe that's better?

I gave it a try, but I'm facing several different troubles 😐

The method creator depends on `nova:users` to run checks on permissions with `Users.canDo`. Creating the actions on `default` and `admins` could be done directly in `nova:users` but it definitely needs to import `Users` in the building process (https://github.com/TelescopeJS/Telescope/pull/1425/files#diff-27d19fc6fa64127fa5bdd448dfd17731R149)

Then, there are all the callbacks, custom fields and views that will rely on the different collections and so packages (categories, posts, etc.). I feel weird about "forcing" a developer to use `nova:subscribe` as a vital dependency to run the app.

While writing, I found your thread about weak deps https://forums.meteor.com/t/weak-dependencies-with-import/25358 😅 

❓ It could looks like: 
```js
// in nova:posts - methods.js
if (Package['nova:subscribe']) {
  import subscribeMethodCreator from 'nova:subscribe';
  subscribeMethodCreator(Posts);
}
```
And same kind of thing for callbacks, custom fields, etc. in packages such as categories, users and whatever.

To be honest, it still feels weird for me like this too. 

Other hypothesis, doing this nested imports in `nova:subscribe` and check if the collection package exists: 
```js
if (Package['nova:categories']) {
  import Categories from 'nova:categories';
  subscribeMethodCreator(Categories);
}
```
Perhaps better?

Lot of uncertainty here, I would tend to use the last option. 

Maybe this kind of solution could work thanks to `reify`'s nested imports, however would `reify` imports and `Package` Meteor API work with the upcoming `webpack` integration? 😱

Help wanted! What's your opinion?

PR #1428, opened 2016-08-16
little detail makes big difference for nob like me

jodumont:
- git clone git@github.com:TelescopeJS/Telescope.git
+ git clone https://github.com/TelescopeJS/Telescope.git


SachaG:
Good idea :)

PR #1430, opened 2016-08-19
new nova i18n package (de_DE)

fargerio:
Here is a german translation.


SachaG:
That's awesome! Vielen Dank :)

PR #1433, opened 2016-08-23
Fix the cancelUpvote to only run once

queso:
Currently you can cancel your upvote many times, this patch fixed the issue on Crater.io


SachaG:
Thanks! I'll fix it manually on `devel`.

PR #1438, opened 2016-09-03
pt_BR (Brazilian Portuguese) Translation

lukasag:
# Brazilian Portuguese translation for Telescope Nova

All strings translated to Brazilian Portuguese
Credits to Lucas R. Simões https://github.com/lukasag


SachaG:
That's great! But actually I would prefer that translations be hosted in their own separate repos. So you can host it on your own GitHub account and just add a link to the main Readme :)

lukasag:
Pull request created. Added to README.md

SachaG:
Thanks so much!

PR #1443, opened 2016-09-06
Use fieldSchema.defaultValue if present

beeva-franciscocalle:
Assigning an empty string to the value when no value is present causes React to throw a warning on check proptyes for those fields having a type different than String


xavcz:
Yeah true @beeva-franciscocalle! 

However, I think it could break forms using prefill options: could you assign the default value to `field.value`, if existing, after the "prefilled values check" please?

xavcz:
Fixed on `devel` 👍

PR #1445, opened 2016-09-11
Update groups.js

comus:
`actions = Array.isArray(actions) ? actions : [actions];`


xavcz:
Nice catch @comus! It adds clarity to these methods 👌 

As of how JS is built it didn't give any error, I guess it'd prevent some bugs if we decide to change the way the group API is built, refactor without using `.concat` or create some enhancements.

PR #1446, opened 2016-09-21
Patch v0.27.1 proposal

xavcz:
After updating the History.md, I thought that would be a good thing for Nova to provide a new patch (semver-y speaking 🖖) fixing most of the issue newcomers have with React versions and also adding some new features, 404 route and latest version of Mup recommandations among others.

@SachaG I know you are focused currently on the [stateofjs.com](http://stateofjs.com), if you agree with this patch proposal, I'll update packages and merge this PR 👍


SachaG:
Awesome work! Yes I think you should merge it :)

PR #1453, opened 2016-09-30
Added browser Notifications + UI 

meteorplus:
Added browser Notifications + UI
TODO: add the hash# so the page slides to the comment newly added
and more tweaking to the notificationsList...

Thanks a lot


singhalarchit:
My application is using telescope-<> rather than nova-<>

xavcz:
Hey @meteorplus! That looks cool feature! 👍 

However, following the Nova's philosophy; it would better as a 3rd party package that people would install if needed, and not a core feature.

Could you make it an external package and do a PR on `devel` updating the README with a link to your package? :octocat:

SachaG:
Great work! But I agree with @xavcz, I think that's the kind of feature that would be perfect as a third-party package :)

PR #1465, opened 2016-10-07
Devel - Send only one email to a user if a user is subscribed to multiple topics

chptung:


chptung:
Made this incorrectly

Issue #1458, opened 2016-10-04
Make voting algorithm extensible/customizable


xavcz:
This is a [starter issue](http://hood.ie/blog/starter-issues.html), *highly inspired* by the great job of the folks at [Hoodie](http://hood.ie/blog/starter-issues.html). 👍

### 🎃🙀👕🔭😻 Hacktoberfest: Trick or Treat!

If you haven’t yet, sign up for [Hacktoberfest](https://hacktoberfest.digitalocean.com/) to earn an exclusive T-Shirt. 

We are sure you can learn a cool trick or two in the process on how to hack & customize Telescope Nova! 🔭

### 🤔 What you will need to know

Meteor, React, Telescope Nova callbacks handling ([video](https://www.youtube.com/watch?v=fc0-2RVz_-I)).

Most of the Telescope Nova structure is extensible. A big part of that is done thanks to `callbacks` as Nova uses a system of successive callbacks to perform some action. 

The callback below is for example run on the client when a post is submitted:
```js
function PostsNewDuplicateLinksCheck (post, user) {
  if(!!post.url) {
    Posts.checkForSameUrl(post.url);
  }
  return post;
}
Telescope.callbacks.add("posts.new.sync", PostsNewDuplicateLinksCheck);
```

### 🎯 The Goal

The voting algorithm is currently self-contained in `nova:voting/lib/scoring.js` ([code](https://github.com/TelescopeJS/Telescope/blob/master/packages/nova-voting/lib/scoring.js))

As suggested in an old issue (two years ago! #421), this would be awesome to be able to make our own algorithm without modifying the core and from our custom package.

A simple set of callbacks is the one for filtering posts view: `postsParameters`. 

This can be a good starting point to take inspiration to create callbacks for `Telescope.updateScore` function : `"scoring.all"` callbacks.

The current `Telescope.updateScore` content would be broken in several functions, these functions being then added as callbacks. These callbacks would be created and run likely the other Telescope Nova's callbacks.

All these callbacks would be added thanks to `Telescope.callbacks.add("scoring.all", myAlgoExtension)` and run inside `Telescope.updateScore` with something like:
```
newScore = Telescope.callbacks.run("scoring.all", score, otherParams)
```

*Nice to have:* We could have a general algorithm (`"scoring.all"`) and some specific algorithm for `posts` or `comments`: `"scoring.posts"` & `"scoring.comments"`. This would be based on the `collection` argument passed to the function `Telescope.updateScore`.

[...]

PR #1481, opened 2016-10-21
#1458-First draft of customized scoring algorithm

S4KH:
#1458 Need more guidance on this. :sos:


xavcz:
Will have a look in the weekend 💯

S4KH:
Need help on calling callbacks from the `Telescope.updateScore` function. There are 3 callbacks added to the `scoring.all` hook(learned from the video :smile: ). Some of them does status check. When all of them on 1 hook what is the sequence of the calls?

SachaG:
@S4KH thanks for taking the lead on this! To answer your questions, callbacks will just be called in the same order that they were added.

S4KH:
How pass parameters to the callbacks? They all have different number of parameters, some callbacks return value that needs to be used as a parameter another function call. `ItemAgeCheck` returns `ageInHours` which must be passed to `PowerAlgorithm`.

SachaG:
You can see how callbacks are called here: 

https://github.com/TelescopeJS/Telescope/blob/master/packages/nova-lib/lib/callbacks.js#L36-L73

The basic format is:

```
Telescope.callbacks.run(callbackName, iteratedItem, invariant1, invariant2, ...)
```
So you could do:

```
const newScore = Telescope.callbacks.run("scoring.all", item)
```

And then each successive callback would modify and return `item`. Or maybe:

```
const newScore = Telescope.callbacks.run("scoring.all", score, item)
```

In which case each callback would modify and return `score`, while passing `item` as an invariant to every callback. Does that make sense?

S4KH:
@SachaG Thanks for your quick response. Then how do stop callback iteration when 1 of the checks fail. 

    if (!!item.status && item.status !== 2) // if item has a status and is not approved, don't update its score
        return 0;

SachaG:
I think if you need to stop the iteration completely, it needs to be done outside the callback loop. There's no way of stopping callbacks right now.

xavcz:
How are you doing @S4KH? 🙂

S4KH:
Kind of stuck. Because of parameter passing to a callback.

SachaG:
@S4KH maybe you can drop by the Slack chatroom if you have any questions?

xavcz:
Hey @S4KH, thanks for what you've done! 🙌 

This PR is in conflict with the new version of Nova, powered with GraphQL: I'm closing it and will continue the work based on what you've done 💯

Issue #1480, opened 2016-10-21
Add a route + component to allow people to change their password after asking for reset


xavcz:
This is a [starter issue](http://hood.ie/blog/starter-issues.html), _highly inspired_ by the great job of the folks at [Hoodie](http://hood.ie/blog/starter-issues.html). 👍
### 🎃🙀👕🔭😻 Hacktoberfest: Trick or Treat!

If you haven’t yet, sign up for [Hacktoberfest](https://hacktoberfest.digitalocean.com/) to earn an exclusive T-Shirt. 

We are sure you can learn a cool trick or two in the process on how to hack & customize Telescope Nova! 🔭
### 🤔 What you will need to know

Meteor, React. And more specifically Meteor Accounts with React ([`std:accounts-ui`](https://github.com/studiointeract/accounts-ui)) + [`react-router`](https://github.com/ReactTraining/react-router). 
### 🎯 The Goal

Build a route to handle password change when a user clicks the link he/she received by email to reset their password.

The process would look like following steps:
1. After clicking the link, the users gets to the app, an `Accounts.ui.config` hook is triggered: `onResetPasswordHook`.
2. This hook takes care of redirection to the `/reset-password` route we defined in `nova:base-routes`: it's just a string with the route path :+1: ([see API](https://github.com/studiointeract/accounts-ui/blob/87b4469ae50638f9ca6c33de346141b72cfa585f/imports/accounts_ui.js#L197-L199)).
3. This route renders a component which gives the `Accounts.ui.LoginForm` in the appropriate `formState`: `STATES.PASSWORD_CHANGE` and run checks to see if the user is connected and if there is actually a reset password token (`Accounts._resetPasswordToken`).
4. The user resets their password, life is beautiful.
### 📋 Step by Step

**Working on your first Pull Request?** You can learn how from this _free_ series [How to Contribute to an Open Source Project on GitHub](https://egghead.io/series/how-to-contribute-to-an-open-source-project-on-github).

---
- [x] Claim this issue (@malively)
- [x] Fork the repo, [setup repository locally](https://github.com/TelescopeJS/Telescope#getting-started) and checkout on `devel` branch.
- [x] Create a component, `UsersResetPassword`, that renders the `Accounts.ui.LoginForm` in the approriate state thanks to `formState` props and `STATES.PASSWORD_CHANGE` (given by `std:accounts-ui`).
- [x] Create a route with the path `/reset-password` which renders this component.
[...]

PR #1491, opened 2016-10-29
Add Reset Password Feature

malively:
This PR is in response to #1480 

The `componentDidMount` method of `UsersResetPassword` was necessary because [the callback for `Accounts.resetPasswordLink`](https://github.com/studiointeract/accounts-ui/blob/87b4469ae50638f9ca6c33de346141b72cfa585f/imports/login_session.js#L77-L83) was not being called. This is due to the fact that `accounts-base` expects the `reset-password` URL to use query fragments (e.g. `/#/reset-password`) in [attemptToMatchHash](https://github.com/meteor/accounts/blob/77b6084c0bd24129c1e7a6d703d9effddaceadcb/packages/accounts-base/url_client.js#L29-L61). Additionally, we modified the resetPassword URL of accounts-base so that it won't generate links with the query fragment as in [this helpful example](http://stackoverflow.com/questions/26603863/override-reset-password-with-accounts-ui-installed).

Shout outs to @cjlarose for helping me with this weird issue.

### DIFF
--- a/packages/nova-base-components/lib/components.js
+++ b/packages/nova-base-components/lib/components.js
@@ -76,3 +76,4 @@ Telescope.registerComponent("UsersName",            require('./users/UsersName.j
 Telescope.registerComponent("UsersMenu",            require('./users/UsersMenu.jsx'));
 Telescope.registerComponent("UsersAccountMenu",     require('./users/UsersAccountMenu.jsx'));
 Telescope.registerComponent("UsersAccountForm",     require('./users/UsersAccountForm.jsx'));
+Telescope.registerComponent("UsersResetPassword",   require('./users/UsersResetPassword.jsx'));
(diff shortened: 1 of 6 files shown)


xavcz:
Hey @malively! Thanks for the PR, that's awesome! 🎉 

Could you add some messages depending on the status of the token and user to enhance the experience?

Like one if the user comes to this page and there are no token (`Accounts._resetPasswordToken` undefined) and one if the user is connected (meaning he has changed successfully his password)

malively:
@xavcz We added support to just display "Password Changed" if a user is already logged in.

As part of the route the token is required to view the page. So simply going to `/reset-password` will redirect you to the 404 page. 

Let me know if there is any other tinkering you would like to be done.

xavcz:
Fair enough! Merged 👍

PR #1504, opened 2016-11-06
Only show comment reply button for logged in users

qge:


SachaG:
Good idea :)

SachaG:
Oh actually, I just remembered that I wanted to leave the reply button for non-logged in users too so we can suggest they sign up when they click it. But I guess we can remove it for now and work on this later.

PR #1507, opened 2016-11-11
Fix react setState race condition

sherryxiao1988:
when multiple components need to call addToAutoFilledValues (I am using the avatar(xavcz:nova-forms-upload) plugn along with the subscription package), this.state is concurrently accessed/changed by multiple components when they are rendered. autofilledValues sometimes only holds the last value.

See here:
http://stackoverflow.com/questions/38065534/race-condition-in-react-setstate


sherryxiao1988:
@xavcz please take a look :) I spent an afternoon to figure out what happened

xavcz:
Ok I see! Nice one @sherryxiao1988!

note: Please next time do your PR on `devel` branch 👍

Rongchao:
very nice

meteorplus:
this is a very important fix and it should be in master!!

PR #1512, opened 2016-11-17
fix wrong comment in deep function

comus:


xavcz:
Why this empty loop is so important? Could you enlight me?

PR #1513, opened 2016-11-17
Newsletter subcription fixes

bengott:
1. Fixes newsletter banner dismissal persistence
2. Renames user setting telescope.newsletter_subscribeToNewsletter to telescope.newsletter.subscribed
3. Also fixes errors from previous setting-renaming migration


bengott:
Does this also need to be renamed?
https://github.com/TelescopeJS/Telescope/blob/master/packages/nova-i18n-en-us/lib/en_US.js#L77

packages/nova-base-components/lib/common/NewsletterButton.jsx
```diff
@@ -9,12 +9,11 @@ class NewsletterButton extends Component {
     super(props);
     this.subscriptionAction = this.subscriptionAction.bind(this);
   }
-  
+
   subscriptionAction() {
```
xavcz: `props.user` is needed here to handle a tricky edge case: you are someone that has permission to edit users & you are editing someone profile, the subscribe button should act on the user you are editing and not yourself.

...📬 🌮 

Another solution could to simply show the newsletter custom field if you are editing your own profile.

xavcz:
@bengott didn't see your comment on internationalization yesterday, yes it would need to be renamed, as it's used on the account page when you edit your profile.

@SachaG what do you think?

SachaG:
Haven't had time to look at this in detail yet but I think we're getting rid of the Settings collection eventually? Would that affect this PR?

bengott:
@SachaG: I don't think so. My changes only involve changes to the settings in user.telescope, not the Settings collection.

packages/nova-base-components/lib/common/NewsletterButton.jsx
bengott: this should be fixed now

xavcz:
@bengott so it looks really good! Just one last thing, could you move the new schema extension  `newsletter` namespace to the custom fields of `nova:newsletter` ? So that the `nova:users` package remains unaware of this in itself.

We will explicitly tell the people that want to update to add your migration script aside :)

bengott:
Like dis???

xavcz:
Haha yes! Basically moving [dis block](https://github.com/bengott/Telescope/blob/48cd5c83d7348eff266f5cb91abba555091015be/packages/nova-users/lib/schema.js#L133-L147) at the place you pointed out! 💯  💃

bengott:
@xavcz: I made the custom_fields change you suggested. Should be good to go now.

Also, I noticed this on the NewsletterSubscribe component:
https://github.com/TelescopeJS/Telescope/blob/devel/packages/nova-newsletter/lib/components/NewsletterSubscribe.jsx#L28

I started to add an i18n message, but this component doesn't seem to have access to context.intl, even after I added it to NewsletterSubscribe.contextTypes. Looks like the NewsletterSubscribe component is instantiated by specifying it in that custom_fields schema LIKE DIS:
https://github.com/TelescopeJS/Telescope/blob/devel/packages/nova-newsletter/lib/custom_fields.js#L30

Is this the reason the message was hardcoded in English and doesn't use i18n? Or am I missing something? (totally possible -- I don't have much experience with internationalization, and I'm honestly still leveling up on React)

xavcz:
:tada: :octocat:

PR #1520, opened 2016-12-04
adds pt-PT translation (portuguese-Portugal)

brunoamaral:


xavcz:
Hey @brunoamaral! Thanks for your PR, however this i18n.json files are not used anymore and will be removed soon! We'll take your other PR instead :)

brunoamaral:
Thank you @xavcz understood. Let me know when and how I can add the translation then.

PR #1522, opened 2016-12-06
fixed issue with user Bio only displaying for Admin users

sweetppro:


packages/nova-base-components/lib/users/UsersProfile.jsx
```diff
@@ -17,7 +17,7 @@ const UsersProfile = ({user}, {currentUser}) => {
     <div className="page users-profile">
       <Telescope.components.HeadTags url={Users.getProfileUrl(user, true)} title={Users.getDisplayName(user)} description={user.telescope.bio} />
       <h2 className="page-title">{Users.getDisplayName(user)}</h2>
-      <p>{user.telescope.bio}</p>
+      {user.telescope.htmlBio ? <p>{Telescope.utils.stripHTML(user.telescope.htmlBio)}</p> : null }
```
xavcz: That's true that we need `htmlBio`, because it's the actual [published field](https://github.com/TelescopeJS/Telescope/blob/master/packages/nova-users/lib/schema.js#L117), and not bio. 

However, as we authorize the `bio` to be written in Markdown and it's already sanitized in the [collection hooks](https://github.com/TelescopeJS/Telescope/blob/master/packages/nova-users/lib/callbacks.js#L33), we should keep the html created.

Keeping the condition to show or not `user.telescope.htmlBio`, could you display it like `htmlBody` in `PostsPage` please? https://github.com/TelescopeJS/Telescope/blob/master/packages/nova-base-components/lib/posts/PostsPage.jsx#L17

sweetppro:
how about this

xavcz:
Yeah that's the right way to go! Thanks a lot! 🎉 

However, there is a remaining issue... In the upcoming days, we will release and merge into `master` the new version of Nova (**v0.3**) that is using exclusively Apollo as the data layer : *this means a lot of breaking changes and new ways to build products with Nova*. 

Your code as it is now won't fit to this new version, so I've taken it and did the update on the `apollo` branch (https://github.com/TelescopeJS/Telescope/commit/5116256a8887abc78079e6a2d0c14cbff2cd4944)
```
{user.__htmlBio ? <div dangerouslySetInnerHTML={{__html: user.__htmlBio}}></div> : null }
```

Thx again for the PR, everyone will be able to access the formatted user's bio in the next release 🚀 
In the meantime, if you stay on v0.27.5, last Nova release with DDP (the Meteor data-layer), I'd recommend to apply this modification in your custom package 👍  

We'll ensure to redirect people that want this "bio update" on previous version of Nova to your PR!

sweetppro:
awesome
looking forward to playing round with Apollo!

PR #1528, opened 2016-12-26
Allow redux middleware extensions

dominictracey:
Integrating a REST api into the redux store. The reducers and actions are easy to extend, but I need to add redux-thunk into the middleware stack, hence the PR.

### DIFF
--- a/packages/nova-lib/lib/redux.js
+++ b/packages/nova-lib/lib/redux.js
@@ -1,2 +1,3 @@
 export const Actions = {};
 export const Reducers = {};
+export const Middleware = {};
--- a/packages/nova-lib/lib/redux.js
+++ b/packages/nova-lib/lib/redux.js
@@ -1,3 +1,3 @@
 export const Actions = {};
 export const Reducers = {};
-export const Middleware = {};
+export const Middleware = [];
--- a/packages/nova-routing/lib/store.js
+++ b/packages/nova-routing/lib/store.js
@@ -1,7 +1,7 @@
 import { createStore, combineReducers, applyMiddleware, compose } from 'redux';
 // import { routerMiddleware } from 'react-router-redux'
 
-import { Reducers }from 'meteor/nova:lib';
+import { Reducers, Middleware }from 'meteor/nova:lib';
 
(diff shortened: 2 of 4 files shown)


packages/nova-lib/lib/redux.js
```diff
@@ -1,2 +1,3 @@
 export const Actions = {};
 export const Reducers = {};
+export const Middleware = {};
```
xavcz: In order to be spread when calling `applyMiddleware`, I believe this variable should be an array instead of an object? Did it work for you like that?

And just a detail, should we name this var at the singular or the plural (so we don't change it later)? Imo, it might makes more sense to have it named `Middlewares` (plu). I'm not English native, you surely know what makes better sense!

packages/nova-routing/lib/store.js
```diff
@@ -1,7 +1,7 @@
 import { createStore, combineReducers, applyMiddleware, compose } from 'redux';
 // import { routerMiddleware } from 'react-router-redux'
 
-import { Reducers }from 'meteor/nova:lib';
+import { Reducers, Middleware }from 'meteor/nova:lib';
```
xavcz: We'll deprecate  `nova:lib` and group everything in `nova:core` to minimize confusion in a near future. We have a foot in each camp atm, but hope to get rid of `nova:lib` asap!

Could you had to `Middleware(s)` to `nova:core`'s entrypoint please? 🙏 See: https://github.com/TelescopeJS/Telescope/blob/devel/packages/nova-core/lib/modules.js#L1-L3

packages/nova-lib/lib/redux.js
dominictracey: "Middleware" is correct for both singular and plural in English, though [the term is not without it's detractors](https://github.com/rack/rack/issues/332).

dominictracey: And good call on the array!

packages/nova-routing/lib/store.js
dominictracey: Moved this as well, though the implementation of the extension points remains in `nova:lib`.

Issue #1517, opened 2016-11-23
Allow excerpt length in settings.json


z3to:
I noticed that currently the length of the excerpt is hard-coded to 30 words in `Posts.before.insert` and `Posts.before.update`. Wouldn't it be desirable to define this setting globally and even allow to alter this value in the `settings.json`?

And as well for th newsletter: Here the length is in hard-coded to 20 word at two positions in `Newsletter.build`.

xavcz:
Hey @z3to! 

Yeah that looks like a nice-to-have feature to be able to manage the excerpt from the settings!

We would have a default value of 30 / 20 like it is now, and it would be replaced by the settings if you specified it!

Are you OK to do a PR? 👍  Feel free to drop by the [Slack room](http://slack.telescopeapp.org) if you need any help :)

xavcz:
Integrated by @philweber 🎩

PR #1535, opened 2017-01-08
#1517 - Configurable excerpt lengths

philweber:


SachaG:
Nice! Just a couple things: we've gotten rid of globals, so `Telescope.settings.get` is now `getSetting`, imported from `meteor/nova:core`. Also `getSetting` takes a default value as its second argument, so you can just write `getSetting('newsletterExcerptLength', 20)`.

philweber:
Yes, I noticed that (in _devel_ branch). Also, it looks like the Post callbacks are being refactored? Should I make changes to the commented-out code in nova-posts/lib/callbacks.js, or add code to the nova-posts/lib/callbacks/callbacks_posts_new.js, etc.?

SachaG:
Oh right, there's no `Posts.before.insert`, `Posts.before.update`, etc. anymore, it's all Nova callbacks now. But in any case the excerpt is not being set in a callback anymore, but directly in the schema (`schema.js`). Sorry it's all a bit confusing!

PR #1540, opened 2017-01-22
Added test for nova:posts

sonnygauran:
When trying to import package, import fails because of missing dependency on `nova:events`

```
meteor test-packages nova:posts
[[[[[ Tests ]]]]]

=> Started proxy.
=> Started MongoDB.
~/.meteor/packages/meteor-tool/.1.4.2_3.1rd9djy++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/dev_bundle/server-lib/node_modules/fibers/future.js:280
                        throw(ex);
                        ^

Error: Can't find npm module 'meteor/nova:events'. Did you forget to call 'Npm.depends' in package.js within the 'modules-runtime' package?
...snip...
=> Exited with code: 1
```

Adding `nova:events@1.0.0` to Package dependency resolves this issue.

To validate, comment out the import for `nova:events`.
This will halt the test and fail.

### DIFF
--- a/packages/nova-posts/package.js
+++ b/packages/nova-posts/package.js
@@ -11,6 +11,7 @@ Package.onUse(function (api) {
 
   api.use([
     'nova:core@1.0.0',
+    'nova:events@1.0.0',
     'nova:users@1.0.0',
     'utilities:react-list-container@0.1.10'
   ]);
@@ -23,4 +24,15 @@ Package.onUse(function (api) {
   api.mainModule("lib/server.js", "server");
   api.mainModule("lib/client.js", "client");
 
+});
+
+
+Package.onTest(function(api) {
+  api.use([
+    'ecmascript',
+    'tinytest',
+    'nova:posts',
+  ]);
+  
+  api.mainModule('test.js');
 });
--- a/packages/nova-posts/test.js
+++ b/packages/nova-posts/test.js
@@ -1,0 +1,6 @@
+import { Tinytest } from "meteor/tinytest";
+import Posts from 'meteor/nova:posts';
+
+Tinytest.add('nova:posts - initialize', function (test) {
+    test.isNotNull(Posts);
+});


xavcz:
Thanks for the PR!

That's a nice catch, indeed `nova:posts` depends at the moment on `nova:events`, because of this `Posts.increaseClicks` method: https://github.com/TelescopeJS/Telescope/blob/master/packages/nova-posts/lib/server/routes.js#L11-L27 

Instead of making `nova:posts` depending explicitly on `nova:events`, what about having this dependency optional thanks to a callback?

I mean, something like:
```js
// same file nova-posts/lib/server/routes.js

Picker.route('/out', ({ query}, req, res, next) => {
  if(query.url){ // for some reason, query.url doesn't need to be decoded
    const post = Posts.findOne({url: {$regex: escapeStringRegexp(query.url)}});

    if (post) {
      const ip = req.headers && req.headers['x-forwarded-for'] || req.connection.remoteAddress;
      runCallback('posts.click.async', {post: post, ip}); // here is the modification
      res.writeHead(301, {'Location': query.url});
      res.end();
    } else {
      // don't redirect if we can't find a post for that link
      res.end('Invalid URL');
    }
  } else {
    res.end("Please provide a URL");
  }
});
```

And we would move the code of `Posts.increaseClicks` into the `nova:events` package, in a new file called `callbacks.js`, and attach it to a `posts.click.async` callback event.

It would open the gate for a lot of new modular stuff like logging client & server-side actions in the event collection, sending data to google analytics or intercom or whatever service: 
➡️ `collection.new.async`/ `collection.edit.async` / `collection.remove.async` -> log event in db? 🌮
➡️ `collection.new.sync`/ `collection.edit.async` / `collection.remove.async` -> log event to any connected metric service (thanks to segment.io?)?

Would it make sense? cc @sonnygauran @SachaG

sonnygauran:
Sure @xavcz, touching base with you in Slack on this.

xavcz:
Closing after discussion on Slack \o/

Thanks for the heads up, will implement that this week! 🚀 :octocat:

PR #1544, opened 2017-01-26
add react-router-redux to sync history with store

comus:
react-router-redux is not necessary. but it is good for keep all things sync with store

what do you think?


xavcz:
We'll see when we need it :) Thanks anyway!

PR #1558, opened 2017-02-05
Rendering Error due to settings.json file

PatrickHurley75:
I have tried to get to the bottom of this rendering issue myself. It is way beyond me. I did a fresh clean install of devel and ran the program with no customisations at all. localhost:3000/settings causes the following rendering error. It is very like the error that keeps showing up on my Galaxy container. Not enough of a problem to cause it to crash but it is causing occasional "fetch" errors during navigation on the site.
I got this from the console as per sacha's log.console update.

{ [Invariant Violation: Objects are not valid as a React child (found: object with keys {title, tagline, logoUrl, logoHeight, logoWidth, faviconUrl, requirePostsApproval, requireViewInvite, requirePostInvite, enableNewsletter, autoSubscribe, emailNotifications, postInterval, RSSLinksPointTo, commentInterval, maxPostsPerDay, startInvitesCount, postsPerPage, language, locale, twitterAccount, facebookPage, googleAnalyticsId, useSegment, segmentWriteKey}). If you meant to render a collection of children, use an array instead or wrap the object using createFragment(object) from the React add-ons.] name: 'Invariant Violation', framesToPop: 1 }


SachaG:
I'm not sure what's going on in this PR… Looks like some kind of Git confusion issue?

PR #1561, opened 2017-02-06
Routing independent from deprecated packages & folder restructuration of nova:lib&core & update std:accounts-ui to 1.2.18

comus:
* not test a lot

* i modify a lot, i will write  a detail description later.
* sorry that i merge apollo package to core and lib


comus:
the changes

1. auto create cookie `meteor_login_token ` when login/logout token changes

    copy from fastrender

    see [auth.js](https://github.com/comus/Telescope/blob/9992f0063ee6e73809400cbbe7c0bac153345c1f/packages/nova-core/lib/client/auth.js)

2. renderContext

    SSR on server side and browser on client side have a renderContext object.

    renderContext created at `nova:core` package. so that other packages can access it.

    renderContext contains those variables used when client/server rendering.

    such as `store`, `apolloClient` and more

    see [client/renderContext.js](https://github.com/comus/Telescope/blob/9992f0063ee6e73809400cbbe7c0bac153345c1f/packages/nova-core/lib/client/render_context.js)

    see [server/renderContext.js](https://github.com/comus/Telescope/blob/9992f0063ee6e73809400cbbe7c0bac153345c1f/packages/nova-core/lib/server/render_context.js)

3. move `createApolloServer` from `nova:apollo` to `nova:core` package.

    why?

    because `renderContext` in `nova:core` needs `apolloClient`.

    `apolloClient` needs `nova:apollo`, and `nova:apollo` depands `nova:core`

    i don't want cyclic dependency. so i split `nova:apollo` to `nova:lib` and `nova:core`

4. create a function `createApolloClient ` at `nova:lib`

    so that `core` can use it

5. i move `oauth-config.js` from `lib` to `core`

    because i think inside `nova:lib` just stores a lot exportable function to let other package to use

    so i think move `oauth-config.js` to `core` is better.

6. `nova:routing` includes `inject-data` and `react-router-ssr`, but i modify a lot.

comus:
the last several changes:

1. implement a dummy `meteor.subscribe` on server side for LoginForm error

2. change nova-lib and nova-core folder structure, if you dont like this way, just tell me, i can rollback these changes.

the main idea of the new package folder structure:

```
nova-core
--lib
  --client (folder)
    --main.js (entry file, client side)
  --server (folder)
    --main.js (entry file, server side)
  --modules (folder, i added all modules related files in this folder)
    --index.js (index modules file that import other modules related files)
```

comus:
finally, i move renderContext from core to lib, because i found that core's components/containers may need renderContext.

for example: withList
```js
...
// graphql query options
options({terms, paginationTerms, client: apolloClient}) {

  // if you don't pass client withApollo, i can get apolloClient from renderContext
  const { loginToken, store, apolloClient, history, reducers, middlewares } =  require("meteor/nova:lib").renderContext.get();
  console.log(require("meteor/nova:lib").renderContext.get())

  const mergedTerms = {...terms, ...paginationTerms};
  return {
    variables: {
      terms: mergedTerms,
    },
    // note: pollInterval can be set to 0 to disable polling (20s by default)
    pollInterval,
    reducer: (previousResults, action) => {

      // see queryReducer function defined below
      return queryReducer(previousResults, action, collection, mergedTerms, listResolverName, totalResolverName, queryName, apolloClient);

    },
  };
},
...
```

or use renderContext on any components

it's awesome!

comus:
@xavcz 

> why are there duplicated code?

for me, i try not to using Meteor.isServer/Client condition. so that if i modify `client/apollo.js` it won't trigger a meteor server restarted.

and it reminds me not use options to assign currentUserToken on client side.
i accidentally do something wrong like this before on client side.

```
apolloClient = createApolloClient({ currentUserToken: global.localStorage['Meteor.loginToken'] });
```

yeah, it s not good. on client side it doesn't need assign currentUserToken.

and if i merge client/server apollo.js together

that will write more code for checking like before

```
const { cookieLoginToken } = config;
// Meteor accounts-base login token stored in local storage, only exists client-side
const localStorageLoginToken = Meteor.isClient && Accounts._storedLoginToken();

// on initial load, prefer to use the server cookie if existing
let currentUserToken = cookieLoginToken || localStorageLoginToken;

// ...a login token has been passed to the config, however the "true" one is different ⚠️
if (Meteor.isClient && cookieLoginToken && cookieLoginToken !== localStorageLoginToken) {
  // be sure to pass the right token to the request 🎉
  currentUserToken = localStorageLoginToken; 
}
```

for me, it 's not really clear, because client side apollo only needs `global.localStorage['Meteor.loginToken']` in `networkInterface.use`

# but... yeah

yeah, now just two lines different. and you're right. may be merge to one file is better for now.

comus:
just finished!

PR #1565, opened 2017-02-07
renderContext fix

comus:


comus:
Today I introduce something about client side rendering/server side rendering lifecycle.

I show you the client/server rendering order.

### 1. first, before react-router render it

the entry point:

client side: `nova-lib/lib/client/render_context.js`

- create renderContext global variable, it contains apolloClient, store...

server side: `nova-lib/lib/server/render_context.js`

- create renderContext `Meteor.EnvironmentVariable`, it contains apolloClient, store...

### 2. second, before react-router render it

client side: `nova-core/lib/client/xxx.js`

- you can modify/access the those global variable inside renderContext
- ex. `const { store } = getRenderContext()`

server side: `nova-core/lib/server/xxx.js`

- you can modify/access the those global variable inside renderContext
- ex.

```js
withRenderContext(() => {
  const { store } = getRenderContext()
})
```

> if you like, you can write above code in `modules/index.js`, because both client/server side have `withRenderContext`
> 
> ex. for both side: `nova-core/lib/modules/xxx.js`
> 
> ```js
> withRenderContext(() => {
>   const { store } = getRenderContext()
>   ...
> })
> ```

### 3. other package can use it as step 2

### 4. finally, react-router render it

on client side: react-router render it when DOM loaded

on server side: react-router render it when last WebApp.connectHandlers middleware trigger

client side: `nova-routing/lib/client/router.js`

- when DOM loaded, react-router render it, ends.

server side: `nova-routing/lib/server/router.js`

- when the final WebApp.connectHandlers middleware runs, react-router render it, ends.


---

what is renderContext?

actually you can get apolloClient with `withApollo` in react component

or on server side you can get all store things use req.

so, why i think telescope needs renderContext?

because

1. for ssr, i can initilize all global variable in nova:lib, the same as client with the same entry point.

2. i want to keep the same coding style as client side.

    if i don't use renderContext, SSR can get store with `req.store`. and client side can get store with global variable `store`.
    
    now, use renderContext, both can get store with
    
    `const { store } = getRenderContext()`
    
3. use renderContext, the lifecycle of SSR almost the same as client side.

i hope that you understand what i talk, ye... my english is not good.

dominictracey:
Is this information going to be in the documentation somewhere? It is super interesting and I think I am going to need it. Our "Standings" widget (and several others under development at [live.rugby.net](http://live.rugby.net/c/aviva-premiership) fetches data from a REST endpoint and it would be great to include that in SSR.

packages/nova-lib/lib/server/render_context.js
```diff
@@ -31,23 +55,49 @@ const Context = function Context(loginToken) {
   }
 };
 
+export const withRenderContextRaw = (func, options = {}) => {
+  const newFunc = Meteor.bindEnvironment((req, res, next) => {
```
xavcz: So now every fiber is aware of both the `loginContext` that gives access to `this.userId` & `renderContext`. That's great!!

packages/nova-lib/lib/server/render_context.js
```diff
@@ -31,23 +55,49 @@ const Context = function Context(loginToken) {
...
+    Fiber.current._meteor_dynamics = Fiber.current._meteor_dynamics || [];
+    Fiber.current._meteor_dynamics[DDP._CurrentInvocation.slot] = req.loginContext;
+    Fiber.current._meteor_dynamics[renderContext.slot] = req.renderContext;
+
+    func(req, res, next);
+
+    if (options.autoNext) {
+      next();
```
xavcz: Go to next middleware automatically 🌮, here especially for the `withRenderContext` middleware

xavcz:
@dominictracey Yeah it will be documented! I'm even thinking about extracting the core logic, stripped from Nova's stuff, and make it a standalone package for the Meteor community who wants to do server-side rendering with Apollo without using deprecated packages!

PR #1566, opened 2017-02-08
Allow hidden form fields

dominictracey:
We have added a "postType" attribute on Post, which is sometimes populated by the results returned by embedly. Videos links, for instance are set to postType=video and our PostPage shows the media embedded. This field is something we don't need members setting so we hide it from them and populate it programatically.

We can't take away the "insertableBy" permission from members because they still do the mutation, but we don't want them changing the value to another that doesn't make sense.


SachaG:
Well, we already have `hidden: true` which removes the field from the form completely, so I wonder if adding the hidden type wouldn't be a bit confusing… I'm also not sure if this is a very "react-y" way of handling things. For example how are you updating this hidden field?

dominictracey:
@SachaG we update the value our custom embedly callback, but maybe your hidden: true method is the way to go. I didn't see anything in the docs regarding it. Is this something that is set on the field's schema?

SachaG:
But I assume that hidden field is in a different component than the Embedly one, so I don't know if a hidden field would help?

You could look at the way the default Embedly component uses `this.context.updateCurrentValues()` to update one field from another. As long as a value is in `this.state.currentValues` it should be submitted with the form, no matter whether there is an actual form item or not.

dominictracey:
Ok - that works. So maybe I should do a PR on nova-docs to include: 

Schema > Forms Properties > hidden

😄

SachaG:
Yeah a PR to the docs would be great!

PR #1569, opened 2017-02-12
better core/lib improvement and update

comus:
- move inject-data from routing to lib
- update apollo.js from ref: https://github.com/apollographql/meteor-integration
- better configureStore function update
- add Utils.findWhere and Utils.defineName
- move apollo_server from core to lib
- can add webapp connect handlers middleware with order number, so it will run by order
- ssr store initialization same as client, it init in nova:lib
- `GraphQLSchema.addToContext({ getViewableFields: Users.getViewableFields })` move to nova:user

and something i want to do, but no time do in this pr
- remove meteorhacks_inject-initial because we have inject-data


SachaG:
That's awesome!

packages/nova-lib/lib/modules/apollo.js
```diff
@@ -1,60 +1,83 @@
...
   const uri = Meteor.absoluteUrl(
-    path, 
-    { rootUrl: getSetting('developmentServerIp', Meteor.absoluteUrl()) }
+    path,
+    { rootUrl: getSetting('developmentServerIp', Meteor.absoluteUrl()) },
   );
-  const networkInterface = createNetworkInterface({
+
+  // allow the use of a batching network interface; if the options.batchingInterface is not specified, fallback to the standard network interface
+  const interfaceToUse = config.batchingInterface ? createBatchingNetworkInterface : createNetworkInterface;
```
xavcz: Sweet the alignment with `apollo` atmosphere package 👍 

Now we need to configure `dataloader` to batch queries server-side too 😄

packages/nova-routing/lib/server/routing.jsx
```diff
@@ -40,6 +40,7 @@ Meteor.startup(() => {
     },
     wrapperHook(req, res, appGenerator) {
       const { apolloClient, store } = getRenderContext();
+      store.reload({ message: 'reload store before render' });
```
xavcz: why reloading the store server-side?

packages/nova-lib/lib/server/apollo_server.js
```diff
@@ -131,19 +102,54 @@ const createApolloServer = (givenOptions = {}, givenConfig = {}) => {
...
+    name: 'graphQLServerMiddleware_bindEnvironment',
+    order: 30,
+  });
 };
 
-Meteor.startup(function () {
+// createApolloServer when server startup
+Meteor.startup(() => {
+  // hello
+  GraphQLSchema.addQuery(`
```
xavcz: nice, but I don't believe we need that 👍

packages/nova-lib/lib/server/meteor_patch.js
```diff
@@ -0,0 +1,98 @@
+import { Meteor } from 'meteor/meteor';
+import { WebApp } from 'meteor/webapp';
+
+import { Utils } from '../modules/index.js';
+
+// our bindEnvironment keep the original function and auto rename the return function
+export const bindEnvironment = (fn) => {
```
xavcz: I understand that having custom `Meteor.bindEnvironment` & `WebApp.connectHandlers.use` brings nice improvements, but do we really need that?

IMO, we should minimize how we patch Meteor 😳

packages/nova-lib/lib/client/render_context.js
```diff
@@ -1,56 +1,51 @@
...
-const reducers = addReducer({ apollo: apolloClient.reducer() });
-const middlewares = addMiddleware(apolloClient.middleware());
-
-const STORE_RELOADED = 'STORE_RELOADED';
+// init context
+const context = {
+  history,
+  loginToken,
+  apolloClient,
+  addAction, // context.addAction same as addAction
```
xavcz: what's the end goal of adding all these utils to the context and not doing like before?

packages/nova-lib/lib/modules/redux.js
```diff
@@ -1,25 +1,78 @@
...
+      getReducers = options.reducers;
+      options.reducers = undefined;
+    }
+    if (!options.reducers && getReducers) {
+      options.reducers = getReducers();
+    }
+    if (options.reducers) {
+      reducers = typeof options.reducers === 'object' ? combineReducers(options.reducers) : options.reducers;
+    }
+    this.replaceReducer(reducers);
```
xavcz: why replacing reducers? because of apollo reducers? I don't get it sorry 😳

packages/nova-lib/lib/modules/utils.js
```diff
@@ -405,37 +405,44 @@ Utils.encodeIntlError = error => typeof error !== "object" ? error : JSON.string
...
-  } catch(__) { 
+    return parsedError;
+  } catch(__) {
     // the error is not internationalizable
     return error;
   }
 };
+
+Utils.findWhere = (array, criteria) => array.find(item => Object.keys(criteria).every(key => item[key] === criteria[key]));
+
```
xavcz: 👍

packages/nova-lib/lib/server/inject_data.js
```diff
@@ -24,19 +28,20 @@ export const InjectData = {
...
     // so we simply turn off injecting if we detect an cors header
     // read more: http://goo.gl/eGwb4e
-    if (res._headers && res._headers['access-control-allow-origin']) {
-      const warnMessage =
-        'warn: injecting data turned off due to CORS headers. ' +
-        'read more: http://goo.gl/eGwb4e';
-      console.warn(warnMessage); // eslint-disable-line no-console
-      return;
-    }
+    // if (res._headers && res._headers['access-control-allow-origin']) {
```
xavcz: so, you believe we are good without that?

packages/nova-users/lib/server/graphql_context.js
```diff
@@ -0,0 +1,4 @@
+import { GraphQLSchema } from 'meteor/nova:lib';
+import Users from '../modules.js';
+
+GraphQLSchema.addToContext({ getViewableFields: Users.getViewableFields })
```
xavcz: 👍

packages/nova-routing/lib/server/routing.jsx
comus: because i am afraid that in the future some other-people-custom-package may want hack/add other reducers to store before render it. yes currently SSR only use apolloClient.middleware. this code just i use to ensure reload store before render it. because sometimes i may forgot reload store after added a reducer. so, yes it can be delete

packages/nova-lib/lib/server/apollo_server.js
comus: for some reason, because i just use core packages of telescope. no posts, no comments...
when i want to create a new project with telescope core packages. the apollo server will alert a error that no query on it, so these code just ensure at least i have one query. so that i run the project created with telescope core packages will no error about this at first time.

xavcz: nice catch!!! 👍

packages/nova-routing/lib/server/routing.jsx
dominictracey: Yes - we have action creators/reducers/middleware in our implementation that I would like to include in the SSR mechanism.

packages/nova-lib/lib/server/meteor_patch.js
comus: yes, i know that. 
i explain why i use these.
custom bindEnvironment i use for keep the original function, also  for get/set their function name for debug to see how many webapp.connectHandlers.stack item on it.
and
another custom webAppConnectHandlers, the improvements is i can add webapp connectHandler with a order number, so finally i can run it by order.
why i write this, because inject-data-middleware must put it to the last,
and i placed inject-data to the nova:lib for other package can use it.
if i don't patch webapp with order feature,
inject-data can not placed in nova:lib
and finally other packages can't use it.
and then i can't remove meteorhacks_inject-initial.
yes, i accept your opinion, we must minimize. and i know i wrote too many code on it.
may be we can think another way to do something like this

packages/nova-lib/lib/client/render_context.js
comus: just want to keep the client side context's function same as server side context's function
on server side.
addAction for add an action to server's actions array
context.addAction for add an action to context's actions array
and then when SSR a request come
the final actions = server's actions array + context's actions array

so i think keep client side context with the same structure is better.
and yes. it is not necessary

packages/nova-lib/lib/modules/redux.js
comus: yes.
for example, on client side,
currently i initialise a empty store object at first in nova:lib
the main reason is i don't want init store in nova:routing.
init a empty store in nova:lib it is good for other packages to access it.
so if you add several reducers from other package.
you must replaceReducer and let the store update with the added reducers

packages/nova-lib/lib/server/inject_data.js
comus: no, i just not very familiar with cors security things. keep and rollback this warning would be better

packages/nova-lib/lib/server/meteor_patch.js
xavcz: Thanks for the explanations, I get the point!! 👌  

Though when packages depend explicitly on `core` instead of `lib`, they cannot access it? 

> i know i wrote too many code on it.

Definitely not too many code dude! ✌️ _After writing thought: when I said minimize, it was about minimizing the impact of patching, not minimizing code_ 😄

packages/nova-lib/lib/client/render_context.js
xavcz: > so i think keep client side context with the same structure is better.

makes sense 👌

dominictracey:
So this PR (I've pulled from comus/devel) causes problems with our extensions. We have a redux-thunk based middleware and a set of actions/reducers to interact with our domain-specific REST api that pulls entities into the same store as apolloClient uses.

From a behaviour perspective, the first issue I see:

We have extended the category schema and added in some queries, mutations, fragments, etc. and the edit form for categories is now unpopulated when the application first starts. This seems to resolve itself after a minute or so. The browser console shows an error `Warning: Unknown prop `slug` on <option> tag. Remove this prop from the element.` when the dialog is unfilled. Once the problem resolves itself, the warning no longer is logged. I speculate the SSR (or pre-hydrated store) doesn't have our schema extensions or something? Then when the query is re-executed on the client, the additional fields are supplied.

xavcz:
> Warning: Unknown prop slug on <option> tag. Remove this prop from the element. 

This is because the `<option>` tag only excepts a `value` & `label` 😄 

> I speculate the SSR (or pre-hydrated store) doesn't have our schema extensions or something?

Oh interesting!

dominictracey:
@xavcz I'm trying to understand the apollo SSR strategies and how they are leveraged in Telescope so just staggering around a little. 😅  So I have no idea what the root cause is of the `<option>` tag error. I can't replicate it in our [production instance](http://live.rugby.net/) so I think it's been introduced with the PR?

In terms of the inject_data.js, it also looks like I'm getting two identical `<script type="text/inject-data">` tags in the head of the SSR-ed html? I'll keep looking at this, but am feeling a little overwhelmed.

dominictracey:
This checks out great with the TRN packages. Thanks so much @comus!

packages/nova-lib/lib/modules/redux.js
```diff
@@ -70,9 +70,13 @@ export const getReducers = () => reducers;
 // **Notes: server side, addMiddleware to server share with every req**
 let middlewares = [];
 
-export const addMiddleware = (middlewareOrMiddlewareArray) => {
+export const addMiddleware = (middlewareOrMiddlewareArray, options = {}) => {
   const addedMiddleware = Array.isArray(middlewareOrMiddlewareArray) ? middlewareOrMiddlewareArray : [middlewareOrMiddlewareArray];
-  middlewares = [...middlewares, ...addedMiddleware];
+  if (options.unshift) {
```
xavcz: thunk 🥇

packages/nova-lib/lib/client/render_context.js
```diff
@@ -1,56 +1,52 @@
...
-    return context;
-  },
-};
+// init store
+context.store = configureStore(context.getReducers, {}, (store) => {
+  let chain, newDispatch;
+  return next => (action) => {
+    if (!chain) {
+      chain = context.getMiddlewares().map(middleware => middleware(store));
+      newDispatch = compose(...chain)(next)
```
xavcz: just adding notes to help document the process later, thunk & init dispatch

PR #1570, opened 2017-02-19
add explicit control for clearing out form fields

cenkce:
- create clearField(fieldName) method on Form.jsx. clearField should add form name to a new this.state.deletedFields array

- create unclearField(fieldName) method (find better name?) that removes field from the deletedFields array 

- pass "clearField" callback from Form.jsx down to the form component

- if a field value is modified after it has been cleared, unclearField should be called to remove the field name should be removed from deletedFields

- modify the submitForm method in Form.jsx to also add all deletedFields fields to the unset object

### DIFF
--- a/packages/nova-forms/lib/Form.jsx
+++ b/packages/nova-forms/lib/Form.jsx
@@ -65,12 +65,15 @@ class Form extends Component {
     this.deleteDocument = this.deleteDocument.bind(this);
     // a debounced version of seState that only updates state every 500 ms (not used)
     this.debouncedSetState = _.debounce(this.setState, 500);
+    this.clearField = this.clearField.bind(this);
+    this.unclearField = this.unclearField.bind(this);
 
     this.state = {
       disabled: false,
       errors: [],
       autofilledValues: (props.formType === 'new' && props.prefilledProps) || {},
-      currentValues: {}
+      currentValues: {},
+      clearedFields: []
     };
   }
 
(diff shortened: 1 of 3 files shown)


SachaG:
Thanks for this! But is it `clearedFields` or `deletedFields`? looks like you have both in there?

packages/nova-forms/lib/Form.jsx
```diff
@@ -333,6 +336,22 @@ class Form extends Component {
...
     }));
   }
+  
+  // add to cleared fields
+  clearField(fieldName){
+    this.setState({
+      deletedFields: [...this.deletedFields, fieldName]
+    });
+    
+    return () => this.removeFromDeletedfields(fieldName);
```
xavcz: Interesting pattern! However how would you trigger the returned function after someone fills again the component?

Btw, `removeFromDeletedfields` isn't defined 👌

packages/nova-forms/lib/Form.jsx
```diff
@@ -333,6 +336,22 @@ class Form extends Component {
...
+  }
+
+  // remove from cleared fields
+  unclearfield(fieldName){
```
xavcz: typo: `unclearfield` ➡️ `unclearField`

packages/nova-forms/lib/Form.jsx
```diff
@@ -492,6 +511,9 @@ class Form extends Component {
           onKeyDown={this.formKeyDown}
           disabled={this.state.disabled}
           ref="form"
+          clearField={this.clearField}
+          unclearField={this.unclearField}
```
xavcz: On the general idea: it maybe not needed to pass `unclearField` to the form components? What about calling when  `updateCurrentValues` is triggered? ➡️ filter `clearedFields` in comparison of `newValues`'s keys

On the implementation: these additions should be passed to `FormGroup` few lines below and then to `FormComponent` (https://github.com/cenkce/Telescope/blob/085253fb003b669c0a355b9a117b7e38d9d9a655/packages/nova-forms/lib/FormGroup.jsx#L9)

packages/nova-forms/lib/Form.jsx
cenkce: ok sorry i forgot to rename

packages/nova-forms/lib/Form.jsx
cenkce: Actually I don't know how exactly system works. I try to learn your by feedbacks.

packages/nova-forms/lib/Form.jsx
cenkce: this pattern from event subscription, for example

var unsubscriber = eventEmitter.subcribe(...); I suppose it could be helpful.

PR #1572, opened 2017-02-22
Add explicit control for clearing out form fields

cenkce:
Test url : https://telescopeapp-cenkce.c9users.io/

### DIFF
--- a/packages/nova-forms/lib/DateTime.jsx
+++ b/packages/nova-forms/lib/DateTime.jsx
@@ -12,11 +12,12 @@ class DateTime extends Component {
 
   // when the datetime picker has mounted, SmartForm will catch the date value (no formsy mixin in this component)
   componentDidMount() {
-    this.updateDate(this.props.value || new Date());
+    this.updateDate(this.props.value || new Date(), true);
   }
 
-  updateDate(date) {
-    this.context.updateCurrentValues({[this.props.name]: date});
+  updateDate(date, isDefault=false) {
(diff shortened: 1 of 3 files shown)


SachaG:
Nice progress! I would love to hear more about your thought progress though. For example why did you add logic in `FormComponent` instead keeping everything in `Form`? And why did you need to modify `DateTime`?

Ideally `FormComponent` should remain super simple, its job is only selecting the right component to display, not managing any kind of internal state. So the only stateful component here should be `Form`.

SachaG:
Oh also what does `isDirty` do?

cenkce:
isDirty state means component has a new value. Because FormComponent is the container component and I think it can manage data flow. Why did i modify datetime component? Beacuse 
1- this.context... is the experimental feature
2- Component doesn't fire onBlur event. Components' data flow is not standard and to my way of thinking this is the architectural problem. Some components' value is updated by FormComponent, others' value is updated by its own. It is confused. I think Single data flow is better solution.

FormComponent is the kind of decorater and container component and in my opinion The FormComponent logic (updating component value, adding clear button and clearing new value) is encapsulated within.

SachaG:
Do you think you could modify the PR to keep `FormComponent` stateless?

cenkce:
No problem I can do

xavcz:
How I are you doing @cenkce?

xavcz:
I believe we can close this, due to inactivity & conflict. Don't hesitate to re-open a new one or ask feedback @cenkce if you work again on that!! 🚀

PR #1584, opened 2017-03-11
Fixed downvoting by modifying vote.js file in nova-voting package

Discordius:
Upvoting worked fine, but there were some bugs with the canceldownvote functionality, and in particular the automatic canceling of upvotes when one downvotes and of downvotes when one upvotes was broken. This should be fixed with the edits below, and seemed to do well in our local testing.


SachaG:
Thanks for catching that! By the way, it's best if PRs are made to the `devel` branch rather than `master`.

SachaG:
I merged it myself manually on devel btw :)

PR #1587, opened 2017-03-15
use npm simpl-schema, meteor aldeed:collection2-core package

comus:


SachaG:
Oh that's awesome, thanks so much for taking care of this! Anything we should know about this? I remember @xavcz ran into some issues when he tried to do the same conversion but I can't really remember what they were…

Also I guess we need to specify `upvoters.$` fields manually now? Is there any way around that?

comus:
1. type: [String] to type: Array
> Also I guess we need to specify upvoters.$ fields manually now?

yes, see: https://github.com/aldeed/node-simple-schema/issues/66#issuecomment-283703385

and 

https://github.com/cult-of-coders/grapher/issues/114#issuecomment-285976880

2. SimpleSchema.extendOptions(Object) to SimpleSchema.extendOptions(Array)

```
SimpleSchema.extendOptions({
  control: Match.Optional(Match.Any),
})
```

become

```
SimpleSchema.extendOptions(['control'])
```

3. `field.type.name` to `field.type.singleType.name`

because now they support one field mutliple type

4. no `decimal: true` option

`type: Number` means `float` by default

Integer use `type: SimpleSchema.Integer`

5. i am afraid that change some fields to `SimpleSchema.Integer` (ex. order) may cause `forms` not works, so currently no `SimpleSchema.Integer`, everything is Number

PR #1589, opened 2017-03-18
a make sense modified std:accounts-ui for telescope

comus:
you may know that std:accounts-ui have some bugs/problems that not very good for telescope use case.

here is what i modified

1. add local copy from std:accounts-ui@1.2.19
2. delete some std:accounts-ui files to make the local version more clean first before modify
3. to reduce the code size, i removed std:accounts-ui's login-without-password feature, because i dont use this feature too.
4. before enable loginform's buttons, i want it wait on `Accounts.loginServicesConfigured` first. for example, after DOM loaded, if i click on google/facebook login button immediately, it may cause an error, because meteor ddp loginServices not loaded.

    see: https://www.dropbox.com/s/y0825fu4dmfapi1/Screenshot%202017-03-19%2001.47.27.png?dl=0

5. better error handling
    for example, if i click google/fackbook login button and then the login window popup, then i close it, it will not cause an error now, because it caught.
    and... `something went wrong` is not a good error message
    ```
        if (error instanceof Accounts.LoginCancelledError) {
          // do nothing
        } else {
          console.warn(error.message || error)
          this.showMessage((error.reason && T9n.get(`error.accounts.${error.reason}`)) || T9n.get('Unknown error'))
        }
    ```
6. fix ‘or use …’ still displayed after login

7. this is most important!!! currently telescope nova use graphql, and ssr loaded currentUser.
    so i think accounts-ui may use currentUser too.
    for example: ssr send currentUser to client/browser, accounts-ui start subscribe currentUser from ddp. accounts-ui get currentUser from ddp. it is not make sense. and cause react client/server side rendering not match. because the method they get the currentUser are different.

8. show a logging in message when logging in.

9. currentUser can return user's services object. it is safe. so that client/browser knows which services the user have, if user has password service, so profile can show changePassword button

i know that i can submit a PR to std:accounts-ui, but the use case between telescope nova and normal meteor app become more different. i mean graphql/apollo.

so i temporarily create a `_accounts-ui` packages here.


SachaG:
That's a great idea. I think regarding accounts, we need to do two things. First, use Apollo to transfer user data so we can drop the dependency on DDP. 

And later, migrate to PassportJS or Auth0 or something so we can drop the dependency on Meteor accounts altogether. 

So this is a great step in that direction!

comus:
That's what I think too. Yes. It's the step 1.

PR #1596, opened 2017-03-27
Enable facebook sharing of posts by supporting facebook scraper reqs

dominictracey:
Here is the debugging tool: https://developers.facebook.com/tools/debug/

This PR:
- changes SSR code to serve up rendered content (including meta head tags from helmet) to facebook scraper.
- changes default thumbnail image to minimum size supported by facebook (200x200)

xavcz/vulcan-social-share provides drop in social media sharing buttons that don't really work without these changes.


dominictracey:
@comus - here is the render_context.js change.

dominictracey:
@sacha and @xavcz - I don't know if changing the aspect ratio of the default thumbnail is acceptable. The current 125x200 gets kicked out by the facebook scraper currently so maybe we could go up to 200x320 if we wanted to meet the minimum size requirements but maintain the current aspect ratio?

PR #1601, opened 2017-03-31
Semi-colon Updates

claudey:
I doubt the app will ever break because of those semi-colon omissions but I added them, just to be safe.


SachaG:
Thanks!

Issue #1606, opened 2017-04-06
How can i ensure that the user has logged in first before navigating to specific route


iamrommel:
Let say i have the route "/setting/adminOnly", when my user copy paste the URL to their browser, i should have the logic to that the user needs to login first before going to this route (if he is not logged in yet).

But i don't know where to put this ensurelogin logic, iron-router has that feature, i'm not sure how it should be implemented in vulcanjs. Any thoughts/hint?

SachaG:
I would do that check inside the React component and redirect in the `componentDidMount` callback. 

React-Router might have a route callback feature, but I'm not sure if it would work inside Vulcan…

meteorplus:
@iamrommel check this out https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Redirect.md

iamrommel:
I like the implementation here https://medium.com/the-many/adding-login-and-authentication-sections-to-your-react-or-react-native-app-7767fd251bd1
specially the EnsureLoggedInContainer, i just that I'm not sure how to create child route of specific path.

If only I can have something like this
```
addRoute({name: 'app', path: '/app', componentName: 'EnsureLogin'});
addChildRoute('app', { {name: 'app.workorder', path: '/app/workorder', componentName: 'Layout'}}  )

```
that will create the child of existing route 'app', which will generate a react-router route like this

```
 <Route component={EnsureLogin}>
    <Route path=”/app/workorder” component={Layout}/>
  </Route>

```

I can have the flexibility to define in my route which should be private and public. Which i guess a good addition in vulcan:routing.

I hope I can make a PR for this.

iamrommel:
This addChildRoute will be helpful also if we want to define sub-routes in our route.

SachaG:
I don't think I've used child routes before. What's the difference between:

```
addRoute({name: 'app', path: '/app', componentName: 'EnsureLogin'});
addChildRoute('app', { {name: 'app.workorder', path: '/app/workorder', componentName: 'Layout'}}  )
```

and

```
addRoute({name: 'app', path: '/app', componentName: 'EnsureLogin'});
addRoute({name: 'app.workorder', path: '/app/workorder', componentName: 'Layout'})
```

?

iamrommel:
Yep, there is no child route yet, that is the name of method what i'm thinking to implement.

[...]

PR #1609, opened 2017-04-07
Devel

iamrommel:
As discussed on #1606 , i did try to implement the addRoute passing extra argument for parentRoute name. I was able to create the subRoute now

I'm not sure if the code works as your expectation and guidelines as your 3rd item from https://github.com/TelescopeJS/Telescope/blob/devel/CONTRIBUTING.md says it should be on their own repo, so please let me know what are your thoughts


SachaG:
I didn't have time to test the feature, but the code looks ok to me so I went ahead and merged it. Thanks for the PR :)

PR #1612, opened 2017-04-07
Should call the property of parentRouteName properly

iamrommel:
I appreciate that you edited the code and point me out that it should throw error. But i think the property of parentRouteName properly should use "[]" and not by "." notation.

I hope we can setup the test and CI for this  as soon as possible so we can guard stuff like this with test.

Still this is related to issue #1606


SachaG:
Oh of course, thanks for catching that!

PR #1615, opened 2017-04-10
Allow prefilledprops to fill edit forms 

dchan14:
so that if we manipulate a document using (drag/ drop or other functions) before displaying to user i would like to save it to database instead of pulling old information before the change.


SachaG:
It looks like something went wrong with that PR, the commits included here are all my own. I just updated the `devel` branch, maybe you can pull again and try another PR?

dchan14:
My fault still working on the code will do pull and upload new pr

Dc

PR #1616, opened 2017-04-10
fix bug where All Categories didn't clear the category filter

ESRogs:
The "All Categories" link in the category option used to set the user's query to whatever it already was. With this change, the query is set to a modified copy of the current query that has category filters removed from it.


SachaG:
Nice, thanks! By the way, please make future PRs to `devel` if you can :)

PR #1617, opened 2017-04-12
Package did not match the version selected

claudey:


SachaG:
Thanks, but it looks like you also commited a vscode file by mistake?

SachaG:
I think it's probably easier for me to just change the version number directly :)

PR #1620, opened 2017-04-18
remove LinkContainer it breaks the react boostrap

meteorplus:
Using href of MenuItem instead of LinkContainer since it break react boostrap design!

### DIFF
--- a/packages/vulcan-base-components/lib/categories/Category.jsx
+++ b/packages/vulcan-base-components/lib/categories/Category.jsx
@@ -26,15 +26,14 @@ class Category extends Component {
 
     return (
       <div className="category-menu-item dropdown-item">
-        <LinkContainer to={{pathname:"/", query: newQuery}}>
           <MenuItem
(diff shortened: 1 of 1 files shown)


SachaG:
How does it break it? Also, if we don't use LinkContainer I don't know if React Router will work?

meteorplus:
It breaks the the react boostrap behavior and css classes! the NavItem is already an a tag so the react router will pick it up!

SachaG:
I thought React Router only picked up Link tag, not regular a tags?

justinr1234:
This is outdated.

But for the sake of argument, you would override `onClick`. As @SachaG mentioned, the `<LinkContainer>` is meant to allow integration with `react-router` specifically so that we can do `history.push` behind the scenes. `<LinkContainer>` is a wrapper around `<Link>` that comes from `react-router`. Using a straight `href` will cause the page to reload, which breaks the Single Page Application design of Vulcan.

PR #1621, opened 2017-04-19
Fix meteor installation on Linux

inside:
* chmod u+x prestart_nova.sh
* Invoke the above script not using sh

The reason for this is that I got this error when doing `npm start`:

```
% npm start

> Nova@1.2.0 prestart /home/inside/src/Vulcan
> sh prestart_nova.sh

Vulcan requires Meteor but it's not installed. Trying to Install...
prestart_nova.sh: 18: [: Linux: unexpected operator
prestart_nova.sh: 22: [: Linux: unexpected operator
prestart_nova.sh: 26: [: Linux: unexpected operator
🛠  Creating your own settings.json file...

🌋  Happy hacking with Vulcan!
📖  The docs are available at: http://docs.vulcanjs.org

> Nova@1.2.0 start /home/inside/src/Vulcan
> meteor --settings settings.json

sh: 1: meteor: not found
```


meteorplus:
this won't work with the ./

the issue here is the sh doesn't support the syntax we are using for check the OS type!

but it will work using bash that's why we're loading bash in top the of the file!

So the solution would be "prestart": "bash prestart_nova.sh",

inside:
What is the error message you have with the ./ ?

`#!/usr/bin/env bash` in prestart_nova.sh is supposed to invoke the script with bash. There should be no need to do `bash prestart_nova.sh`.

meteorplus:
dude I tried it on my linux workspace and it break up but try this one and it will work! I don't want to make another pull request just for this change it please thanks

it was already forking with sh prestart_nova.sh

but after adding the os detection the syntax changed and it breaks in sh

inside:
Hi man!

The ./ is working on my ubuntu and so does the prefix with bash. So here's the commit!

Let me know.
Best,

meteorplus:
thanks bro merci :)

inside:
Haha, je vous en prie :+1:

defionscode:
@meteorplus seems like switching out sh for bash has been PRd at least a couple of times. How can we get that change merged in? (asking due to your Contributor tag)

justinr1234:
Outdated and prestart is disabled: https://github.com/VulcanJS/Vulcan/commit/bbd2dbd0063d15a8ec174fa452f3091c85049635#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

PR #1624, opened 2017-04-20
missing package and prestart fix

meteorplus:
"prop-types": "^15.5.8",

and prestart to use bash instead of sh since bash works on both linux and mac os

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "npm": "^3.0"
   },
   "scripts": {
-    "prestart": "sh prestart_nova.sh",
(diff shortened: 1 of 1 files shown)


defionscode:
Looks like prop-types is already in the upstream.

justinr1234:
Oudated and prestart is disabled: https://github.com/VulcanJS/Vulcan/commit/bbd2dbd0063d15a8ec174fa452f3091c85049635#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

PR #1627, opened 2017-04-26
Add embedly and forms dependencies to vulcan-base-components

meshulam:
The vulcan-base-components package assumes that Posts have a `thumbnailUrl` field. This field is provided by vulcan-embedly, so if you include the base components in a project without including embedly, you get a GraphQL error:

    Error: GraphQL error: Cannot query field "thumbnailUrl" on type "Post".

Similarly, base-components rely on the SmartForm component from vulcan-forms.


justinr1234:
Outdated, base components is no longer in core.

PR #1628, opened 2017-04-26
Fix for React bootstrap Styles

meteorplus:
Fix for React bootstrap Styles with the correct version!
thanks


justinr1234:
Outdated and we are changing the direction of the UI. Discussion here: https://github.com/VulcanJS/Vulcan/issues/1823

PR #1631, opened 2017-05-04
Added support for Embed API

meteorplus:
Added support for Embed API depending on the key length
no need to add anything or change anything! not need to add any package just add ytour embed api key on embedlyKey : "your_key_here" and it will work right away

### DIFF
--- a/packages/vulcan-embedly/lib/server/embedapi.js
+++ b/packages/vulcan-embedly/lib/server/embedapi.js
@@ -1,0 +1,48 @@
+import {  getSetting } from 'meteor/vulcan:core';
+
+export const embedAPI = function(url){
+    
+      var embedAPIKey = getSetting('embedAPIKey');
+      var thumbnailWidth = getSetting('thumbnailWidth', 200);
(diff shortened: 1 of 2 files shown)


SachaG:
Thanks! But it seems like doing things this way results in a lot of duplicate code. Maybe you could refactor the function so that it has better flow and is easier to understand?

SachaG:
Oh also you definitely can't do `embedlyKey.length > 32`, that's way too hacky! Please use a separate setting name :)

meteorplus:
ok thanks :)

I will clean it up and refactor it

meteorplus:
okay cleaned it up and made a different file for the embed api call

SachaG:
Won't this throw an error if people have no Embedly or EmbedAPI keys at all?

SachaG:
I think you should rethink the code organization a little bit. Try to write it in a way that will also be easy to extend in the future if people want to add a third similar service.

SachaG:
Actually I'll refactor the package myself in a way that makes it possible to add additional services. Once I'm done you can make another PR to integrate EmbedAPI cleanly :)

PR #1640, opened 2017-05-23
Added PropType package and changed from component to PureComponent

MHerszak:


package.json
```diff
@@ -36,6 +36,7 @@
     "history": "^3.0.0",
     "html-to-text": "^2.1.0",
     "immutability-helper": "^2.0.0",
+    "import": "0.0.6",
```
xavcz: Is it needed?

package.json
```diff
@@ -47,6 +48,7 @@
     "mingo": "^0.8.1",
     "moment": "^2.13.0",
     "optics-agent": "^1.0.5",
+    "prop-types": "^15.5.10",
     "react": "^15.4.1",
```
xavcz: Should we update React & React DOM also to the latest?

package.json
MHerszak: @xavcz This import is not needed and I did not add it to the package.json. I merged with devel first and thought it came from there.

package.json
MHerszak: @xavcz I'd say yes. There are no real "threats" until v16.

package.json
xavcz: 👍

package.json
xavcz: 👍

SachaG:
Should I merge this? Or are you planning other changes?

MHerszak:
@SachaG yes merge it. I will probably keep adding changes. Not today or tomorrow, though.

PR #1644, opened 2017-05-27
Fix bug in vulcan-core/lib/modules/containers/withList.js

adalidda:
This PR fixed the bug in vulcan-core/lib/modules/containers/withList.js 

instead of
  const { collection, limit = 10, pollInterval = 20000 } = options,

should write

  const { collection, limit = getSetting('postsPerPage', 10), pollInterval = 20000 } = options,

I can't attach withList.js so I transform it into withList.js.txt.

It's my first PR so sorry in advance if I make a mistake.

Thank You
Adalidda

[withList.js.txt](https://github.com/VulcanJS/Vulcan/files/1033241/withList.js.txt)


SachaG:
Thanks! By the way, in the future please make PRs to the `devel` branch. I'll take care of this one for now :)

PR #1647, opened 2017-05-30
Use Boostrap V 3.3.7!

meteorplus:
Please I already have asked this many times

Use Boostrap V 3.3.7  which is the only supported version on React bootstrap otherwise everything looks weird and doesn't work properly!


iamrommel:
Does replaceComponent does not work to override the default implementation

SachaG:
Yes, replaceComponent should work. Maybe we can add a removeMetaTag function or something as well.

meteorplus:
it works but the react bootstrap component are built for bootstrap 3 not ' thus they don't work or behave as they are meant to!

Bootstrap 3 components built with React.
https://github.com/react-bootstrap/react-bootstrap

ayhid:
@meteorplus  shoud now you can add the styles you want  by overriding head tags checkout the documentation here
http://docs.vulcanjs.org/head-tags.html

justinr1234:
Outdated

PR #1651, opened 2017-06-16
use correct code for Mailchimp "already subscribed" state

klokie:
If the user is already subscribed, the return code is numeric `214`, returned as `code` rather than `message`.


SachaG:
Thanks! By the way, in the future please make PRs to the `devel` branch :)

klokie:
Ok! Guess I should have read "CONTRIBUTING.md" 😜
Cheers

PR #1659, opened 2017-06-27
added Swedish translation

klokie:
I added Swedish translation (`vulcan-i18n-sv-se`) based on English (`vulcan-i18n-en-us`). 
Enjoy!


SachaG:
This is great! But I'm sorry, actually we don't add i18n packages directly into the main repo. Instead, you should publish them as a Meteor package under your own name, and then submit a PR to the documentation (https://github.com/VulcanJS/vulcan-docs) to add a link in the i18n section.

PR #1666, opened 2017-07-06
Added failure form callbacks and success form callbacks to forms

Discordius:
We added the `addToSubmitForm` function that's passed down to children of the form components a while ago. This adds two very similar functions that allow a component to add callbacks on form success and form failure.


SachaG:
Awesome!

PR #1673, opened 2017-07-19
Fix remove permission strings

ochicf:
Transform collection name to lower case as it is done with the other permission strings.


SachaG:
Thanks!

PR #1675, opened 2017-07-22
Graphql interfaces

ochicf:
- added interface support for collections
- very simple inline-fragments detection (should be improved before merging)
- added example-interface package


ochicf:
@SachaG you can use the `example-interfaces` to test it.

One thing that need rework and improvement is the inline-fragment detection. It should be much more restrictive. Right now works because subfragments are only considered when written as `...SubFragmentName`, because I use the space just after the `...` to dismiss it and consider it to be an inline fragment. For example, `... SubFragmentName` (note the space) would not be taken as a subfragment.

SachaG:
Thanks, great work! I'll let you know once I have a chance to use interfaces.

PR #1676, opened 2017-07-24
Fix permission check bug for vulcan-core default remove mutation

thisjeremiah:
Fixed permission check bug for vulcan-core default remove mutation by adding a .toLowerCase() call for the action strings .


SachaG:
Thanks, but I think this was already fixed by a previous PR, sorry!

thisjeremiah:
Oops, okay. I must've been checking master, not devel branch.

PR #1678, opened 2017-07-25
add function to remove tags from head

ayhid:


SachaG:
Hmm maybe I'm missing something, but if we push tags to `Headtags` inside `Headtags.jsx`, how will we be able to remove them?

ayhid:
You are absolutely right, I have to find a way to push tags elsewhere and be able to remove them from another package.
In our slack discussion you suggested pushing tags somewhere in the  `vulcan:base-components` do you have more details on the exact place? (sorry I am still discovering vulcan, any guidance is more than welcome)

SachaG:
Could be anywhere really. If you want, I can merge this is first and then take care of moving where the tags are pushed myself?

ayhid:
If it's easier for you, yes please, this way I see where you will do the push.
Thanks

PR #1681, opened 2017-07-27
Providing some an order to the user profiles

Zash007:
Added a logical order: Name, email, bio, web, twitter. Then anything else.


SachaG:
Thanks! Small bit of feedback: maybe instead of 1, 2,3 you could use 10,20,30? This way people can insert their own fields in between. Also, don't hesitate to reorder the actual code in the same order.

Zash007:
Cool - I'll do that instead :)

SachaG:
Thanks! By the way, please make PRs to the devel branch if possible :) I'll merge this one manually for now.

PR #1695, opened 2017-08-08
fixed newsleter and getSetting

Spartano:


SachaG:
Thanks! Just a few comments:

- Notes like `i changed the position of default value because it overides the setting.json values` are more appropriate as PR comments than code comments. 
- Isn't `frequency` supposed to already be an array?

Spartano:
When i console log `frequency` it returns an object.

packages/vulcan-lib/lib/modules/settings.js
```diff
@@ -9,10 +9,11 @@ export const getSetting = (setting, defaultValue) => {
     // if setting is an object, "collect" properties from all three places
     if (typeof rootSetting === 'object' || typeof privateSetting === 'object' || typeof publicSetting === 'object') {
       return {
+        ...defaultValue,
```
Spartano: // i changed the position of default value because it overides the setting.json values
        //example: "newsletterFrequency": [2,3] inside cron.js will be overwriten by the default --> result [1,3]

SachaG:
Thanks! Also, in the future can you make PRs to `devel` instead of `master`? I'll merge this manually myself this time :)

PR #1702, opened 2017-08-11
Update graphql.js

mechanical-turk:
this is a followup on the question i asked on slack. my understanding is that graphql gets confused when we give it `$` or `.`. Since any field that uses `$` in their name will also use `.`, I thought we could just amend the rule so that not only `Array`s get skipped, but also `Object`s.


SachaG:
Hmm, I don't know if it's a good idea to do it this way. If field names cannot contain a `.` I think we should have a separate test that throws an error for that, rather than just ignoring the field.

PR #1703, opened 2017-08-12
[fix] give default value to unset

mechanical-turk:
[fix] give default value to unset


SachaG:
Thanks! Btw, in the future please make PRs to the `devel` branch if possible :)

PR #1709, opened 2017-08-18
Display errors on password reset form

meshulam:
I encountered two problems that prevented field- and form-level errors from displaying on the password reset form.

First, the validatePassword helper sets errors on the 'password' field but the reset form listens for messages on 'newPassword'. 

Second, the resetPassword callback was passing a string to showMessage instead of an I18n key.


SachaG:
Oh, actually I'm afraid this might conflict with a recent commit I just made to `devel`: https://github.com/VulcanJS/Vulcan/commit/03c7a8e4c8e839656b054b767e0aee05044df705

If you can, could you make another PR to `devel` and make sure there's no conflicts?

PR #1711, opened 2017-08-24
Added create-react-class dependecy

meteorplus:
Otherwise it will throw Error: Cannot find module 'create-react-class' at first npm start

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "body-parser": "^1.15.2",
     "classnames": "^2.2.3",
     "cookie-parser": "^1.4.3",
+    "create-react-class": "^15.6.0",
     "crypto-js": "^3.1.9-1",
     "dataloader": "^1.3.0",
     "deepmerge": "^1.2.0",


SachaG:
Where do we use that module? I searched the codebase and couldn't find any reference to it.

meteorplus:
A dependency is using the old way of .createClass and it's throwing an error on the console

W20170904-03:10:19.089(0)? (STDERR) Warning: Accessing createClass via the main React package is deprecated, and will be removed in React v16.0. Use a plain JavaScript class instead. If you're not yet ready to migrate, create-react-class v15.* is available on npm as a temporary, drop-in replacement. For more info see https://fb.me/react-create-class
W20170904-03:10:19.090(0)? (STDERR) Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs

meteorplus:
nope seems we're using it in vulcan forms

/packages/vulcan-forms/lib/row.js:
	9: var Row = React.createClass({

/packages/vulcan-forms/lib/row.js:
	9: var Row = React.createClass({

SachaG:
Oh that file doesn't exist anymore, it was a leftover from an older version.

meteorplus:
@SachaG it's still there

var React = require('react');
import PropTypes from 'prop-types';
var classNames = require('classnames/dedupe');

var Row = React.createClass({

I just cloned it from master and had to install create-react-class to get it working

SachaG:
I don't think it's there on `devel`? is it?

meteorplus:
just master I think...

PR #1713, opened 2017-08-25
RegisterFragment should respect comments in fragment literals

klokie:
The fragment parser doesn't ignore comments if they include `...`, e.g.
```
fragment CategoriesList on Category {
# ...dontMindMe
  }
}
```
This PR should fix that.
Also I've modified the regex for parsing GraphQL fragment names to follow the GraphQL spec more exactly.


SachaG:
That's great, thanks!

PR #1715, opened 2017-08-27
add  terms.selector otherwise it will be ignored

meteorplus:
add selector from terms otherwise it will be ignored and never used!!

this was making me go nuts!!!


SachaG:
Actually, this is on purpose. Accepting a `selector` object from the client can open up big security holes, since it lets the client ask for any object in your database.

meteorplus:
ah ok, merci

PR #1716, opened 2017-08-28
Display errors on password reset form

meshulam:
Re-opening #1709 against devel. Still worked without any changes needed.


SachaG:
Thanks!

PR #1718, opened 2017-09-03
show admin menu to only admins

meteorplus:


justinr1234:
Outdated PR and this was fixed here: https://github.com/VulcanJS/Vulcan/blob/devel/packages/example-forum/lib/components/users/UsersMenu.jsx#L28

PR #1721, opened 2017-09-07
Fix error: Cannot read property twitter of undefined, in edit account

valgalin:
Hi @SachaG ,

I encountered this error when I was trying to add a Twitter username in my account. When I clicked the submit button, it returns this error message:

**Cannot read property 'twitter' of undefined**

So I changed a little code in the `helper.js` file of the `vulcan:users` package. And also in the `UsersProfile.jsx` file.

Please let me know if I did this correctly or not.

Thanks! :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,5810 @@
+{
+  "name": "Vulcan",
+  "version": "1.2.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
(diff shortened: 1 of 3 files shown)


SachaG:
Good catch! But actually we only need the second part to make it work (`user.profile.twitter` to `user.twitterUsername`).

SachaG:
I'll add the fix myself :)

PR #1722, opened 2017-09-08
Added submit option to `updateCurrentValues` in Form.jsx context

Discordius:
I needed some kind of way to submit a form from a child component. However, a straightforward `submitForm` method in context didn't work, since I wasn't able to predict when my changes made via other functions such as `updateCurrentValues` would go through. 

The most elegant solution that I've found was to add an optional flag to `updateCurrentValues` that when set to true submits the form as soon as the values are set. It's not the most elegant solution, but I wasn't able to come up with a better one on the fly. Happy to change this if someone has better ideas.

### DIFF
--- a/packages/vulcan-forms/lib/components/Form.jsx
+++ b/packages/vulcan-forms/lib/components/Form.jsx
@@ -75,6 +75,8 @@ class Form extends Component {
     this.setFormState = this.setFormState.bind(this);
     this.getLabel = this.getLabel.bind(this);
     this.getErrorMessage = this.getErrorMessage.bind(this);
+    // version of submitForm that is made available to context, behaves like updateCurrentValues but submits afterwards
+    this.submitFormContext = this.submitFormContext.bind(this);
 
     this.state = {
       disabled: false,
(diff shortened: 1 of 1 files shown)


SachaG:
What if `submitForm()` took an optional `state` argument and did the `setState` itself?

SachaG:
Or maybe there's a way to do `await this.context.updateCurrentValues({deleted: true});` to make sure `setState` finishes before `this.context.submitForm()` is called?

Discordius:
Yeah, I was thinking about `submitForm()` taking an optional state argument. Also seems fine to me. 

The other problem with `submitForm()` is that we already have an internal function called `submitForm` and so we would either need to come up with a new name for the context function, or have some inconsistency between the naming of the context fields and the names of the function.

Discordius:
> Or maybe there's a way to do await this.context.updateCurrentValues({deleted: true}); to make sure setState finishes before this.context.submitForm() is called?

Since `setState()` doesn't return a promise that sadly wouldn't do anything. One could wrap `setState` in a promise, but that feels ugly to me, especially since `setState` already comes with a second argument that can take a callback that's executed as soon as the state updates. 

However making it so that `submitForm` takes a state argument seems fine to me. Happy to make that change to the PR if that seems reasonable to you.

SachaG:
OK yeah let's go with `submitForm(state)` then. Either that or make `updateCurrentValues` accept a callback, and then call `submitForm` inside *that*. Which would mirror the way `setState` takes a callback, so it might make sense?

SachaG:
Relevant thread: https://github.com/facebook/react/issues/2642

Discordius:
Ok, implemented it as `submitForm(state)`.

SachaG:
So in this implementation we can only change `state.currentValues` on form submit, not anything else on the state. I guess that's probably ok?

PR #1734, opened 2017-09-28
Update allow/deny package

devarthurribeiro:
Update the allow/deny package due to security failure reported at: https://blog.meteor.com/meteor-allow-deny-vulnerability-disclosure-baf398f47b25


SachaG:
Good idea, thanks!

PR #1738, opened 2017-10-05
Dropped isomorphic-fetch in favor of cross-fetch

lquixada:
[isomorphic-fetch](https://github.com/matthew-andrews/isomorphic-fetch) has a bug that prevents it from running in a react native environment. Since it is no longer maintained, it will never be fixed. That also means dependencies are outdated. [cross-fetch](https://github.com/lquixada/cross-fetch) is React Native compatible.


SachaG:
Oh awesome, thanks!

PR #1744, opened 2017-10-10
Bump the react-bootstrap version to get rid of "isMounted is deprecated..." error when opening modals

luhagel:
bump react-bootstrap -> 0.31.3


SachaG:
Thanks!

PR #1748, opened 2017-10-13
 Add callbacks to vulcan:routing SSR hooks

wayfarerboy:
Adding callbacks allows for access to the SSR app component and associated css in the routing's hook functions without needing to hack core files.

This makes it possible, for example, to add MUI wrapper components and inject compiled css with the JSS library (see https://material-ui-1dab0.firebaseapp.com/guides/server-rendering).

Here's how I'd implement the callbacks in my own custom module: https://gist.github.com/wayfarerboy/96c0f1d13b3b7d3d2743501a1c7c0187


SachaG:
Nice work! It'd be great to also add a section to [the docs](https://github.com/VulcanJS/vulcan-docs) explaining how to use them :)

PR #1750, opened 2017-10-17
Abstract out bootstrap-specific components in vulcan:forms

ErikDakoda:
Abstract out bootstrap-specific components in vulcan:forms and register them using registerComponent so that they can be overridden by other gui frameworks like material-ui.


SachaG:
Oh that's a great idea, thanks!

PR #1754, opened 2017-10-23
Correct prescript for its operation in windows, making it compatible on all three platforms: linux, MacOS and Windows.

codedeep:
Correct prescript for its operation in windows, making it compatible on all three platforms: linux, MacOS and Windows.
I add Vulkan's logo on the console.
logovulkan_in_console2

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # Vulcan (formerly Telescope)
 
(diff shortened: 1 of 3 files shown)


SachaG:
Oh that's awesome, nice job!

PR #1764, opened 2017-11-02
Implemented redux-idle-monitor integration

Discordius:
This is an integration based on https://github.com/noderaider/redux-idle-monitor and https://www.npmjs.com/package/react-redux-idle-monitor. 

Not sure about whether this is the API that makes most sense. Right now the integration comes with 3 default idle states "AWAY", "INACTIVE" and "EXPIRED", and the duration of inactivity until the client enters the respective idle state can be set in the settings.json file. I also added a bunch of callbacks that allow you to do things in the relevant idle state transitions, but I generally feel like this implementation is currently a bit halfway between fully-customizable and magic, and I am not super happy about that. 

Happy about recommended changes.

### DIFF
--- a/packages/vulcan-voting/lib/modules/vote.js
+++ b/packages/vulcan-voting/lib/modules/vote.js
@@ -93,8 +93,8 @@ const addVoteServer = ({ document, collection, voteType, user, voteId }) => {
   delete vote.__typename;
   Votes.insert(vote);
 
-  // update document score
-  collection.update({_id: document._id}, {$inc: {baseScore: vote.power }});
+  // update document score & set item as active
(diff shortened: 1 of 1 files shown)


SachaG:
If it works and doesn't add too many client-side dependencies I'm happy to accept it. 

Btw, it's better if you don't commit your `packages` and `versions` files. I guess there's an argument to be made for adding them to `.gitignore`…

SachaG:
Oh also could you submit a PR to the docs repo with an example of how this works (maybe in https://github.com/VulcanJS/vulcan-docs/blob/master/source/redux.md)? I'm not sure how the callbacks interact with the Redux stuff…

PR #1765, opened 2017-11-08
Added error for unregistered fragments

MHerszak:
I had a fragment not registered. Setting an error to make sure which one is missing is good for debugging.


MHerszak:
If you merge make sure to add Emails component in the debug package. I commented it out and did not reverse it. This will create issues when you use the debug package and rely on Emails component.

SachaG:
Thanks! But do you think you could submit a PR that only contains the commit relevant to this specific issue?

MHerszak:
How can I add individual pull requests? Every time I push a change and I want to add another pull request the only option is no option to add another pull request. Is there something I should consider before submitting the pull request?

SachaG:
Here's what I would do:

1) fork the Vulcan repo
2) make the changes you want
3) commit only the relevant files
4) push back to your github remote
5) submit PR

Issue #1767, opened 2017-11-09
[devel] Error Creating your own settings.json file...check files and permissions


martinhbramwell:
As discussed in Slack on Tuesday 2017/11/07
https://vulcanjs.slack.com/archives/C02L990AF/p1510080557000405

For ```devel``` branch users, it is impossible to run Vulcan with ```meteor npm start```.

It fails with the error "Error Creating your own settings.json file...check files and permissions'.

Seemingly, the intermittency is caused by there being two copies of ```prestart_vulcan.js``` one in the root directory and a second one somewhere below ```.meteor/local```.  The first is called by ```npm``` initially.  Subsequent calls come from within ```.meteor/local/build/main.js``` and try to copy a non-existant ```.meteor/local/build/programs/server/sample_settings.json```.

PR #1768, opened 2017-11-09
Bug (#1767): Set good work dir before copying sample_settings.json

martinhbramwell:
This adds a simple function to determine the real project source root directory before trying to copy sample_settings.json.


MHerszak:
@martinhbramwell my question is why is it necessary to copy the file in the first place. What is the use case?

martinhbramwell:
As far as I can see, the only purpose is to ensure a problem free start up experience for new users, as we all know and love about Meteor. From that viewpoint, the bug is much worse than the problem it's meant to solve. I did get it doing what it's meant to do, but I have seen that the sample settings cause the "newsletter" module to make Vulcan fail horrendously and  incomprehensibly. I've been trying to get to know Vulcan, but find I am constantly tripped up by surprise problems, hence my three PRs to date, and another on the way. I want to see Vulcan get really popular, but it won't with the current getting started experience.

SachaG:
What does `wrk_dir = '../../../../..';` do here? How are you determining this path?

martinhbramwell:
There are two possible locations from which the script can execute, which I really don't expect will ever change: the root of the project and the root of Meteor's run time of the project.  The latter is 5 levels below the former, so ```wrk_dir``` simply points back up to the top where the file is sure to be found.

SachaG:
Why would the script execute from two locations though? If that script doesn't behave in an understandable way, we might as well remove it.

martinhbramwell:
I believe it executes from within Meteor's runtime directories because Meteor is aggressive about collecting all possible changes when a developer updates any file not prefixed with '.'

Evidently it uses `npm prestart` as part of its changes collection process, which means that you can alter stuff like 'sample_settings.json' and Meteor will reload it each time!

I see no point in copying 'sample_settings.json' over and over again, so perhaps a better solution is to copy it when : 

1. there's no `settings.json` and
1. the base path of the script is **not** in Meteor's run time directories

I can redo it like that if you want.

martinhbramwell:
This PR would pull from the **wrong branch** in my fork.

Closed in favour of new PR  #1783

Issue #1669, opened 2017-07-11
TypeError: Cannot read property '_id' of undefined: Bug description and Fix


adalidda:
Hi,

I have 2 database (one in French and one in English). These 2 database work perfectly with VulcanJS version n-1. Now I am migrating my package for VulcanJS to VulcanJS 1.5.

I notice that my database in English work perfectly with my new package based on VulcanJS 1.5 while my database in French generate the following error:
W20170711-17:56:31.316(7)? (STDERR) 						throw(ex);
W20170711-17:56:31.317(7)? (STDERR) 						^
W20170711-17:56:31.318(7)? (STDERR) 
W20170711-17:56:31.319(7)? (STDERR) TypeError: Cannot read property '_id' of undefined
W20170711-17:56:31.321(7)? (STDERR)     at createComment (packages/vulcan:getting-started/lib/server/seed.js:63:13)
W20170711-17:56:31.322(7)? (STDERR)     at createDummyComments (packages/vulcan:getting-started/lib/server/seed.js:119:3)
W20170711-17:56:31.323(7)? (STDERR)     at packages/vulcan:getting-started/lib/server/seed.js:147:5
W20170711-17:56:31.324(7)? (STDERR)     at Function.time (/home/kauprey/Production-V4.0/.meteor/local/build/

So I identify that the package vulcan:getting-started generated this bug !

By removing this package with the command meteor remove vulcan:getting-started, my new package based on VulcanJS 1.5 work perfectly with my 2 database.

Hope this description help to fix the bug in the package vulcan:getting-started

Thank You
Adalidda

kamish24six:
Thanks, but I get the response: Vulcan: getting-started is not directly dependent on this project And when I upload the app to the herooku I get an error

martinhbramwell:
This issue needs higher priority, imho.  It is very **frustrating for newcomers** to Vulcan.

There is a race condition where dummy users are created asynchronously and then called upon for creating other records, *before their instantiation has to time to complete*.

This little hack solves the problem ...

    if (!Users.find().count()) {
      createDummyUsers();
    }

    while (  Users.find().count() < 3  ) {
      Meteor._sleepForMs(500);
    };

    if (!Pics.find().count()) {
      createDummyComments();
    }
... , but a proper asynch solution would be better, and instructive for newbs.

eric-burel:
Seems to have been fixed a while ago, closing for now

PR #1769, opened 2017-11-09
Bug (#1669): Wait for users to exist before referring to them

martinhbramwell:
Fixes issue #1669.
Adds a simple wait loop in each example module to ensure initial start up proceeds without breaking.

### DIFF
--- a/packages/example-forum/lib/server/seed/seed_posts.js
+++ b/packages/example-forum/lib/server/seed/seed_posts.js
@@ -145,6 +145,11 @@ if (getSetting('forum.seedOnStart')) {
     if (!Users.find().count()) {
       createDummyUsers();
     }
+
+    while (  Users.find().count() < 3 ) {
+      Meteor._sleepForMs(500);
+    };
+
     if (!Posts.find().count()) {
       createDummyPosts();
     }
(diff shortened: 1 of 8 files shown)


SachaG:
I understand the issue but this doesn't seem like a good fix… Isn't there a cleaner way?

SachaG:
Also I would've thought the fact that Meteor uses Fibers would take care of race conditions like this? Isn't it supposed to ensure everything runs in the right order?

justinr1234:
The proper solution would be in this file: https://github.com/VulcanJS/Vulcan/blob/master/packages/example-forum/lib/server/seed/seed_posts.js

```
import { Promise } from 'meteor/promise';

// ...

  const createUser = function (username, email) {
    const user = {
      username,
      email,
      isDummy: true
    };
    return Promise.await(newMutation({
      collection: Users, 
      document: user,
      validate: false
    }));
  }

```

This would cause any call to `createUser` to wait in the Meteor Fibers as @SachaG has mentioned.

justinr1234:
Fixed here: https://github.com/VulcanJS/Vulcan/pull/1849

Issue #1770, opened 2017-11-10
[devel] [blocking]  'example-permissions' admin privileges appear not to work


martinhbramwell:
_Browser console error_ :- 

    Warning: React.createElement: type is invalid -- 
    expected a string (for built-in components) or a class/function (for composite components) 
    but got: object.
    You likely forgot to export your component from the file it's defined in.
    Check the render method of `NavLoggedIn`."

Cause:- In ```packages/permissions-example/lib/components/common/Header.jsx``` the ```ModalTrigger``` finds 'object' not 'function' as React element.

PR #1771, opened 2017-11-10
Fix issue #1770

martinhbramwell:
PicsNewForm must be called as a registered component, not as an imported object.

This PR addresses issue #1770.

Three different tutorial example packages are affected.


SachaG:
Thanks!

PR #1772, opened 2017-11-10
Changes to support vulcan-material-ui

ErikDakoda:
- Pass all properties of fieldSchema.form on to the form component
- Restore missing delete link


SachaG:
Awesome as always!

PR #1778, opened 2017-11-22
Add pre-validate callback on new user creation

mitkodkn:
I had the necessity to modify the options which are passed to the user schema validation function. That's why I created a synchronous callback for it.

I'm concerned about the name of the hook. Please, give your advice and opinions.

### DIFF
--- a/packages/vulcan-users/lib/server/on_create_user.js
+++ b/packages/vulcan-users/lib/server/on_create_user.js
@@ -8,6 +8,8 @@ function onCreateUserCallback (options, user) {
   delete options.password; // we don't need to store the password digest
   delete options.username; // username is already in user object
 
+  options = runCallbacks(`users.new.pre_validate`, options);
+
   // validate options since they can't be trusted
   Users.simpleSchema().validate(options);
 


MHerszak:
Interesting. Callbacks concern aspects which are usually `before`, `after`, `error`, it happens that there is only one hook to do it all and VulcanJS framework doesn't apply mongo helpers to add aspects. The callback doesn't have to be called if I don't add it, but why would it be an advantage to add a callback with the word pre into a hook which isn't a `before` aspect. As far as I am concerned, a before hook would help with stopping the entry in some case but this callback won't. I am not opposed but what could be added to this hook? What is your use case? Why can't it be done in the available hook?

mitkodkn:
In my case, I've added a **firstName** and **lastName** fields to the User schema which I would like to be required. Everything went perfectly fine until I tried to implement Facebook and Twitter authentication.

When a user allows the developer application to use their data, an additional object is passed to the options parameter in the callback, passed to the **onCreateUser** hook in the **vulcan-accounts** package. However, I need to somehow modify the options object to match the User schema before it gets validated.

```javascript
function onCreateUserCallback (options, user) {

  const schema = Users.simpleSchema()._schema;

  // ...

  // Here comes the problem. I want to modify the options
  // parameter to make the firstName and lastName properties
  // appear in the user object as they are defined in the
  // schema. If they are not, the validation will fail.

  // validate options since they can't be trusted
  Users.simpleSchema().validate(options);
 
  // ...

  return user;
}
```
If there's a better way to do the things in my case, I would really appreciate it if I get an advice.

MHerszak:
You mean you want to sanitize the user object when you create a user with either fb or tw?

mitkodkn:
Yes, before the schema validation.

MHerszak:
I misunderstood the hook. I think it is a good idea. Which means you have more control over oAuthentication. The legacy branch has this implemented but there seems to be no schema validation.

SachaG:
Usually in that case I'd just make `firstName` and `lastName` not required, but this might be a better way. How about `users.new.validate.before for the name?

I wonder if we need the same hook for default mutators, because we often run into the same issue there. For example `createdAt` should be required on all posts, but because it's not yet inserted at the validation step, it has to be marked as `optional`. 

Another option would be to validate *after* default values are inserted. I guess it depends if you see validation as just a way to clean up client data, or as a way to enforce strict rules no matter what…

PR #1784, opened 2017-12-13
Adhoc sql route example, 'example-adhoc-route'

martinhbramwell:
Hi Sacha

So here's my sequelize adaptation for you.

I've tried to set it up so that you have only a small number of well defined changes to review, basically in 3 files : 
 - `.meteor/packages`
 - `package.json`
 - `packages/vulcan-lib/lib/modules/collections.js`

A 4th file `prestart_vulcan.js` is a simpler fix for issue #1767, and not really anything to do with the central issue.

All but 2 of the new files reside in the new directory `example-adhoc-route`.

The 2 that do not `packages/vulcan-lib/lib/modules/mongoCollection.js` and  `packages/vulcan-lib/lib/modules/mongoCollection.js` are the collection adapters for Mongo & Sequelize respectively.


martinhbramwell:
Sorry for the delayed reply. I was travelling.

I will look into a new PR minimising the impact on Vulcan soon.  Yes, I'll maintain my own version, that's understandable. Would you at least consider a PR on the docs that references my example?

The name is as it is, because when I started writing the ad hoc route example, I really did not expect that there was no SQL capability for Vulcan's GraphQL implementation.  I will change the name. I agree it is misleading.

SachaG:
Oh sure a PR to the docs would be great. Maybe a "Other Databases" or "Other Datasources" section in the Data Layer section?

PR #1788, opened 2017-12-20
Fix replication of existing image when user click on remove link if e…

seerahulsingh:
…nabled multiple

### DIFF
--- a/packages/vulcan-forms-upload/lib/Upload.jsx
+++ b/packages/vulcan-forms-upload/lib/Upload.jsx
@@ -169,7 +169,7 @@ class Upload extends PureComponent {
     const newValue = this.enableMultiple() ? removeNthItem(this.state.value, index): '';
     this.context.addToAutofilledValues({[this.props.name]: newValue});
     this.setState({
-      preview: newValue,
+      preview: '',
       value: newValue,
     });
   }


seerahulsingh:
@SachaG Probably a small change, but there a bug when user clicks on remove link of an image, all existing images replicates if enabled multiple.

If you think changes is made on the right place, please merge it.

SachaG:
Oh, thanks for fixing that! I think maybe using `preview: null` would be a little cleaner? But yeah that should work.

SachaG:
Oh also please make future PRs to the `devel` branch. I'll add in the change myself for now :)

seerahulsingh:
OK, so this is fixed in `devel` branch, when you are planning for next release?

SachaG:
Soon, hopefully this week-end :)

seerahulsingh:
Ok thanks.

One more thing, To fix the issue temporarily on my project, I made this changes on my local system in Vulcan core, But when I deployed my project it didn't take the changes on production. 

So when meteor does the bundling, from where it looks for Vulcan core. Can I point it to my local one?

SachaG:
Are you using the 2-repo install? If so git-based deploys like Heroku won't take your core modifications into account since your Vulcan core files will not be in the repo. You can use Meteor Up or Galaxy instead; or else use a git submodule for your core files.

seerahulsingh:
Yes I'm using 2 repo install. And deploying my app on digital ocean using pm2-meteor. 

Can't seem find where to make changes so that my local change in Vulcan core takes affect.

SachaG:
Could you use Meteor Up? If not you won't be able to use the 2-repo install as such, but you can either do something similar using git submodules; or else just use a single repo. It won't change how your app works at all, it will just make updating/contributing a bit harder…

PR #1789, opened 2017-12-22
Delete Posts and Comments when a user is deleted

deligence:
Set options to delete Posts and Comments when a user gets deleted. Currently it creates an issue when a user is deleted. No post gets loaded on home page when a user who has some posts gets deleted.


MHerszak:
It is not a good idea to be so specific in lib. The way this should work is to add a hook on user deletion. Delete all references (based on userId). You will have to do that for the comments package and the post package. This pr will only create issues for others using Vulcan.

SachaG:
@MHerszak is right, we can't really do it this way, sorry. I'll add more details in the issue thread.

PR #1790, opened 2017-12-27
Cleanup 1.8.1

luhagel:
Couple of minor things I found while upgrading if I find more, I'll append them here.
* vs code typings aren't being ignored in the .gitignore, now that Vulcan ships with a .vscode folder, those shouldn't be tracked
* the data table EditButton had two extensions
* small typo in the prestart script
* since it's not placed in a .whatever directory, meteor loads the prestart script which causes all sorts of fuzz, created a .vulcan directory and moved it there


SachaG:
Nice! Actually I'm not sure if we should even keep that prestart script. Probably best to fold it in the Vulcan CLI (https://github.com/VulcanJS/vulcanjs-cli)

PR #1795, opened 2018-01-02
fix duplicate email returns "unknown error' 

schuttsm:
Bumped up against this trying to reproduce another bug.

Repro steps:
1) create a user & account & logout
2) create a 2nd user account with the same email but a different login.  Notice that the error is "unknown error" rather than "email already exists"

Turns out it was just a missing i18n string.  Now it says "email already exists".  much more helpful.


SachaG:
That's great, thanks!

PR #1797, opened 2018-01-02
deleted user post bug

schuttsm:
Addresses: https://github.com/VulcanJS/Vulcan/issues/1776

* fix bug where post from deleted user causes this example to break.  Now it just hides those posts


justinr1234:
@schuttsm This code is already done inside `Users.isAdmin`: https://github.com/VulcanJS/Vulcan/blob/master/packages/vulcan-users/lib/permissions.js#L174

cc: @SachaG looks like this can be closed

PR #1799, opened 2018-01-06
Upgrade to React 16.2

luhagel:
Updated a whole bunch of dependencies and removed all legacy PropTypes/React.createClass implementations to make updating to React 16 possible.

Stable from what I could test so far, probably better to give it a test run before merging.


SachaG:
That's awesome, thanks so much! I'm not sure about `FormWrapper` and `shouldComponentUpdate` btw, at one point it was updating too much so I had to hack this in but it might be fixed now. 

Anyway I'll merge this to `devel` since it's something we need to do sooner or later anyway :)

PR #1801, opened 2018-01-06
Add VSCode jsconfig

luhagel:
Once again, now that Vulcan ships with the .vscode folder included, we should take full advantage of it.
The jsconfig adds automatic typing acquisition and enables deeper intellisense.


SachaG:
Thanks! I'm still learning my way around VS Code, but I definitely want to take better advantage of its features like adding autocompletes, autoimports, etc.

PR #1802, opened 2018-01-06
Debug Groups & Colors

luhagel:
A few changes to support debugGroup() and debuggGroupEnd() in addition to debug().
Also makes use of those in the callback & resolver logs, in addition to some colors.

screen shot 2018-01-06 at 1 47 02 pm


SachaG:
Oh that's a great idea, thanks!

PR #1803, opened 2018-01-06
fix admin delete user hide bug

schuttsm:
Fix a simple bug in vulcan:admin user list page after a user delete

Repro:
1) go to /admin
2) edit another user, scroll down & click delete, click "ok" to confirm delete

notice the modal stays displayed after the user is successfully deleted


SachaG:
Thanks!

PR #1808, opened 2018-01-06
ReactDOM.hydrate() may not be defined for everyone

MHerszak:
ReactDOM.hydrate() may not be defined for everyone, it will now use hydrate if it is defined or fallback to render

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "express": "^4.14.0",
     "flat": "^4.0.0",
     "formsy-react": "^0.19.5",
-    "formsy-react-components": "^0.10.1",
+    "formsy-react-components": "^0.11.1",
     "graphql": "^0.9.6",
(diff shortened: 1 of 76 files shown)


SachaG:
Why would it not be defined for everyone?

MHerszak:
I am running React, and ReactDom on version 15. and not 16. Hydrate seems to be a feature of 16.x

SachaG:
But we're also upgrading to React 16 anyway?

MHerszak:
Yeah. I just thought I don't want to force others to update as well. This can be taken out after a while.

SachaG:
But don't we *want* everybody to be on the same version of React? Handling multiple versions at the same time seems like a recipe for issues…

MHerszak:
Okay, will close this pull request.

PR #1811, opened 2018-01-11
Form loading state fix

Sebi55:
In some cases it seems that the loading property is set to true altough the document is already there.


SachaG:
:+1:

PR #1813, opened 2018-01-11
Use getComponent for childRoutes

Sebi55:


SachaG:
I don't use child routes much myself, but that sounds like a good idea. Thanks!

Sebi55:
I used it often when I want to share the same layout for different routes.

I just watched the hangout. This would maybe be needed to change when switching to react router v4.

PR #1814, opened 2018-01-12
Fix Newsletter SSR

luhagel:
Alright, really weird one related to React16 SSR changes.
https://github.com/facebook/react/issues/10591
Basically, React changed it's diff algorithm and now reuses components differently.
So, when the check for showing the newsletter was inside the constructor, instead of generating something along the line of
```html
<div className="newsletter">
{ ... rest of Newsletter code }
</div>
<div>
{ ...content }
</div>
```
the actual result was
```html
<div className="newsletter">
  <div>
  { ...content }
  </div>
</div>
```

Setting showBanner inside `componentDidMount` fixes that.


luhagel:
closed, will make a cleaner one

PR #1816, opened 2018-01-12
Set default waiting state to false to avoid SSR override

luhagel:
Currently, the waiting state get's overwritten by the server version, leaving signIn buttons permanently disabled.


eric-burel:
Hi,
Almost got it working, but how do can I confirm that I am loading the correct meteor package dir ? It seems that your PR isn't taken into account in my app, so either it does not work or I don't apply it. Are you sure setting this state to false is enough ? Can't the tracker run once and make waiting false again ?

SachaG:
I haven't had this issue myself but it seems like a reasonable fix. I'll merge it.

PR #1821, opened 2018-01-18
Apollov2 Integration start

MHerszak:
This is only the move from v1 client to v2 client.


MHerszak:
I need to get all other commits straight, first.

PR #1822, opened 2018-01-18
Apollov2 start

MHerszak:
This is the first commit to Apollo client v2.


SachaG:
Hmm, still not a fan of 

```
// ReactDOM.hydrate() may not be defined for everyone, it will now use
 +        // hydrate if it is defined or fallback to render
```

Also, do you think you could make a PR that only includes the relevant commits when you have time?

MHerszak:
Yeah, I can do that again. I should have done the changes after my merge.

PR #1828, opened 2018-01-25
fix allVotes resolveAs

farjang:
I couldn't get allVotes, after talking to habryka he said this could be a bug. This code works for me, please take a look.


MHerszak:
You are correct. The schema has `documentId` and not `itemId`. However, could you please open another request to `VulcanJS:devel`. Thank you!

PR #1832, opened 2018-01-25
Vulcan updates jan 2018

justinr1234:


SachaG:
Wow, this does a lot of stuff… I really appreciate all the work you put in but I can't really merge this, there's just too much in there! It'd be great if you can submit smaller, separate PRs for individual changes. 

Or otherwise, I'll try to find some time to look through all these changes and manually apply them myself.

justinr1234:
@SachaG Ah, was just following what you said about submitting 1 PR with all my changes when we spoke in private.

SachaG:
Oh you're right, sorry. I just didn't know there were that many! No worries, I'll look through them over the next couple days :)

MHerszak:
@justinr1234 what is `AccountsEnrollAccount`. What is the use case for this?

justinr1234:
Right now if you register a user on the server and send them an enrollment link token email, when they land back on the page they hit the route `/#/enroll-account/:token`. This is handled by default by the Accounts package and calls the callback to do enrollment. Vulcan is setup to intercept this but the problem is that the account enrollment form is generally in a Dropdown. Therefore we are dropped on the home page with no form to set the user password at all. This new component mimics the password reset flow, except it is for a brand new user created without a password by the server.

justinr1234:
As a side conversation, I wanted to discuss refactoring the login form to not use session variables for intercepting these values from the accounts system.

justinr1234:
re: package-lock - I updated that on purpose as I upgraded the framework to Meteor 1.6.1.

MHerszak:
I just went through your changes. I like the `withList.js` update. It doesn't require you to import collections anymore. Maybe consider using the debug package for all your console.logs? I'd be happy to merge if @SachaG is ok with it. There are a lot of changes due to linting I presume?

justinr1234:
The debug package may not be enabled so I think keeping console.log is good also some of it isn’t debug. Some of it is runtime error logging which should be there.

If you take a look at the individual commit diffs you can see the real changes. The last commit is the lint cleanup that has the absurd change log!

Re: linting - I think we should implement a free hookup of circle ci that prevents merging PRs until they have passed linting (and eventually testing). Thoughts?

MHerszak:
There are a couple of console logs which are on the master and I am not really happy with it. Maybe we could use the npm debug package?

Style linting is a good idea! And so is debug and so is commenting changes ;)

justinr1234:
The npm debug package can get you into some weird situations. I think it's best to just use `console.log` directly as there's some stuff that happens with the native function in terms of output that is better than using the npm debug package.

MHerszak:
What are those weird situations? That would be good to know.

justinr1234:
Last time I used it there were issues with it outputting objects that were inspectable. It kept trying to log them as strings instead of like what happens in the chrome dev console where you can click the object tree and example it.

SachaG:
Let's hold off on merging this, I'd rather take the time to look at each change. Also let's open new issues to discuss related issues like the debug package or login form.

SachaG:
Btw why did you remove `packages/_boilerplate-generator`?

justinr1234:
It was incompatible with the upgraded meteor. Why was it in the repo?

SachaG:
To fix an issue with Meteor CSS. 

I guess I'll do my best to make this PR work somehow but in the future we can't really work like this… i.e. updating Meteor should be a separate PR, and when it's unclear what something does deleting it outright is not always the right answer :)

justinr1234:
They upgraded the package so I was pretty sure it was good to go. Either way, I’ll just do single PRs going forward.

SachaG:
OK, so I did my best to merge everything in there that made sense to me. I grouped the stuff I left out in this other branch's commit: 

https://github.com/VulcanJS/Vulcan/commit/c989a9a3fd220682542b715595de08eea563af3b

Leaving aside the Meteor update stuff, which I'll do separately, the main thing is the changes to FormComponent. Can you explain those?

SachaG:
Oh also FYI I'm going to move the Forum code either to the starter repo, or to its own repo altogether, having it in core makes everything harder to manage I feel.

MHerszak:
Just tested the devel branch and git the following error:
`ReferenceError: registerSetting is not defined at utils.js (packages/vulcan:lib/lib/modules/utils.js:17:1)`

I am going to add it back to utils.

justinr1234:
The formwrapper changes are to pass the correct prop values down to the components.

SachaG:
Do you think you could make a separate PR just for the FormWrapper stuff then? And maybe add some more code comments to make what you're doing more explicit?

justinr1234:
@SachaG https://github.com/VulcanJS/Vulcan/pull/1847

PR #1835, opened 2018-01-26
upgrade vulcan-subscribe to use example-forum for posts & categories

klokie:
Hi @SachaG, I noticed that the refactoring that consolidated `vulcan:posts` and `vulcan:categories` into `example-forum` in 6a9f53e and 62dffa6a9d91b0d1351062ae8ccaf68692dd3748 didn't update the `vulcan-subscribe` module to work with it. I hope this patch should fix it.


MHerszak:
Your "only" changes are within `callbacks.js` within the first 11 lines the rest is linting, correct? Why do you want to create a dependency for the subscriber package with the example-forum package?

Discordius:
Yeah, the subscriber package shouldn't have to depend on example-forum, though right now about half of its functionally is for that specific use-case. 

For the linting: It seems that for this PR you probably just want to drop your first commit in this PR, which made all the style changes. You can then add those back in another PR, but seems better for them to be separate.

SachaG:
Also please make the PR to `devel` :)

justinr1234:
@SachaG why is a core package referencing `example-forum`?

SachaG:
This is leftover code from before, I don't use that package myself and I didn't create it so I haven't really been maintaining it…

PR #1837, opened 2018-01-26
Fix circular dependencies between schemas and collections

justinr1234:
`collection.js` is importing `schema.js` for `createCollection` but `schema.js` is importing `collection.js` for varying runtime function definitions resulting in undefined circular load order on files.

### DIFF
--- a/packages/example-forum/lib/modules/categories/schema.js
+++ b/packages/example-forum/lib/modules/categories/schema.js
@@ -5,7 +5,6 @@ Categories schema
 */
 
 import { Utils } from 'meteor/vulcan:core';
-import { Categories } from './collection.js';
 
 export function getCategoriesAsOptions (categories) {
   // give the form component (here: checkboxgroup) exploitable data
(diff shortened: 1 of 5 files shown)


MHerszak:
Nice! While you are at it. How about a little warning in case you cant find a collection? Find returns 'undefined' and will break it 💯

justinr1234:
@MHerszak I would assert it should break in that case. Thoughts?

MHerszak:
you could just add console.error ;)

justinr1234:
Leaving it would allow the stack trace to print out properly showing where it failed and triggering someone to realize they misspelled it. I'm still asserting not to catch the error!!!!!!

MHerszak:
This might be good to be included as props like `collectionName` for SmartForms as well.

justinr1234:
@MHerszak https://github.com/VulcanJS/Vulcan/pull/1840

SachaG:
Thanks!

PR #1839, opened 2018-01-26
Revert "showEdit on Card"

MHerszak:
Same here made the mistake to not look at the branch.

### DIFF
--- a/packages/vulcan-core/lib/modules/components/Card.jsx
+++ b/packages/vulcan-core/lib/modules/components/Card.jsx
@@ -119,10 +119,10 @@ const CardEditForm = ({ collection, document, closeModal }) =>
     }}
   />
 
-const Card = ({className, collection, document, currentUser, fields, showEdit = true}, {intl}) => {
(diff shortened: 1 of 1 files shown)


SachaG:
Maybe I don't understand how Git/GitHub works, but shouldn't the reverting commit also be made to `master`?

It looks like `master` still has all 3 commits: https://github.com/VulcanJS/Vulcan/commits/master

While `devel` is now a mess of commits and reverses…

https://github.com/VulcanJS/Vulcan/commits/devel

In this situation I think it's better for you to either leave the commits on `master` and ensure they're made to `devel` from now on; or if you're not sure just wait and ask me.

MHerszak:
@SachaG Yeah, I thought the same thing after I made the revert. Sorry about the mess. Can we add a template to the pull-request? Just a checkbox with pull requests to devel. The contribution info does say it put it is not immediately visible when you do the pull request.

SachaG:
Yes that's a good idea. Can you add that suggestion here: https://github.com/VulcanJS/Vulcan/issues/1842 ?

PR #1840, opened 2018-01-26
Allow SmartForm to use collection or collectionName

justinr1234:


MHerszak:
You are a machine!

SachaG:
Oh, I didn't even know you could specify a function as a propType. 

How much do we really want to do this though? I see how it can be helpful, but then what's the rationale for doing it for this specific prop and not the others?

I just wonder if we're not overthinking things and maybe we should just be throwing an error inside the component's code rather than in a "non-obvious" (to me at least) place like that.

justinr1234:
@SachaG The idea to avoid having to import a collection at file load time. This creates circular dependencies between files that aren't necessary. The collection needs to be accessed at runtime, but when you do `import { collection } from './file.js'` you're creating a circular dependency in some cases. There's no particular need to have a reference to the collection at import time, but there are multiple cases where having it imported is actually wrong. (especially after you continue to upgrade to newer versions of Meteor, node, and babel, which have stricter checking for these things).

SachaG:
Oh yeah I totally get the idea for the PR itself, it's a great idea. I was just thinking out loud about that `propType` thing specifically.

justinr1234:
Oh, the reason for the proptype function was to make it so it's an either-or requirement on the props. https://stackoverflow.com/a/42284491

SachaG:
OK, makes sense. I don't know if we want to systematically do this (there are a lot of other places with the same issue of accepting two props) but we can leave it here ofr now.

PR #1844, opened 2018-01-27
Consolidate getCollection code

justinr1234:


MHerszak:
Just went through the same pain.

PR #1845, opened 2018-01-27
Fix withDocument

MHerszak:


SachaG:
That doesn't seem right?

SachaG:
That doesn't seem right?

MHerszak:
The collection wasn't defined when it reached line 10 'undefined'. How do you mean?

MHerszak:
@SachaG I just check 'Consolidate getCollection code' same fix for withList.

SachaG:
OK, but I meant including 4 commits in the PR? I'll fix it myself.

PR #1847, opened 2018-01-27
Fix React prop warnings for SmartForm

justinr1234:
Most of this has to do with default values: https://github.com/VulcanJS/Vulcan/blob/devel/packages/vulcan-forms/lib/components/FormComponent.jsx#L51

**Checkbox**

- Expects value to be a boolean
- https://github.com/twisty/formsy-react-components/blob/v0.11.1/src/checkbox.js#L21

**CheckboxGroup**

- Expects `value` to be an array
- `formsy-react-components` tries to get `props.value` as an array: https://github.com/twisty/formsy-react-components/blob/v0.11.1/src/checkbox-group.js#L42 which hits the mixin of Formsy: https://github.com/christianalfoni/formsy-react/blob/v0.19.0/src/Mixin.js#L36

**select**
- If the select is of type `multiple`, then the value must be an array
- If the select is NOT `multiple`, then the value must be a scalar
- Therefore, the default value should not be `[]` when `multiple` is not specified. It should be `''`.

### DIFF
--- a/packages/vulcan-forms/lib/components/FormComponent.jsx
+++ b/packages/vulcan-forms/lib/components/FormComponent.jsx
@@ -67,7 +67,7 @@ class FormComponent extends PureComponent {
     } else if (typeof this.props.control === 'string') { // else pick a predefined component
 
       switch (this.props.control) {
-        
+
         case 'number':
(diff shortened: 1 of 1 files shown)


SachaG:
Thanks!

PR #1848, opened 2018-01-27
CircleCI Config

luhagel:
Alright, this should enable us to run CircleCI with every build/pr/whatnot.
Right now, it's only running the linter, more to come.

_Already using a container with jdk in order to enable webdrivers in the future._

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,94 @@
+# Javascript Node CircleCI 2.0 configuration file
+#
+# Check https://circleci.com/docs/2.0/language-javascript/ for more details
+#
+version: 2
+jobs:
+  build:
+    docker:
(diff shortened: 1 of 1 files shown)


luhagel:
While we're at it, I'm not allowed to enable CircleCI for Vulcan, so that'd need @SachaG 's input

justinr1234:
@luhagel Thanks for this. Can you add comments to the circle CI file to explain what the different lines are doing? It's a bit confusing for Circle CI newbies like myself.

SachaG:
> While we're at it, I'm not allowed to enable CircleCI for Vulcan, so that'd need @SachaG 's input

How do I do that?

luhagel:
@justinr1234 sure thing
@SachaG Go to circleci.com, log in with github, select the vulcanjs org on the top left and add Vulcan as a project ;)

PR #1849, opened 2018-01-27
Fix seeding not being sequential & fix broken example packages

justinr1234:
- Use `Promise.await` during `Meteor.startup` to simulate `await` inside a Meteor Fiber
- Switch all places where we're calling the `async` / `Promise`-based `newMutation` during seeding to also be an async function that returns the promise returned by calling `newMutation`
- Inside all the new `async` functions, swap to using the `Promise`-based Mongo apis accessed via `rawCollection` as described in this article: https://blog.meteor.com/using-promises-and-async-await-in-meteor-8f6f4a04f998
- Cleanup some linting while inside the files being changed
- Convert to ES6 syntax while inside files being changed
- Parallelize seeding creations of individual items using `Promise.all`
- Hide `Upload` link for `example-instagram` and `example-membership` based on appropriate criteria
- Pull cloudinary preset from Meteor settings for `example-instagram` and `example-membership`
- Add getting started content removal for all example packages
- Refactor getting started removal content for `example-forum` to properly move across 2 files for `seed_posts.js` and `seed_categories.js`. Previously `seed_posts.js` was removing seeded `Categories` content which is not correct. Now, the removal function inside `seed_categories` will call the `seed_posts` function in a Higher Order Component style.
- Update `example-membership` to use `<Components.PicsList />` instead of `<PicsList />`
- Fix `example-membership` to reference proper URL for header logo (it was previously using `example-forum`
- Fix `vulcan:forms-upload` to properly respect an array of images and not try to render an empty array when in `multiple` mode


justinr1234:
Properly addresses: https://github.com/VulcanJS/Vulcan/pull/1769/files

MHerszak:
I'd say pull but as of yesterday's mistake with one pull request to master I might just wait it out.

justinr1234:
fwiw I tested each example package separately and cleared by DB between runs.

SachaG:
Nice work. Btw. from now on please make PRs for the example packages to the Starter repo (https://github.com/VulcanJS/Vulcan-Starter). I want to remove them from the core repo. 

Also seeing this makes me think we should have some kind of `vulcan:seed` package to avoid repeating the same code all the time?

justinr1234:
I was thinking the same thing about a seed package.

PR #1850, opened 2018-01-28
Added getCollection to withNew

MHerszak:


SachaG:
Thanks. I guess let's do the same for `withEdit` and `withRemove` too then?

MHerszak:
Yes, just wanted to make sure to not make mistakes this time :)

PR #1853, opened 2018-01-28
Respect checkAccess for total resolver

justinr1234:


SachaG:
Btw I wonder if it would make more sense to get rid of the Total resolver and have the list resolver return an object with both `results` and `total` fields? Maybe at some point we should take another look at that part of the API.

justinr1234:
Maybe, but in that case you can just do `results.length`

SachaG:
Well no because `results.length` gives you the number of results currently loaded, while `total` gives you the *total* count for matching results in your db (i.e. could be a hundred thousand).

justinr1234:
Good call, that could make the most sense.

PR #1855, opened 2018-01-28
optionsAsStrings was not defined

MHerszak:
I just implemented @justinr1234 fix for the select component.


justinr1234:
Looks good

PR #1861, opened 2018-01-29
packages update to meteor 1.6.1

MHerszak:
updated all vulcan packages to meteor 1.6.1. Tested as well and it works locally.


SachaG:
Thanks!

PR #1862, opened 2018-01-29
Omit formState in AccountsSwitcher

MHerszak:
this.props and state both pushed formState.


MHerszak:
The conflict is only a one-liner and can be deleted, which will resolve it.

SachaG:
Wouldn't the formState from props just get overridden by the one from state? I don't know if the omit is necessary?

MHerszak:
The later once are being overwritten is what the specifications say. I just don't believe it is good practice but it doesn't hurt either.

SachaG:
Given that I want to scrap this whole accounts package anyway I think we can leave it for now.

PR #1868, opened 2018-01-31
Datatable - replace SPACE with - in the datatable-item-* className

vincro:
Currently column names such as "Company Name" for example that do not exist in schema get converted to:

`datatable-item-company name`

rather than being converted to:

`datatable-item-company-name`


SachaG:
Nice. But I might be wrong, but doesn't `replace` only replace the first instance of a string? So there would still be cases where it doesn't work?

justinr1234:
Yes it only does the first letter. Need a regex replace for multiple.

vincro:
Indeed you are correct.  I will fix and push change.

SachaG:
Nice work :)

Issue #1865, opened 2018-01-29
Allow upsert mutation to accept terms & parameters in the search


justinr1234:
`search` uses `${collectionName}Input` type in the mutation params. It should just use plain `JSON` and respect `selector` and `options` by calling `let { selector, options } = await collection.getParameters(terms, {}, context);`. Therefore in the `upsert` default mutation, `search` would be equivalent to `terms` in the default `list` resolver.

SachaG:
What `search` are we talking about exactly?

justinr1234:
The first parameter of the upsert was the `search` which is just the key/value pairs to match on in collection.find(). I realized later that this is not keeping in line with the rest of Vulcan and should instead follow the same `terms` approach as a query.

SachaG:
Oh I didn't notice that. I'm not sure what it should be actually, `terms` are specifically used for list resolvers. 

Looking at your code again though, I think the lack of `Utils.performCheck` opens a security hole because we never check if the current user can update the current document. So you could potentially use this `upsert` mutation to modify an existing document and gain access to it, it seems? Am I missing something?

I guess I also didn't really stop to ask why we needed an upsert mutation in the first place. I personally never had to use one before, so unless there's a strong reason to add it maybe the extra complexity involved means it'd be better to leave it out of Vulcan core?

justinr1234:
It follows security checks because it doesn’t do any actual work until it calls the existing mutations. All it does is see if a document exists with the given params. If it does, it calls the edit mutation. If it doesn’t exist, it calls a new mutation. Those mutations handle the security check.

Upsert is useful if you want to search and update something not using `_id` or `slug` and then insert it into the database otherwise. It is absolutely something that should be in core. Same argument as why Meteor and Mongo provide the operation.

SachaG:
Oh gotcha, I got confused and thought you were calling `newMutation` and `editMutation`. We really need to stop calling those "mutations" and use the word "mutator" instead to distinguish them from the default mutations. 

As for the `search` issue I'll defer to you as I don't really have a concrete use case to guide me. It does feel like it would be different from `terms` here though. On the other hand `search` doesn't feel appropriate, maybe we just want to call it `selector` then? 

Unlike in the `list` scenario we're not fetching any data so there is no security risk in letting the client directly specify a selector (I think?).

SachaG:
[...]

PR #1869, opened 2018-02-01
#1865 Fix upsert

justinr1234:
Allows to search by any valid mongo search

### DIFF
--- a/packages/vulcan-core/lib/modules/containers/withUpsert.js
+++ b/packages/vulcan-core/lib/modules/containers/withUpsert.js
@@ -4,7 +4,7 @@
 
  Sample mutation:
 
-  mutation moviesUpsert($search: MoviesInput, $set: MoviesInput, $unset: MoviesUnset) {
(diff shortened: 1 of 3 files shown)


SachaG:
Cool. I wonder if eventually we want to do something like Primsa and specify a `MoviesWhere` type?

https://gist.github.com/gc-codesnippets/fed6ffe8c34a06b1682cc227dfcfd549#file-schema-graphql-L132-L361

It would be more work and less flexible than a JSON object, but probably safer from a security/validating point of view.

justinr1234:
I think it would be more consistent if we followed the query `terms` & `parameters` approach. I was thinking we could literally just make upsert accept a combination of `editMutation` and `list` query but limit the database call to `findOne` instead of liist’s `find`.

Basically we could make a `single` query that behaves like `list` but on a `findOne`. Then if that returns successfully, we call `editMutation` and if it doesn’t match anything we call `newMutation`.

PR #1872, opened 2018-02-01
Fix embed for posts callback

justinr1234:


MHerszak:
Nice. What is the difference between `import { HTTP } from 'meteor/http'` and `Meteor.http`? I have not looked into this, yet.

justinr1234:
I think they moved it to its own package at some point. `Meteor.http` doesn't exist.

MHerszak:
Good call!

PR #1874, opened 2018-02-01
Fix example-forum events callback

justinr1234:


SachaG:
Why would that trigger an error? If the event doesn't exist wouldn't `existingClickEvent` just be `undefined`?

justinr1234:
I think the code is wrong altogether. `vulcan-events` doesn't export a collection.

SachaG:
Ohh right I did change that. Yeah this would need a refactoring to use `track`. And maybe we then add a `unique: true` option to `track` to make it check for existing events first.

justinr1234:
For the time being this fixes it, but it isn't the right fix long term. I don't have time to mess with fixing it right at the moment!

SachaG:
ok

PR #1875, opened 2018-02-02
Fix the Edit and Reply cancel methods

valgalin:
I noticed that the "Edit" and "Reply" cancel methods in CommentsItem.jsx file are not working. For example, when I edit a comment, the "Cancel" option is not working. Likewise with the Reply form.

I confirmed with @justinr1234 for this fix and he said that this is okay because "... SmartForms code already prevents default before calling this callback."

For more details, please see issue reference: https://github.com/VulcanJS/Vulcan/issues/1870

Thanks!


SachaG:
Great! But can you make a PR with just the `CommentsItem.jsx` changes (and not `versions`)?

Also, just fyi I will move all example code (including the Forum example) to the starter repo soon (https://github.com/VulcanJS/Vulcan-Starter/)

PR #1879, opened 2018-02-03
Create Stripe Subscriptions on startup if requested

justinr1234:
- Contains a helper function that tries to retrieve the plan with this call: https://stripe.com/docs/api/node#retrieve_plan
- If the plan doesn't exist, it will create it using this call: https://stripe.com/docs/api/node#create_plan
- Configurable via private setting `stripe.createPlans`: `true` or `false`
- Converts the `plan` from Vulcan's `createProduct` into the `id` required by the stripe API
- Converts any properties of the `createProduct` object from Vulcan that aren't required for the Stripe api in their custom `metadata` object. See fields here: https://stripe.com/docs/api/node#create_invoiceitem


SachaG:
Oh cool! Are you using the Stripe features too? I'm glad to get more eyes on that code, it was my first time writing this kind of integration so I would love your feedback. 

Also, I'm realizing that the subscription features are not documented, I should fix that.

PR #1885, opened 2018-02-05
Fix stripe plan startup

justinr1234:


SachaG:
Btw I ended up removing `alwaysUseTest` because I thought people can manually input a test key as a real key. But do you think it's useful?

justinr1234:
I liked it because it makes keeping your settings clear.

PR #1887, opened 2018-02-05
Add stripe callbacks

justinr1234:
Allow modifying data sent to stripe before it is sent. Allow callbacks to hook into the raw charge data immediately after it goes to stripe.


SachaG:
Nice :)

PR #1891, opened 2018-02-06
Add stripe invoice / subscription items

justinr1234:
Associated docs update: https://github.com/VulcanJS/vulcan-docs/pull/76

### DIFF
--- a/packages/vulcan-payments/README.md
+++ b/packages/vulcan-payments/README.md
@@ -32,8 +32,23 @@ As well as the following private setting (can be stored in the setting's root or
     }
   },
   "stripe": {
-    "secretKey": "sk_test_sfdhj34jdsfxhjs234sd0K",
-    "createPlans": false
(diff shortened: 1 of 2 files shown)


SachaG:
I wonder if the naming isn't getting a bit confusing? Because you add a `plan` property to create a subscription, but a `subscription` property to add an invoice item…? 

Maybe I need to refresh my memory as to how Stripe plans/subscriptions/invoice items/etc. work but I have to confess I'm not super clear on how this all fits together right now…

SachaG:
Maybe you could give a simple example of how you'd use all of these together? What was your use case for adding these features?

justinr1234:
If you look at the docs, it shows what’s required to use the different stripe features.

- Charge
- Subscription
- Subscription Item (basically a subscription that belongs as a sub item on a subscription)
- Invoice Item (1 off charge that is charged within a subscription cycle)

The required items for those are in the docs and the logic I implemented follows that logic.

PR #1901, opened 2018-02-12
center submit and cancel button

pravdomil:
**before**
before

**after**
after

### DIFF
--- a/packages/vulcan-forms/lib/components/FormSubmit.jsx
+++ b/packages/vulcan-forms/lib/components/FormSubmit.jsx
@@ -17,6 +17,7 @@ const FormSubmit = ({
                     }) => (
   <div className="form-submit">
     
+    <div className="text-xs-center">
     <Button type="submit" bsStyle="primary">
       {submitLabel ? submitLabel : <FormattedMessage id="forms.submit"/>}
     </Button>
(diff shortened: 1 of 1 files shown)


pravdomil:
@vincro yes you can do that but by default it should be in center or on right right side, I think that's how it's done in most forms on the internet, what do you think?

justinr1234:
All a matter of personal preference, there is no standard way of doing it. Left aligned is probably the safest. Probably aligning it to the left side of the text boxes even.

SachaG:
Thanks for the PR!

But yeah I'm not sure if the current alignment is the best, but centering feels even weirder. I'd rather leave it be and let people change their CSS if they want to, sorry :(

P.S. for future PRs, try to make them to the `devel` branch :)

justinr1234:
There should be a way to default the repository settings to make PRs against devel

pravdomil:
@SachaG change your default branch to `devel` in github repo settings

SachaG:
Yes I agree on the right would probably be better. But I don't want to add extra styles at this point. Maybe in the future if we use Styled Components. 

And for the branch, if I do that it will also default to `devel` when people clone the repo, I'm not sure if that's what we want.

pravdomil:
@SachaG that's how https://github.com/meteor/meteor done it (devel is their default branch)

SachaG:
Yeah ok I guess we can do the same then.

pravdomil:
@SachaG ok and center that button please

PR #1904, opened 2018-02-16
Various minor bug fixes

ErikDakoda:
- Correct Callbacks schema to ensure compatibility with more recent versions of simpl-schema
- Add option to VulcanEmail.send to not swallow errors
- Only use md5 module of crypto-js to reduce bundle size


SachaG:
:+1:

PR #1906, opened 2018-02-18
 Apollov2 2nd update

MHerszak:
This is the first stage of getting rid of redux. ApolloClients has no reducer anymore. The container HOCs will have to fetch from cache and Minimongo must go. Anyway, this version puts data into the cache but the components still refer to old reducers. So this will be the next update. Talking about getting rid of Minimongo, the subscriptions will have to be invoked somewhere. There is a pull request for subscriptions which puts subscriptions into the default Mutations which isn't the correct place for them, in my opinion. Some thoughts on where to put subscriptions, either you put a re-fetch into a Component mutation, or you utilize the middleware context, or you do some work on the cache which I have not figured out yet. However, I prefer the middleware solution for now. On that note, subscriptions should go in a separate package since those might not be necessary for all collections, this was also proposed in the pull request. Here is a thought querying several ApolloClient related issues. ApolloClient seems to change faster than I like and most people are unhappy that MDG doesn't provide for a redux reducer anymore. They would have hoped for a softer cut. I do feel this will put a dent in a fast adoption for ApolloClient 2.0. However, I really like what MDG delivered so far!

### DIFF
--- a/packages/vulcan-core/lib/modules/callbacks.js
+++ b/packages/vulcan-core/lib/modules/callbacks.js
@@ -1,9 +1,9 @@
 import { addCallback, getActions } from 'meteor/vulcan:lib';
 
 /*
-  
-  Core callbacks 
-  
+
+  Core callbacks
+
 */
 
 /**
@@ -13,7 +13,7 @@ import { addCallback, getActions } from 'meteor/vulcan:lib';
  * @param {Object} Apollo Client reference instantiated on the current connected client
  */
 function RouterClearMessages(unusedItem, nextRoute, store, apolloClient) {
-  store.dispatch(getActions().messages.clearSeen());
+  // store.dispatch(getActions().messages.clearSeen());
   
   return unusedItem;
 }
--- a/packages/vulcan-lib/lib/client/render_context.js
+++ b/packages/vulcan-lib/lib/client/render_context.js
@@ -15,8 +15,8 @@ export const initContext = () => {
   const history = browserHistory;
   const loginToken = global.localStorage['Meteor.loginToken'];
   const apolloClient = createApolloClient();
-  addReducer({ apollo: apolloClient.reducer() });
-  addMiddleware(apolloClient.middleware());
+  addReducer({ apollo: apolloClient.store });
+  // addMiddleware(apolloClient.middleware());
 
   // init context
   context = {
@@ -37,24 +37,23 @@ export const initContext = () => {
     return next => (action) => {
       if (!chain) {
         chain = context.getMiddlewares().map(middleware => middleware(store));
-        newDispatch = compose(...chain)(next)
+        newDispatch = compose(...chain)(next);
       }
       return newDispatch(action);
     };
-  })
-}
+  });
+};
 
 // render context object
-export const renderContext = { 
+export const renderContext = {
   get: () => {
 
     if (typeof context === 'undefined') {
       initContext();
     }
 
-    return context
-
-  } 
+    return context;
+  }
 };
 
 // render context get function
--- a/packages/vulcan-lib/lib/modules/apollo.js
+++ b/packages/vulcan-lib/lib/modules/apollo.js
@@ -1,4 +1,9 @@
-import ApolloClient, { createNetworkInterface, createBatchingNetworkInterface } from 'apollo-client';
+import { ApolloClient } from 'apollo-client';
+import { InMemoryCache } from 'apollo-cache-inmemory';
+import { createHttpLink } from 'apollo-link-http';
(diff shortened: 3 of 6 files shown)


SachaG:
> ApolloClients has no reducer anymore. 

Do you mean the `reducer` option in `withList`? Is that gone? If so that's a problem, maybe we should address this first before moving to 2.0. 

> Minimongo must go

We don't use Minimongo though, are you referring to [Mingo](https://github.com/kofrasa/mingo)? We could remove it but I'm not sure what to replace it with. I'd rather keep it for now. 

> Talking about getting rid of Minimongo, the subscriptions will have to be invoked somewhere. 

We also don't use subscriptions, so I'm not sure what you're referring to?

> most people are unhappy that MDG doesn't provide for a redux reducer anymore.

I think the lack of Redux is not a huge problem for Vulcan, but yeah I'm surprised that there is no easy upgrade path to keep using it.

MHerszak:
Regarding subscriptions, there is a pull request for Vulcan which includes the use of GraphQL subscriptions. But I don't think the events to invoke those subscriptions are in the right place, yet. I also don't think every collection needs subscriptions. This is why I would put it into a separate package. This comment was only a side note. I was under the assumption we would make use of subscriptions using ApolloClient 2.0?

"ApolloClient 2.0 replaced the redux cache with an abstract cache implementation. `apollo-cache-inmemory` is the recommended cache implementation for Apollo Client 2.0. `InMemoryCache` is a normalized data store that supports all of Apollo Client 1.0's features without the dependency on Redux."

The cache utilizes `__typename` for storing and updating. We can still make use of `dataIdFromObject` which will update a warm cache and force a re-fetch when the data you compared with `dataIdFromObject` has changed.

We will retrieve the 'reducer' data from the client cache and `read` and `write` queries to access the client cache directly.

Regarding Mingo, I was skipping through and saw what I thought was Minimongo :D, nvm. ApolloClient has filters to 'display' seperate views, for instance. And you can query it as if you would call a GraphQL query. Meaning, `queryReducer` might still be useful but `Mingo` can be replaced because the client cache has its own query language to direct access the `InMemoryCache`.

``` Example of a filter as a Mutation
const VISIBILITY_MUTATION = gql`
  mutation SetFilter($filter: String!) {
    visibilityFilter(filter: $filter) @client
  }
`;
```
There are still a few loose ends, but I keep working on it. Perhaps, I could show you my sketches to make all the layers more visible? In the end, it is a change because there will be no Redux anymore.

justinr1234:
Subscriptions would definitely need to be separate because implementing them with the default one is not production worthy. To make a production worthy one, you need to use something like redis.

MHerszak:
@justinr1234 can you elaborate on that? What do you mean by 'production worthy'? Any why would we need redis?

justinr1234:
https://github.com/apollographql/graphql-subscriptions/blob/master/README.md

>>>
Note that the default PubSub implementation is intended for demo purposes. It only works if you have a single instance of your server and doesn't scale beyond a couple of connections. For production usage you'll want to use one of the PubSub implementations backed by an external store. (e.g. Redis)

MHerszak:
That means Vulcan would provide an interface to a messaging system, correct?

justinr1234:
I believe it would mean we’d provide a wrapper around the subscriptions systems that Apollo supports like mqtt and redis. So similar to how Vulcan wraps the Apollo server details, we would either wrap the service or allow config for people already running an external server.

So for example, if you didn’t set any settings we’d default to EventEmitter version (the non-production ready version). If you set `redis`, but non-external, we could launch and manage a redis instance. If you set `redis`, but you managed it externally, we’d just use your config details and setup subscriptions assuming your external redis server.

justinr1234:
End of the day using the subscriptions in practice is as easy as `pubsub.publish` and something like `withSubscription` HoC.

SachaG:
> Regarding Mingo, I was skipping through and saw what I thought was Minimongo :D, nvm. ApolloClient has filters to 'display' separate views, for instance.

The idea behind using Mingo was to let you reuse the same parameters/terms on the client and server for sorting and filtering. I see 3 options basically: 

1. (current) use Mongo syntax on both client and server.
2. use Mongo syntax on server and filter syntax on client.
3. use filter syntax on both. 

Option 3 would mean "translating" filters into Mongo objects on the server, so kinda like what we do with connectors. It's probably the "right" solution, but outside the scope of the Apollo 2 migration I would say.

justinr1234:
What is filter syntax? I like keeping Mongo. For instance, graphcool made their own and I feel like it’s very clunky compared to Mongo: https://blog.graph.cool/designing-powerful-apis-with-graphql-query-parameters-8c44a04658a9

PR #1907, opened 2018-02-18
Fix TrackerComponent and include in repo to fix Accounts.LoginFormInner

### LINKED ISSUE #1815: Accounts button disabled
Hi,
The account buttons are disabled in production [(app here)](https://simply-charge-staging.herokuapp.com/), though they are working if I manually disable them in the React console.

It seems that the `Accounts.loginServicesConfigured()` events does not happen, so the buttons are never activated (waiting state [here](https://github.com/VulcanJS/Vulcan/blob/master/packages/vulcan-accounts/imports/ui/components/LoginForm.jsx#L101)).

I also have this empty collection in my DB:
`meteor_accounts_loginServiceConfiguration`

The bug appeared, of course, totally widly, the app use to work correctly.

Root URL seems correct too:
`ROOT_URL:        https://simply-charge-staging.herokuapp.com`

nolandg:
Fixes #1815 by fixing `tracker-component` (which `LoginFormInner` extends) to no longer use legacy internal React variable and includes TrackerComponent directly in this repo instead of relying on external out-of-date npm package. This fixes `Accounts` buttons appearing to be disabled in production mode because their internal states were not being set properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,6 @@
     "speakingurl": "^9.0.0",
     "stripe": "^4.23.1",
     "styled-components": "^2.1.1",
-    "tracker-component": "^1.3.14",
     "underscore": "^1.8.3",
     "url": "^0.11.0"
   },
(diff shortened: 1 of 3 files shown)


SachaG:
Awesome! This will be super helpful, thanks! :)

PR #1913, opened 2018-02-23
Add incorrect password string to en-us i18n

nolandg:
I know sometimes you might not want to differentiate between incorrect password vs login but that logic should not happen at the level of translation. Without this string the error is just "Unknown".

### DIFF
--- a/packages/vulcan-i18n-en-us/lib/en_US.js
+++ b/packages/vulcan-i18n-en-us/lib/en_US.js
@@ -2,6 +2,7 @@ import { addStrings } from 'meteor/vulcan:core';
 
 addStrings('en', {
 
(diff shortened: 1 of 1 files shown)


nolandg:
Sorry, forgot to make a new branch. Hope you don't find 2 diverse fixes in one. This also fixes a problem with my earlier fix of TrackerComponent. I didn't realize `LoginFormInner` was relying on functional `setState()`. This is what was preventing error messages from getting to the child components.

SachaG:
No problem, thanks! And the whole `LoginFormInner` thing was a hack to "outsource" state handling to another component when it was broken because of the TrackerComponent thing that you fixed. So we can probably get rid of the outer wrapper now (although I also want to just scrap the whole accounts package and rewrite it from scratch at some point).

PR #1918, opened 2018-02-28
Update es_ES.js

codedeep:
Add translation to "accounts.error_incorrect_password" and fix emailRequired to email_required


SachaG:
👍

PR #1921, opened 2018-03-01
Fix button position

codedeep:
correct buttons in form submit


SachaG:
I don't know if that button position is a good idea, because the space next to the "submit" button is usually reserved for a "cancel" action. In this case it's "delete", which is very different and also a destructive action. So I don't think we should have any ambiguity here?

codedeep:
fix botton delete in formsubmit

PR #1923, opened 2018-03-06
vulcan:debug : Components Dashboard now shows all hocs

Apollinaire:
Previously, the hocs registered with options (ex: `registerComponent('Foo', FooComponent, [withBar,barOptions])`   ) were not shown in the dashboard from the debug package


SachaG:
Oh thanks! By the way it's not a big deal but we tend to use single quotes everywhere :)

Apollinaire:
Ok i'll be careful about this for next PRs :)

Apollinaire:
Actually it's Prettier that changed all the single quotes to double. I'll config that so i don't have to think about it anymore

PR #1925, opened 2018-03-10
Remove sanitizeHtml import on client

ochicf:
By importing this module on client many other server side only modules end up being bundled to the client and causing some issues with older browsers or [Cordova](https://github.com/VulcanJS/Vulcan/issues/1798), since those modules are not transpiled.

I've just refactored the code to prevent statically importing `sanitizeHtml` on the `modules/utils.js` (that is bundled to the client) and that it is only imported on the server.


SachaG:
Oh nice work, thanks!

PR #1927, opened 2018-03-14
Fix TypeError: this.formControl is undefined @ FormComponent.jsx:30:6

ochicf:
Some user's of our site are experiencing this issue when submiting a form on Windows 7 and 10, both in Chrome and Firefox.

I've changed the check performed in order to update current values: `this.formControl` is not checked for null strict equality, but to safe retrieve the `getValue` property from it.

Another more restrictive option would be to check if `getValue` is indeed a function, but I think is unnecessary since `this.formControl` is being set in the `refFunction`.

### DIFF
--- a/packages/vulcan-forms/lib/components/FormComponent.jsx
+++ b/packages/vulcan-forms/lib/components/FormComponent.jsx
@@ -27,7 +27,7 @@ class FormComponent extends PureComponent {
 
   handleBlur() {
     // see https://facebook.github.io/react/docs/more-about-refs.html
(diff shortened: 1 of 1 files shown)


SachaG:
Hmm I wonder if this is the best solution? Wouldn't it work with just `if (this.formControl)`?

ochicf:
Yes, I think that'd work too in the context we are dealing with, but I was just making things safer ;)

I'll commit the change.

ochicf:
Closing since this has become outdated.

PR #1941, opened 2018-03-29
use this.getCollection() instead of props.collection

Apollinaire:
In vulcan-form, in Form.jsx, convertschema was using props.collection instead of this.getCollection() that was creating an error on using collectionName and not collection in Smartform: 
```js
<Components.SmartForm collectionName="myAwesomeCollection"/>
```

This fixes it


SachaG:
thanks!

PR #1945, opened 2018-03-30
[forms] Do not display a negatve character count

Apollinaire:
In updateCharacterCount, test if the character count is positive, if not return 0


SachaG:
Isn't it more informative to show the negative count than show 0?

Apollinaire:
I think that for us devs it is, but for the normal user it can be weird and not intuitive. 
In my opinion the best solution would be to block adding any additional character once the counter reaches zero, but that would complexify the code for updatevalue and getvalue ( that i don't understand already haha )

PR #1953, opened 2018-04-11
Update sample_settings.json

jkbaseer:
Adding Mailchimp settings if you are configuring Subscribe button and daily automated emails.


eric-burel:
Closing as settings are registered, so they now appear in the `/debug` interface, I don't want to bloat the sample settings with too specific features.

PR #1959, opened 2018-04-20
Fix nested forms after props renaming

Apollinaire:
Commit 45d508c broke Nestedforms, this should do the trick. 
On Form.jsx, control was replaced by formInput in 45d508c, it should have been input. 
On FormNested, added inputProperties to omitted props passed to FormNestedInput, to avoid overwriting inputProperties of nested fields


SachaG:
Good catch, thanks!

PR #1960, opened 2018-04-20
Fix Form props

justinr1234:
Props don't get recomputed when they change causing stale values to show up. This happens if you change routes on your page (lets say PostsSingle) without reloading the page. This will cause react to not re-load the components and so therefore since these values are only constructed in the constructor, they will never be recomputed when the props change.


SachaG:
Hmm that's definitely a problem. Putting this stuff in `render` seems a bit hacky though, wouldn't using a lifecycle hook be better?

justinr1234:
@SachaG updated

SachaG:
Makes sense, thanks!

PR #1961, opened 2018-04-21
Add MongoDB aggregation to Collections

Sebi55:
It would be great to allow mongodb aggregation. I am personally using geoNear for example.
Since version 1.9.0 there is also connectors for mongodb. Maybe it would be better to add it there?
This was my solution for earlier releases.


SachaG:
I'm not super familiar with Mongo aggregation. Is it something that would have to be added in core, or could you instead add it to specific collections once they are created?

PR #1963, opened 2018-04-24
Decompose complex objects to array of _id for CheckboxGroup

justinr1234:
For CheckboxGroup, if your schema returns an array of objects (such as Categories for Posts), then the new form code will use that array of objects as the value for a CheckboxGroup. In this instance, we should check if the current value of `document` is an array of objects and then build a list of `_id` since this is a sane default behavior.

### DIFF
--- a/packages/vulcan-forms/lib/components/FormComponent.jsx
+++ b/packages/vulcan-forms/lib/components/FormComponent.jsx
@@ -198,10 +198,18 @@ class FormComponent extends PureComponent {
           return <Components.FormComponentCheckbox {...properties} />;
 
         case 'checkboxgroup':
-          // formsy-react-components expects an array value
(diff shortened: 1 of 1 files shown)


SachaG:
Isn't that a bit too much magic? Also even in the post categories example, you should still be dealing with an array of strings, right? I kinda see what problem you're addressing but I'm not convinced this is the right way/right place to address it. 

Also I think the way to get the value should be `this.getValue()` here?

SachaG:
I think the problem here comes from the schema field and GraphQL field having the same names? If you look at the current forum code that field is named `categoriesIds`.

SachaG:
https://github.com/VulcanJS/Vulcan-Starter/blob/master/packages/example-forum/lib/modules/categories/custom_fields.js#L12

justinr1234:
Yes this is exactly right. Thanks!

PR #1964, opened 2018-04-27
Fix field errors display

Apollinaire:
Added `error.properties` to follow fae7b5a03282b1e1586f5483a1f2717ed22c8cef .
Kept `error.data` for backwards compatibility, not sure that it's needed. If it is, it might be good to add it to FormErrors too


SachaG:
Thanks! Yeah let's keep `data` for now. I'm going to write some docs for error handling soon, I think it'll come in handy.

Issue #1967, opened 2018-04-30
Form components rerendering issue


SachaG:
Even if `FormComponent` doesn't re-render, its form input component children sometimes do. Use https://github.com/acdlite/recompose/blob/master/docs/API.md#pure to prevent this?

justinr1234:
Technically you shouldn't be mutating your props anyways so I guess this is good. I'd say leave the PureComponent which will force people into following the pattern that you shouldn't mutate your props. So if you need to add an item to an array, you should return a new array reference rather than adding it to an existing array.

render() {
   const arrayWithAdditionalValue = [...this.props.originalArray, 'extraValue'];

   return <ChildComponent theArray={arrayWithAdditionalValue} />;
}

ochicf:
I've a proposal that would decrease form controls rerendering: do not pass the entire document to each form control by default.

Right now the entire document is passed to every form control, making that any change in the form will rerender everything. I think that most of form controls do not need the entire document: they have enough with their value.

Of course more complex controls (usually they will be custom) they will need values from other fields, thus needing the entire document. Right now they will be able to retrieve them by retrieving the `getDocument` function from context, but I'd also pass the document in the contex to ease that use case, named either `document` or `currentDocument` (I think the latter matches better existing context properties such as `initialDocument`, `currentValues`, etc.).

Example of field using the another field's value:

```js
import PropTypes from 'prop-types';
import getContext from 'recompose/getContext';

import MyComponentNeedingEntireDocument from './somewhere';

export default getContext({
  currentDocument: PropTypes.object,
})(MyComponentNeedingEntireDocument);
```

The bad thing about this is that could break existing form controls that rely on the presence of `document` prop, so those would be updated. I've quickly checked the bootsrap package and none of them are using it.

What do you guys think?

SachaG:
I like the idea in theory but when I tried similar ideas I always ended up with code that was too complex (see the whole `getValue` mess we had before). I'm open to suggestions though if you find a clean way to do it.

PR #1968, opened 2018-04-30
Change arrow functions into purecomponents

Apollinaire:
Arrow functions are not Pure and were updating at every form change. See #1967.
At this point, there are still some unwanted updates, but i think these are from `formsy-react-components`, so there is not much we can do about it unless getting rid of Formsy.

### DIFF
--- a/packages/vulcan-ui-bootstrap/lib/components/forms/Checkbox.jsx
+++ b/packages/vulcan-ui-bootstrap/lib/components/forms/Checkbox.jsx
@@ -2,6 +2,10 @@ import React from 'react';
 import { Checkbox } from 'formsy-react-components';
 import { registerComponent } from 'meteor/vulcan:core';
 
-const CheckboxComponent = ({refFunction, inputProperties, ...properties}) => <Checkbox {...inputProperties} ref={refFunction} />
-
+class CheckboxComponent {
+  render() {
+    const { refFunction, inputProperties, ...properties } = this.props;
+    return <Checkbox {...inputProperties} ref={refFunction} />;
+  }
+}
 registerComponent('FormComponentCheckbox', CheckboxComponent);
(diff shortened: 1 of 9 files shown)


Apollinaire:
Don't merge this until i have added the commits that remove formsy from input components

Apollinaire:
Edit: @SachaG I'm sorry but i don't have time currently to do the whole switch and rewrite all the components from scratch. 
Also it seems that 94fc8fffe9ea848c5d3099da93ba0da2172a445e broke error handling at the field level. The field does not pass in error mode, the error text is not displayed and instead the whole form is disabled

SachaG:
Should I still merge this or not then?

Apollinaire:
you can, it saves the updates on the components in ui-bootstrap/Form, but they have to be rewritten from scratch to get rid of Formsy. 
i made the default input like this: 
```js
import React from 'react'; 
import { registerComponent } from 'meteor/vulcan:core'; 
 
class Default extends React.PureComponent { 
  handleChange(event) { 
    this.props.inputProperties.onChange( 
      this.props.inputProperties.name, 
      event.target.value 
    ); 
  } 
 
  render() { 
    const { refFunction, inputProperties, ...properties } = this.props; 
    return ( 
      <div className="form-group row"> 
        <label htmlFor={this.props.path} className="col-sm-3 col-form-label"> 
          {inputProperties.label} 
        </label> 
        <div className="col-sm-9"> 
          <input 
            name={inputProperties.name} 
            id={this.props.path} 
            className="form-control" 
            value={inputProperties.value} 
            onChange={this.handleChange.bind(this)} 
          /> 
        </div> 
      </div> 
    ); 
  } 
} 
```
it's pure bootstrap html, i did not use react bootstrap because i couldn't reproduce easily the actual horizontal layout. This code works on regular fields *but not in nested fields*. FormNested must change the scope or something and the value is not updated.

SachaG:
Maybe I'll wait a bit to merge to `devel` then. I would like to get the UI stuff out first before taking on the performance issues.

PR #1969, opened 2018-04-30
Include document prop in mutationErrorCallback method

bernardodiasc:
This helps with the problem reported in https://github.com/VulcanJS/Vulcan/issues/1958 while using `errorCallback` on `SmartForm` component.

### DIFF
--- a/packages/vulcan-forms/lib/components/Form.jsx
+++ b/packages/vulcan-forms/lib/components/Form.jsx
@@ -594,7 +594,7 @@ class Form extends Component {
       this.props
         .editMutation(args)
         .then(this.editMutationSuccessCallback)
-        .catch(this.mutationErrorCallback);
+        .catch(error => this.mutationErrorCallback(document, error));
     }
   };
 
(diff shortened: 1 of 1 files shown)


SachaG:
Can't you already access the document via the props though? I'm confused by that comment. Also where is this actually used?

bernardodiasc:
If I understood correctly, the `document` here is the data from `submitForm` https://github.com/VulcanJS/Vulcan/blob/devel/packages/vulcan-forms/lib/components/Form.jsx#L537

Example usage:

```
<Components.SmartForm
  collection={Posts}
  mutationFragment={getFragment('PostsPage')}
  errorCallback={(post, error) => {
    // do something here
  }}
/>
```

bernardodiasc:
Good call. Now it's ready for review again. Thanks!

bernardodiasc:
I noticed that this catch callback shouldn't happen for `removeMutation` because it will trigger re-render when the Form component is already unmounted.

SachaG:
Good point!

PR #1973, opened 2018-05-02
Update Flash component and withMessages container

bernardodiasc:
While testing the Posting edit action with latest from `devel` I stumbled on an issue:

```
Form.jsx:517
TypeError: Cannot read property 'id' of undefined
    at Flash._this.getProperties (Flash.jsx:33)
    at Flash.render (Flash.jsx:44)
    at finishClassComponent (modules.js?hash=ffcdd80922dc5d0dd7afe11426e1665a70c9d671:94083)
    at updateClassComponent (modules.js?hash=ffcdd80922dc5d0dd7afe11426e1665a70c9d671:94051)
    at beginWork (modules.js?hash=ffcdd80922dc5d0dd7afe11426e1665a70c9d671:94676)
    at performUnitOfWork (modules.js?hash=ffcdd80922dc5d0dd7afe11426e1665a70c9d671:97508)
    at workLoop (modules.js?hash=ffcdd80922dc5d0dd7afe11426e1665a70c9d671:97537)
    at renderRoot (modules.js?hash=ffcdd80922dc5d0dd7afe11426e1665a70c9d671:97568)
    at performWorkOnRoot (modules.js?hash=ffcdd80922dc5d0dd7afe11426e1665a70c9d671:98143)
    at performWork (modules.js?hash=ffcdd80922dc5d0dd7afe11426e1665a70c9d671:98064)
```

These js undefined errors breaks the entire application frontend.

---

Hi @SachaG, please let me know if this helps or if you have something different in mind.

I noticed recent changes here https://github.com/VulcanJS/Vulcan/commit/a086e6e584cfec0650da0acf92118bf776354b04 and found a fix. Not sure how much this would overall affect the platform tho.

As far as I understood, the `flash` action now will be triggered like this
```
this.props.flash({ id: 'posts.edit_success', type: 'success', properties: { title: post.title } });
```

### DIFF
--- a/packages/vulcan-core/lib/modules/components/Flash.jsx
+++ b/packages/vulcan-core/lib/modules/components/Flash.jsx
@@ -21,19 +21,19 @@ class Flash extends PureComponent {
   }
 
   getProperties = () => {
-    const { errorObject } = this.props;
-    if (typeof errorObject === 'string') {
+    const messageObject = this.props.message;
+    if (typeof messageObject === 'string') {
       // if error is a string, use it as message
       return {
-        message: errorObject,
+        message: messageObject,
         type: 'error'
       }
     } else {
       // else return full error object after internationalizing message
-      const { id, message, properties } = errorObject;
+      const { id, message, properties } = messageObject;
       const translatedMessage = this.context.intl.formatMessage({ id, defaultMessage: message }, properties);
       return {
-        ...errorObject,
+        ...messageObject,
         message: translatedMessage,
       };
     }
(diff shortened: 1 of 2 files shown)


SachaG:
Yes that looks correct :) See also http://docs.vulcanjs.org/errors.html

Btw I do think `errorObject` is a better name than `messageObject` because that object itself can have a `message` property. So I thought naming everything "message" could get confusing…

bernardodiasc:
> Btw I do think errorObject is a better name than messageObject because that object itself can have a message property. So I thought naming everything "message" could get confusing…

Got it, sounds good to me as well. I changed because this object can also hold `success` messages, not only the `error` messages :)

SachaG:
Yeah, I hesitated for the same reason… but I couldn't find another better name…

PR #1974, opened 2018-05-02
Changes to SmartForm behaviour

ErikDakoda:
- Moved UI portions of FormComponent to FormComponentUi.jsx in vulcan-ui-bootstrap
- Added user alert when the user navigates to another route while there are unsaved changed in the form
- Added optional Revert button in FormSubmits to allow the user to discard any changes to the form; this is activated by passing a "revertCallback" property to SmartForm (which can be empty: () => {})
- Added support for "form.getHidden" field property; this should be a function that return a boolean to dynamically show/hide the control based on the value of other document properties; example: getHidden: () => function () { return !this.document.useKey; }
- Note: the existing "hidden" does not work for this because it's evaluated only once, when the form is created, for the same reason getHidden is actually a function that returns a function
- "document" is now passed to form components in props and props.inputProperties
- Added two functions that form components can access in the child context: refetchForm() to refetch the document from the database (in case it was updated by a background process or to revert the form), isChanged() to determine if there are any unsaved changes
- Fixed bugs in Form.shouldComponentUpdate() and Form.getDocument()
- Note: For any phrases I have added to en_US.js I also added it to es_ES.js with the comment // TODO: translate


ErikDakoda:
Hi @SachaG! My first big pull request - let me know if you want me to explain or change anything. Once you accept this I will create another pull request to merge vulcan-material-ui into VulcanJS/Vulcan. No worries, I will continue to maintain it. I can also update the documentation with the new features I added and explain how to switch between ui packages.

packages/vulcan-forms/lib/components/Form.jsx
```diff
@@ -450,7 +466,45 @@ class Form extends Component {
       return newState;
     });
   };
-
+  
+  handleRouteLeave = () => {
```
justinr1234: I would argue behavior should be optional, toggled by a flag on the props, and default of being off. I love the idea, but it would need to be optional for sure for instance in the case where people use their routes for information:

/Page
/Page#value
/Page/value

ErikDakoda: Hi Justin! I don't have a problem adding a prop to SmartForm to toggle this feature or even a global settings file value. Let's see what @SachaG  thinks. But remember this behaviour wouldn't show up on all pages, just pages with a SmartForm on it. Do you have a use case where the route changes but the user stays on the same SmartForm? Maybe we can find a way for handleRouteLeave to detect this.

justinr1234: In general you don't want to have a default behavior for all forms to prompt a user about leaving a route. The decision should be on the developer to manually opt-in to UX behavior that causes a jarring action (prompting). Making this the default would impact any app out there that uses SmartForm (which we don't know how many that is). For them to upgrade and then realize anywhere they're using forms they now have this new prompting behavior would have unintended consequences for developers.

ErikDakoda: I agree, I wasn't disagreeing with you.

justinr1234: Gotcha, the only use case I'd have would be where you'd scroll to different portions of the page or form based on floating UI buttons (up / down arrows for instance).

ErikDakoda: Yeah, that makes sense. I'll test that and make sure the warning doesn't come up in that case. How do you add the hash? Do you do it? Something like `location.hash = newHash; browserHistory.replace(location);` ?

justinr1234: Should be able to do `withRouter` and do `props.router.replace`

justinr1234: But maybe browser history I’d asssume?

SachaG:
Thanks! It'll take me a while to review the whole thing, but as a first thought I'm not sure `FormComponentUi.jsx` whould be in the bootstrap package since it doesn't contain any bootstrap-dependent code? 

Also maybe a better name would be `FormComponentWrapper` or `FormComponentInner`? What do you think?

SachaG:
Oh and I agree with @justinr1234 that the user alert should be optional, if we have it at all. I have to admit I'm a bit wary of adding router logic to SmartForm, it's already complex enough as it is… I suppose there's no easy way to get that same behavior without modifying `Form.jsx` though (unless we can do something using HoCs somehow?) so for now it's probably ok.

Apollinaire:
> I'm not sure `FormComponentUi.jsx` whould be in the bootstrap package since it doesn't contain any bootstrap-dependent code?

The thing with that is if you want to be able to switch styles easily (like just enabling/disabling the ui packages) between bootstrap and material-ui **you need to have every component that renders something you can see on your browser in both packages**, with their own way of doing.
For example, the m-ui `FormComponent` uses its own way of displaying the character count, using a material-ui component, and also its own renderClear icon. We have to split the logic from the displaying everywhere if we want this to work as intended and Vulcan apps to stay pretty :smiley:

SachaG:
@Apollinaire good points. I guess we can leave it in the bootstrap package then.

ErikDakoda:
Hi @SachaG ! I realize this is a complex PR, and I knew I would have to make adjustments to it before it's accepted. The situation was that in my complex forms there were things SmartForms just could not do, so I have been using a modified version of it, which I would like to fold into the core, because I think they are all useful features.

My answers:

> Moved UI portions of FormComponent to FormComponentUi.jsx in vulcan-ui-bootstrap

As @Apollinaire said above, although the bootstrap version of FormComponentUI contains no bootstrap, the Material UI version does contain material-ui. Until now we had a complete duplicate of FormComponent.jsx in vulcan-material-ui, which is dangerous duplication, so I just factored out the display portions. I'd be happy to rename it FormComponentInner.

In fact, I think I will have to do something similar for Upload.jsx, because as it is currently, doesn't match the rest of a Material UI form.

> Added user alert when the user navigates to another route while there are unsaved changed in the form

I will add both a global setting and a SmartForm prop to control this.

> Added optional Revert button in FormSubmits to allow the user to discard any changes to the form; this is activated by passing a "revertCallback" property to SmartForm (which can be empty: () => {})

> I personally don't really see the need for this as you can always press "back", close the modal, reload the page, etc. to discard a form's changes. Also this is not a feature that has ever been asked for, so I'm a bit reluctant to add it.

I am flexible on this one, though it adds little code and adds a nice optional use case. I can get rid of it if you want.

> Added support for "form.getHidden" field property; this should be a function that return a boolean to dynamically show/hide the control based on the value of other document properties; example: getHidden: () => function () { return !this.document.useKey; }

> I can see how that'd be useful for building stuff like conditional forms, but if we do go down that route I would like to spend more time thinking about the proper API. So I'd probably leave this out of this PR, it seems unrelated to the other improvements.

This is an absolute necessity for my forms, so I don't know what I'd do without it as my code is going in to production soon.

> "document" is now passed to form components in props and props.inputProperties

> I'm not sure if that's a good idea, because document will change any time there is any change to currentValues, which could make it harder to control form component re-renders. My thinking was that by only passing currentValues you ensure you're dealing with a much smaller object, which should in theory help with performance. But maybe I'm wrong, what's your use case for needing access to document inside a component? (Also if we do want access to document it's probably cleaner to pass it down as props like everything else instead of using context?)

In my form the user selects an ecommerce platform (Shopify, Magento, OpenCart) and based on that selection, other form fields change, because each platform has a different API and requires different visible fields and different values in other fields. I also have a component that grabs a screenshot based on the value of another form field that contains a URL. currentValues is problematic because it does not take into consideration default and deleted values. After lots of trial and error the only bug free behaviour uses the current state of the full document.

And I am actually passing it down in props.

> Added two functions that form components can access in the child context: refetchForm() to refetch the document from the database (in case it was updated by a background process or to revert the form),

> Are we sure that this.data.refetch() will refetch the right thing? Where would this be called from? Should we think about generalizing this and putting it on the context of withDocument instead?

In my form, once the user has entered the API info for Shopify I present a TEST button that calls the server which in turn calls Shopify and updates the document with the results. When the async call is completed, I refetch the document. I suppose it would be possible to add a context to withDocument to do the same thing.

> isChanged() to determine if there are any unsaved changes

> Wouldn't looking at currentValues accomplish the same thing? Are there cases where you might have changes even with currentValues being empty?

Yes, when, you delete a value, for example.

> Fixed bugs in Form.shouldComponentUpdate() and Form.getDocument()

> What's the idea behind the allow custom controls to determine if they should update part? Also why is getHidden stuff mixed in there?

While it's true that basic components only need to know if their own state has changed to determine if they should be updated, custom controls may depend on other fields changing - so let them use their own componentShouldUpdate. The same with getHidden - controls may show or hide based on the value of other fields changing, for example my ecommerce platform select, or a Country select. Alternatively we could add "form.alwaysUpdate" to the field schema definition instead of just allowing all custom controls to always update.

> My overall feedback is that there's a lot of good ideas here, but this PR probably tries to do a bit too much, and it'd be better to split out the parts that are not controversial so we can merge them first :)

I am open to whatever, so let me know how to go about doing that.

packages/vulcan-forms/lib/components/Form.jsx
ErikDakoda: Hi @justinr1234! I just checked and when only the hash portion of the route changes, the alert does not come up.

justinr1234:
- Yes to revert callback (default optional): This can be useful if the user doesn't want to navigate to reset a form. It may also be that other things are occurring on the page that are preventing navigation reset being a viable option.
- `getHidden`: My feedback on this would be should we be returning `null` for that use case? I would normally assume `hidden` implies that it is a hidden form field. I actually had this conversation with @SachaG previously about what `hidden` means on a form. I think overall this implementation looks fine to me.
- Passing `document`, `help`, `placeholder`, `errors`, `...form` to `inputProperties`: I completely disagree with this being the proper place to pass these things down. I'm fine with passing down props to components to be used, but it shouldn't be inside `inputProperties`. From my understanding, `inputProperties` should only be things that are appropriate to go on a an actual HTML form control. Specifically, if it doesn't work on a `<input>`, `<textarea>`, etc..., then it shouldn't be in `inputProperties`. This specifically becomes a problem because React throws warnings about setting properties on DOM elements that are not actually supported by the DOM. Meaning `<input document={{asdf: 'value'}} />` throws an error. In summary, I like the idea of the control having access to the `document` and other properties you've added, but not inside of `inputProperties`.
- I like the idea of the child component being able to refetch. I don't think it should live in `withDocument`.
- Agree with allowing controls to decide on their updating (opt-in)

ErikDakoda:
As far as inputProperties, to be honest I was confused about what should go into props and what in props.inputProperties. For example here is the bootstrap implementation of Default.jsx:
```
const Default = ({refFunction, inputProperties, ...properties}) => 
  <Input {...inputProperties} ref={refFunction} type="text" />
```
As you can see, all props are thrown away (`...properties` actually serves no purpose here) and only `inputProperties` are passed on to the control.

@Apollinaire's implementation of nested forms for vulcan-material-ui (and I thank him for jumping in while I was busy) is almost identical to the above, except that `<Input/>` is a MUI custom input component and it requires help, placeholder, errors, etc. because it renders them as a unit.

Actually I think what's going on here is that FormComponent should not decide what's a prop and what's an inputProperty. This decision should be made by FormComponentUi because it's an implementation detail that may vary from ui library to ui library.

justinr1234:
Right, I use antd in my project. I just override all the properties and handle as necessary. In your case I’d imagine you overriding Default and passing on more props in addition to inputProperties.

SachaG:
Yeah the distinction is that `inputProperties` are the specific properties that concern the actual input component, because I needed a way to whitelist the properties that are accepted by the different react-bootstrap components. Agree with delegating this to `FormComponentUi`.

SachaG:
Oh one more thing, could you add comments before every method/function indicating what it does? Even if it seems obvious now, when coming back to the code 6 months from now it can be easy to forget all the details.

ErikDakoda:
Okay, no problem, I will go ahead and fix all issues we discussed in this thread and amend the pull request. I appreciate all the feedback.

SachaG:
Also it's not a big deal if it's too much hassle, but if there's any way that you can split the PR in two so we can deal with changes in smaller batches that'd be great. For example, maybe you could separate the `Form` stuff from the `FormComponent` changes?

SachaG:
Btw the current implementation of `hidden` is that if it's `true` (or evaluates to `true`, if it's a function) it's simply removed from the form altogether. So this happens well before `FormComponent` is called: 

https://github.com/VulcanJS/Vulcan/blob/devel/packages/vulcan-forms/lib/components/Form.jsx#L225-L230

I'm not sure why we need `getHidden` on top of this?

ErikDakoda:
@SachaG I will give it one more try to get my use case to work with the existing `hidden` property.

Question for you: to split this pull request in two, do we need to reject this pull release and then I start two new ones?

SachaG:
> Question for you: to split this pull request in two, do we need to reject this pull release and then I start two new ones?

Probably yeah… unless that's too much trouble on your side?

ErikDakoda:
I will give it a shot - I am just worried about interdepencies. I'll do my best.

BTW Good news, I am able to get rid of `getHidden` and accomplish my use case with the existing `hidden`, which is a much nicer solution. I wrote getHidden a while back before a lot of refactoring of SmartForms.

SachaG:
> I will give it a shot - I am just worried about interdepencies. I'll do my best.

Don't worry about it then. If you can simplify/comment your code then it'll be good too :) And glad to hear about `hidden` :)

ErikDakoda:
I have amended this pull request with the following:

- Renamed FormComponentUi to FormComponentInner
- Moved all rendering portions of FormComponent to vulcan-ui-bootstrap FormComponentInner
- Added setting `forms.warnUnsavedChanges` and SmartForm property `warnUnsavedChanges`
- Removed `getHidden` everywhere; when a function, the `hidden` field property now also receives the current state of the document
- Updated `Form.clearForm` and `Form.mutationSuccessCallback` so that the user can continue working on the document after submitting it
- The form now scrolls to the top when a submit results in errors
- Fixed bug in `FormComponent.shouldComponentUpdate`
- Fixed bug in `FormComponent.handleChange` - number fields could not be cleared, only set to 0
- Fixed a bug in `FormComponent.getValue` - it returned the initial value of a checkbox even after it was set to `false`, and a number even after it was set to 0
- Fixed small bugs in `FormErrors` and `FormSubmit`

ErikDakoda:
@justinr1234 and @SachaG I have made all required changes to this pull request.

I didn't mean to include "Refactor of Upload component to be UI agnostic" in this pull request. I didn't realize all my commits would end up here. Let me know what you want me to do.

SachaG:
Yes I think it'd be better to leave the Upload stuff for a different PR. Also it seems like you commited changes to `.meteor/packages` by mistake?

ErikDakoda:
Moved to Pull Request #1980

PR #1975, opened 2018-05-03
withList Loading prop adjustment

Apollinaire:
Follows the recommanded loading from Apollo. See the Apollo docs about [`data.networkstatus`](https://www.apollographql.com/docs/react/api/react-apollo.html#graphql-query-data-networkStatus) and [`data.loading`](https://www.apollographql.com/docs/react/api/react-apollo.html#graphql-query-data-loading).

One of the issue here was that `loading` did not cover the `refetch` method. The `loading` prop will now be set to `true` whenever some query is on flight, might it be an initial query, a refetch, a loadmore, a poll, or a variable change.
We can still differentiate a `loadmore` state from a `loading` state with the prop `loadmore` which behaviour has not changed.

### DIFF
--- a/packages/vulcan-core/lib/modules/containers/withList.js
+++ b/packages/vulcan-core/lib/modules/containers/withList.js
@@ -154,7 +154,7 @@ const withList = (options) => {
                 results = props.data[listResolverName],
                 totalCount = props.data[totalResolverName],
                 networkStatus = props.data.networkStatus,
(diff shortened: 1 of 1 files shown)


SachaG:
This sounds right but I think there was a reason why I wrote it that way, maybe some bug with `data.loading` or something. Anyway we can change it and see what happens.

PR #1976, opened 2018-05-03
Add selector to editMutation & subfields to withMutation

justinr1234:
- Fix nested debug statements in resolvers when there's a graphql exception (prevent a pyramid of debug statements)
- Add `selector` as an option to `editMutation`. This allows you to match not only on `documentId` alone but also on anything you want. It will default to still using `documentId` if you do not provide a `selector`. This allows you to avoid race conditions where you only want to update a document based on a condition. For instance: `editMutation({ set: { wasUsed: true }, selector: { someOtherValue: 'myValue', wasUsed: false } })`. This would only update the record if some other thread or server had not set the value while this thread was attempting to set the same value.
- Allow a selection of subfields to be defined for `withMutation`. This is necessary if your mutation returns a complex object type instead of a basic Scalar. As it stands, the default Vulcan `withMutation` usages only have mutation which return things like `String` or `Float`. If you have a mutation which returns `Comment` for instance, it will fail with the classic GraphQL error `must have a selection of subfields`. This allows you to specify in your `withMutation` options which fields you want to use. `__typename` is included by default if you specify any subfields. Subfields could also be `[]` if you only wanted to include `__typename`.

### DIFF
--- a/packages/vulcan-core/lib/modules/containers/withEdit.js
+++ b/packages/vulcan-core/lib/modules/containers/withEdit.js
@@ -37,8 +37,8 @@ export default function withEdit(options) {
         mutationName = collection.options.mutations.edit.name;
 
   return graphql(gql`
-    mutation ${mutationName}($documentId: String, $set: ${collection.options.collectionName}Input, $unset: ${collection.options.collectionName}Unset) {
-      ${mutationName}(documentId: $documentId, set: $set, unset: $unset) {
+    mutation ${mutationName}($documentId: String, $set: ${collection.options.collectionName}Input, $unset: ${collection.options.collectionName}Unset, $selector: JSON) {
+      ${mutationName}(documentId: $documentId, set: $set, unset: $unset, selector: $selector) {
         ...${fragmentName}
       }
     }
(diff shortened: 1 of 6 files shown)


SachaG:
> Allow a selection of subfields to be defined for withMutation

Wouldn't fragments work better for this?

justinr1234:
Fragments didn’t work for me, wasn’t sure if that’s because of GraphQL not allowing it in mutation definitions or because something was up with my fragment. My fragment was working for queries in my testing but mutation didn’t work

SachaG:
Did you use fragments the same way they're used in, say, `withEdit`? 

https://github.com/VulcanJS/Vulcan/blob/devel/packages/vulcan-core/lib/modules/containers/withEdit.js

SachaG:
Maybe we could leave the `withMutation` thing out of this specific PR and then see if there isn't a way to use fragments?

And just a note regarding the `selector` thing, while I do think it's a good change I also want to revisit mutations completely at some point down the line to make them follow the guidelines outlined here: https://dev-blog.apollodata.com/designing-graphql-mutations-e09de826ed97

justinr1234:
Can you maybe give it a try @SachaG ? I can't seem to figure out why it doesn't like it when I include fragments.

eric-burel:
Closing as stale, feel free to open an issue if some of the features you wanted to add are not yet there in the latest version.

I think you can avoid the race condition now using the new filtering API. If you pass `{_id: "foobar", wasUsed: false}` as your filter, it will do what you want for example and avoid race conditions, I think. 
Note that the `_id` must be in the filter with other field. If you explicitly update per id, and use a filter separately to set the other fields, the filter is just ignored.
Don't hesitate to open an issue if you suspect some race conditions could happen.

PR #1978, opened 2018-05-05
check for redirect before trying to redirect

Apollinaire:
using `withAccess` without setting redirect used to crash the client because there was no test on `redirect`.  New behaviour is: if redirect is not a string then no redirection is done and the component simply is not displayed. If redirect is a string then we push to that direction in the constructor, as it was previously


SachaG:
Thanks!

PR #1980, opened 2018-05-09
Changes to SmartForm behaviour (updated)

ErikDakoda:
- Moved UI portions of FormComponent to FormComponentInner.jsx in vulcan-ui-bootstrap

 - Added user alert when the user navigates to another route while there are unsaved changed in the form (disabled by default)

 - Added setting forms.warnUnsavedChanges and SmartForm property warnUnsavedChanges to enable user alert

 - Added optional Revert button in FormSubmits to allow the user to discard any changes to the form; this is activated by passing a "revertCallback" property to SmartForm (which can be empty: () => {})

 - Added two functions that form components can access in the child context: refetchForm() to refetch the document from the database (in case it was updated by a background process), isChanged() to determine if there are any unsaved changes

 - For any phrases I have added to en_US.js I also added it to es_ES.js and fr_FR.js with the comment // TODO: translate

 - Updated Form.clearForm and Form.mutationSuccessCallback so that the user can continue working on the document after submitting it

 - The form now scrolls the flash message into view when a submit results in errors

 - Fixed bugs in FormComponent.shouldComponentUpdate() and Form.getDocument()

 - Fixed bug in FormComponent.handleChange() - number fields could not be cleared, only set to 0

 - Fixed a bug in FormComponent.getValue() - it returned the initial value of a checkbox even after it was set to false, and a number even after it was set to 0

### DIFF
--- a/packages/vulcan-forms/lib/components/Form.jsx
+++ b/packages/vulcan-forms/lib/components/Form.jsx
@@ -22,7 +22,10 @@ This component expects:
 
 */
 
-import { registerComponent, Components, runCallbacks, getCollection, getErrors } from 'meteor/vulcan:core';
+import {
+  registerComponent, Components, runCallbacks, getCollection,
+  getErrors, registerSetting, getSetting, Utils
+} from 'meteor/vulcan:core';
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { intlShape } from 'meteor/vulcan:i18n';
@@ -34,7 +37,13 @@ import unset from 'lodash/unset';
 import compact from 'lodash/compact';
 import update from 'lodash/update';
 import merge from 'lodash/merge';
+import find from 'lodash/find';
+import isEqualWith from 'lodash/isEqualWith';
+
 import { convertSchema, formProperties } from '../modules/schema_utils';
+
+registerSetting('forms.warnUnsavedChanges', false,
+  'Warn user about unsaved changes before leaving route', true);
 
 // unsetCompact
 const unsetCompact = (object, path) => {
@@ -68,7 +77,7 @@ const computeStateFromProps = (nextProps) => {
 */
 
 class Form extends Component {
-  constructor(props) {
+  constructor (props) {
     super(props);
 
     this.state = {
@@ -119,7 +128,18 @@ class Form extends Component {
 
   */
   getDocument = () => {
-    const document = merge({}, this.state.initialDocument, this.defaultValues, this.state.currentValues);
+    const deletedValues = {};
+    this.state.deletedValues.forEach(path => {
+      set(deletedValues, path, null);
+    });
+    
+    const document = merge(
+      {}, 
+      this.state.initialDocument, 
+      this.defaultValues, 
+      this.state.currentValues, 
+      deletedValues
+    );
 
     return document;
   };
@@ -223,9 +243,10 @@ class Form extends Component {
 
     // remove all hidden fields
     if (excludeHiddenFields) {
+      const document = this.getDocument();
       relevantFields = _.reject(relevantFields, fieldName => {
         const hidden = schema[fieldName].hidden;
-        return typeof hidden === 'function' ? hidden(this.props) : hidden;
+        return typeof hidden === 'function' ? hidden({ ...this.props, document }) : hidden;
       });
     }
 
@@ -404,7 +425,11 @@ class Form extends Component {
     return {
       throwError: this.throwError,
       clearForm: this.clearForm,
-      submitForm: this.submitFormContext, //Change in name because we already have a function called submitForm, but no reason for the user to know about that
+      refetchForm: this.refetchForm,
+      isChanged: this.isChanged,
+      submitForm: this.submitFormContext, //Change in name because we already have a function
+                                          // called submitForm, but no reason for the user to know
+                                          // about that
       addToDeletedValues: this.addToDeletedValues,
       updateCurrentValues: this.updateCurrentValues,
       getDocument: this.getDocument,
@@ -428,7 +453,9 @@ class Form extends Component {
   }
 
   /*
+  
   Manually update the current values of one or more fields(i.e. on change or blur).
+  
   */
   updateCurrentValues = newValues => {
     // keep the previous ones and extend (with possible replacement) with new ones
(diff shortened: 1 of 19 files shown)


SachaG:
Awesome, let me review it :)

packages/vulcan-forms/lib/components/Form.jsx
```diff
@@ -119,7 +128,18 @@ class Form extends Component {
 
   */
   getDocument = () => {
-    const document = merge({}, this.state.initialDocument, this.defaultValues, this.state.currentValues);
+    const deletedValues = {};
+    this.state.deletedValues.forEach(path => {
+      set(deletedValues, path, null);
+    });
```
SachaG: is there a reason we're `set`ting things to `null` instead of using lodash's `unset`?

packages/vulcan-forms/lib/components/Form.jsx
```diff
@@ -450,22 +477,93 @@ class Form extends Component {
...
+    const changedDocument = this.getDocument();
+    
+    const changedValue = find(changedDocument, (value, key, collection) => {
+      return !isEqualWith(value, initialDocument[key], (objValue, othValue) => {
+        if (!objValue && !othValue) return true;
+      });
+    });
+    
+    return typeof changedValue !== 'undefined';
+  };
```
SachaG: I think maybe I asked you before, but why not just look at `currentValues` and `deletedValues` to see if there's any changes?

packages/vulcan-forms/lib/components/FormComponent.jsx
```diff
@@ -1,44 +1,58 @@
...
-    const value = this.getValue(props);
+    this.state = {};
+  }
 
-    if (this.showCharsRemaining(props)) {
-      const characterCount = value ? value.length : 0;
-      this.state = {
-        charsRemaining: props.max - characterCount,
-      };
+  componentWillMount () {
```
SachaG: Any specific reason to put this on `componentWillMount` rather than in the constructor?

packages/vulcan-forms/lib/components/FormComponent.jsx
```diff
@@ -1,44 +1,58 @@
...
-  //   const { path } = this.props;
-  //   const hasChanged = currentValues[path] && currentValues[path] !== this.props.currentValues[path];
-  //   const hasError = !!errors[path];
-  //   const hasBeenDeleted = deletedValues.includes(path) && !this.props.deletedValues.includes(path)
-  //   return hasChanged || hasError || hasBeenDeleted;
-  // }
-
+  shouldComponentUpdate (nextProps, nextState) {
+    // allow custom controls to determine if they should update
+    if (!['nested', 'number', 'url', 'email', 'textarea', 'checkbox',
```
SachaG: Hard-coding a list of field names like this seems weird to me. We could at least abstract it out in a `isCustomInput()` function

packages/vulcan-forms/lib/components/FormComponent.jsx
```diff
@@ -88,7 +100,15 @@ class FormComponent extends Component {
...
-        value = currentValue || documentValue || '';
+        //value = currentValue || documentValue || '';
+        // note: the previous line does not work when a checkbox is 'false' or a number is '0'
+        value = currentValue;
+        if (typeof value === 'undefined' || value === null) {
+          value = documentValue;
+        }
+        if (typeof value === 'undefined' || value === null) {
+          value = '';
+        }
```
SachaG: seems like something weird is going on here?

packages/vulcan-lib/lib/modules/components.js
```diff
@@ -112,3 +113,27 @@ export const populateComponentsApp = () => {
...
   return compose(...sourceComponent.hocs)(targetComponent);
 }
+
+/**
+ * Returns an instance of the given component name of function
+ * @param {string|function} component  A component or registered component name
+ * @param {Object} [props]  Optional properties to pass to the component
+ */
+//eslint-disable-next-line react/display-name
+export const instantiateComponent = (component, props) => {
```
SachaG: What is this for? If it's to let people pass either a component or a component name, it think it's better to just accept two different props (e.g. `afterComponent` and `afterComponentName`); or just force it to be a string. I'm not sure the extra magic is a good idea.

packages/vulcan-ui-bootstrap/lib/components/forms/FormComponentInner.jsx
```diff
@@ -0,0 +1,208 @@
...
+  renderExtraComponent = (extraComponent) => {
+    if (!extraComponent) return null;
+  
+    const properties = this.getProperties();
+    
+    return instantiateComponent(extraComponent, properties);
+  };
+  
+  
+  renderComponent = () => {
```
SachaG: I thought this would stay in `FormComponent` and we would just pass down the component as a prop. I think `FormComponentInner` should stay as light as possible, if it's going to be duplicated between each various UI package.

packages/vulcan-forms/lib/components/Form.jsx
ErikDakoda: 1. If there is anything in defaultValues, isChanged would always be true (even if nothing changed)

2. Entering a value and then clearing it or clearing a value and then entering the same value does not leave currentValues and deletedValues in the same state.

packages/vulcan-forms/lib/components/FormComponent.jsx
ErikDakoda: You can't call setState in the constructor. So the alternative would be code duplication.

packages/vulcan-forms/lib/components/FormComponent.jsx
ErikDakoda: Yes, I agree, that's ugly, I'll create `isCustomInput()`. Inside isCustomInput, could I use `type.toLowerCase() !== type` because custom inputs always start with a capital letter?

packages/vulcan-forms/lib/components/FormComponent.jsx
ErikDakoda: So the original code was:
```
value = currentValue || documentValue || '';
```
But that does not work to set a checkbox to `false` or a number to `0`. So essentially the bug was that if you had a number input with the value "10" and then you typed "0" it would go back to 10. We can't use the || shorthand here.

Maybe I should use something other than `typeof value === 'undefined' || value === null`?

I spent many hours tracking these little bugs down.

packages/vulcan-forms/lib/components/FormComponent.jsx
SachaG: Let's keep `isCustomInput()` like this for now since it's explicit at least. We'll figure out something more stable later.

packages/vulcan-forms/lib/components/FormComponent.jsx
SachaG: I meant you have the `if (typeof value === 'undefined' || value === null) {` line twice. Is that on purpose?

packages/vulcan-ui-bootstrap/lib/components/forms/FormComponentInner.jsx
ErikDakoda: Let me look at this with a fresh mind tomorrow and see what I can do about that.

SachaG: Sounds good :)

packages/vulcan-lib/lib/modules/components.js
SachaG: OK, let's keep it then.

packages/vulcan-forms/lib/components/FormComponent.jsx
SachaG: Would this accomplish the same thing?

```
value = '';
        if (typeof currentValue !== 'undefined' && currentValue !== null) {
          value = currentValue;
        } else if (typeof documentValue !== 'undefined' && documentValue !== null) {
          value = documentValue;
        }
```

ErikDakoda: Yes, I think so.

packages/vulcan-ui-bootstrap/lib/components/forms/FormComponentInner.jsx
ErikDakoda: Looks like I can accomplish what you are looking for. I do like having more in the shared `FormComponent` and as little as possible in the multiple `FormComponentInner`s, as long as each UI has all the control it needs to render in their native way.

ErikDakoda:
@SachaG There is one more issue I just found. I don't think `getDerivedStateFromProps` is ever invoked. I think that's because we are still on 16.2.0 and that's a 16.3.0 feature. Should we add back componentWillReceiveProps for now, or should we upgrade to 16.3.0?

ErikDakoda:
Hi @SachaG I have done all the changes.

SachaG:
Great. I think we can probably simplify/refactor `FormComponentInner` even more down the line but it should work for now.

SachaG:
Thanks for being so understanding and improving the PR btw! :)

PR #1982, opened 2018-05-09
Restrict value altering to allow multiple datatypes in FormComponent

ochicf:
## Problem

Values are casted to the first field's datatype without taking into account whether the field can have other types or if the actual values matches the first type.

## Use case

Have multiple data types for one field. As a concrete example, we are working on a files package that defines a field's type like:

```
const FileOrId = SimpleSchema.oneOf(
  // accept generic object so the file itself can be send the first time and be saved
  // into a files collection
  { type: Object, blackbox: true },
  // once the file is saved, the field will the document id as a reference to the document
  // in the files collection
  String,
);
```

The described problem causes the id to be merged with an object, so if the id is `'1234'` it is converted to `{0: '1', 1: '2', 2: '3', 3: '4'}`.

## Proposed solution

Cast value only if the current value is array or object, and if it is also a valid type. I don't know if it would be necessary perform the same check to the document Value.

### DIFF
--- a/packages/vulcan-forms/lib/components/FormComponent.jsx
+++ b/packages/vulcan-forms/lib/components/FormComponent.jsx
@@ -7,6 +7,8 @@ import { registerComponent } from 'meteor/vulcan:core';
 import debounce from 'lodash.debounce';
 import get from 'lodash/get';
 import merge from 'lodash/merge';
+import find from 'lodash/find';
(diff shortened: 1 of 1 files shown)


SachaG:
Yeah the current function is a bit hacky… I don't know if this is the best solution, but it certainly sounds better than what we have now, thanks!

PR #1984, opened 2018-05-10
Await between hooks in runCallbacks

### LINKED ISSUE #1949: Callback hooks issue with async/await
Currently a `collection.before.sync` callback gets called like this in `mutator.js`:

```
modifier = await runCallbacks(`${collectionName}.edit.before`, modifier, document, ...);
```

So any async/await code inside a callback hook is resolved at the mutator level.

The issue is that if two hooks both have async/await code, since the promise only resolves at the mutator level, it won't resolve *in between* hooks. Meaning that the second hook will receive a Promise as its `modifier` argument instead of a normal object.

The solution would be to `async/await` inside `callbacks.js`, where `runCallbacks` is defined. But the problem is that `runCallbacks` is used in many places where it *can't* (or shouldn't) be async, and handling async properly in all those places would be a big refactoring. 

For now an easy enough workaround is to only use a single async callback on each hook, but it's not a great long-term solution.

ochicf:
Fixes #1949.

Features:

- keeps the `runCallbacks` function syncronous by default
- async context starts when one of the callbacks returns a promise
- callbacks will be executed in order and in serial, as they would when executing them synchronously
- callbacks can be defined return promises or can be defined `async` and `await` for values

I'm adding an example to show how the order execution is kept the same even when adding async callbacks.

```js
import { Meteor } from 'meteor/meteor';
import { addCallback, runCallbacks } from 'meteor/vulcan:core';

addCallback(
  'callbacks.sync.async.test',
  function firstCallback(value) {
    return new Promise(resolve => {
      Meteor.defer(() => resolve([...value, 1]));
    });
  },
);

addCallback(
  'callbacks.sync.async.test',
  function secondCallback(value) {
    return [...value, 2];
  },
);

addCallback(
  'callbacks.sync.async.test',
  function thirdCallback(value) {
    return new Promise(resolve => {
      Meteor.defer(() => resolve([...value, 3]));
    });
  },
);


addCallback(
  'callbacks.sync.async.test',
  function fourthCallback(value) {
    return [...value, 4];
  },
);

(async () => {
  console.log(`running 'callbacks.sync.async.test'`);
  const result = await runCallbacks('callbacks.sync.async.test', []);
  console.log('finished:', result);
})();
```

Output when executed with `debug` setting:

```js
// running 'callbacks.sync.async.test'
// >> Running callback [firstCallback] on hook [callbacks.sync.async.test]
// >> Started async context in hook [callbacks.sync.async.test] by [firstCallback]
// >> Running callback [secondCallback] on hook [callbacks.sync.async.test]
// >> Running callback [thirdCallback] on hook [callbacks.sync.async.test]
// >> Running callback [fourthCallback] on hook [callbacks.sync.async.test]
// fourthCallback
// finished: [ 1, 2, 3, 4 ]
```


SachaG:
Thanks! I confess I'm still a bit shaky with this whole Promises thing so I'm glad you could figure it out :)

SachaG:
I'll fix the conflicts and merge it soon :)

ochicf:
You're welcome. I faced this issue too and I was wanting to solve it for a while now (I bypassed it by checking in every callback if the accumulator was a promise).

I have a comment maybe it's worth mentioning, though. Since the function `runCallbacks` can now handle both sync and async callbacks, but as you commented in the issue there are use cases when only sync callbacks are allowed, I think it would be a good idea to add a flag when registering the hook that can indicate whether the hook is sync only or not. Then, we could check for that flag when the first promise is returned - in a callback for that hook- and warn that an async is not allowed and skip that value.

If you fancy the idea I can elaborate it a bit more on Monday (I'm on mobile now) and implement it.

SachaG:
Perfect, thanks!

PR #1985, opened 2018-05-10
add toggleClassName props to Dropdown

lc3t35:
this solves https://github.com/VulcanJS/Vulcan/issues/1983


SachaG:
Actually I think I liked your idea of adding `toggleProps` better, seems more flexible?

lc3t35:
no, don't close ;)

PR #1988, opened 2018-05-21
Made Upload component UI-independent

ErikDakoda:
Made Upload component UI-independent

- Moved UI-related code from Upload.jsx in vulcan-forms-upload to UploadInner.jsx and UploadImage.jsx in vulcan-ui-bootstrap
- Kept UploadImage.jsx separate so that it can be overridden with custom components that could, for example, show rounded corners for an avatar or allow cropping
- Added several optional field schema properties next to inputProperties.presets:
  - dropzoneStyle: style attribute for the dropzone to set its size
  - imageStyle: style attribute for images to set their size
  - uploadImageComponentName: name of a component to use instead of UploadImage
- Re-added the ability to save single images in String fields; otherwise Upload couldn't be used for Users.avatarUrl, for example


ErikDakoda:
Outdated

PR #1990, opened 2018-05-22
Display field label on error messages in SmartForm

Apollinaire:
So i've worked a little on getting back to the original error displaying of errors in SmartForm, which included the label of the field concerned by the error.

**Current Behaviour**: The label of the field is not passed as an error value. The current messages are, for example : `Field "{name}" is required` where {name} is actually the name of the field in the schema, not its label; So if my field is called `userFavoriteMovie` and the label is "Favorite Movie" the error will display `Field "userFavoriteMovie" is required` instead of `Field "Favorite Movie" is required`, which is in my opinion not user-friendly. 

**Behaviour after this PR**: the errors are enhanced with: `properties.label` and we can now define our error message in i18n as: `Field "{label}" is required` . I used properties.label and not properties.name to avoid overriding the current API where the actual name of the field is returned, because someone could use the name too and i think it makes more sense in terms of naming. 

**I need feedback on the API about keeping properties.label or overwriting properties.name**
I will update the default translations to fit the API we chose in this PR too, so don't merge it now

### DIFF
--- a/packages/vulcan-forms/lib/components/Form.jsx
+++ b/packages/vulcan-forms/lib/components/Form.jsx
@@ -347,6 +347,12 @@ class Form extends Component {
       defaultMessage: this.state.flatSchema[fieldName].label,
     });
   };
+  
+  getLabelFromPath = path => {
+    const splittedPath = path.split('.');
+    const fieldName = splittedPath[splittedPath.length - 1];
+    return this.getLabel(fieldName);
+  }
 
   // --------------------------------------------------------------------- //
   // ------------------------------- Errors ------------------------------ //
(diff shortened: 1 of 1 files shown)


SachaG:
Sorry for not reviewing this sooner! Actually @ochicf submitted a very similar PR except that it converts the name into a label inside the error component, which I think is a bit better (since the error component can be replaced, and gives you more flexibility): 

https://github.com/VulcanJS/Vulcan/pull/2010

Maybe I can close this one?

Apollinaire:
Yes I saw the other PR, it's better to merge the one from ochicf rather than this one

PR #1992, opened 2018-05-23
SubSchema Fields

ErikDakoda:
- Added support for fields that have their own subschema
- The code flattens the sub-fields, so the path for `address: { street }` becomes `'address.street'`
- The parent's values are also merged into each child - this way field properties like `editableBy` and `group` can be set on the parent for all children, and can still be overridden by children
- Minor changes were needed in several places to properly support paths, so `currentValues[path]` becomes `get(currentValues, path)`
- In Form.getData(), replaced underscore's `pick` with lodash's `pick` which properly supports paths and returns a new object
- Fixed a bug in collection.getParameters that made it impossible to specify a `limit` with addView or addDefaultView

### DIFF
--- a/packages/vulcan-forms/lib/components/Form.jsx
+++ b/packages/vulcan-forms/lib/components/Form.jsx
@@ -44,6 +44,7 @@ import compact from 'lodash/compact';
 import update from 'lodash/update';
 import merge from 'lodash/merge';
 import find from 'lodash/find';
+import pick from 'lodash/pick';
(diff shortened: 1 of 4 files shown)


SachaG:
Thanks! I'll add another field to the forms example to serve as a test case for this.

SachaG:
Oh ok, I see how you've implemented it now. I thought you would adapt `FormNested` somehow but you're not even using it, which is probably just as well. OK let's keep it this way for now then :)

PR #1994, opened 2018-06-05
Opencrud

SachaG:


justinr1234:
How does returning a complex type affect caching on the Apollo client side? If we have record with `someId` of type `Movie` and we return a complex type:

```
type SingleMovieOutput{
  result: Movie
}
```

Where result contains an update to `someId`, How do we get Apollo to know that the record with `someId` has been updated?

justinr1234:
The way the delete mutation looks do we need a `DeleteMulti`? `EditMulti`?

justinr1234:
Isn’t `unset` a different concept than `null`? Can you link to the motivation in the article and here?

SachaG:
> Where result contains an update to someId, How do we get Apollo to know that the record with someId has been updated?

I think that happens automatically thanks to the unique `_id` identifier and the `__typename` field.

> The way the delete mutation looks do we need a DeleteMulti? EditMulti?

Maybe later but for now I think focusing on the existing simple CRUD operations is a good first step.

> Isn’t unset a different concept than null? Can you link to the motivation in the article and here?

I guess you could set a field to `null` without actually deleting it but I don't really see the point. `set`/`unset` is a Mongo thing so I wanted to move away from it.

packages/vulcan-core/lib/modules/containers/withMulti.js
```diff
@@ -241,49 +236,47 @@ const queryReducer = (previousResults, action, collection, mergedTerms, listReso
   const mingoQuery = new Mingo.Query(selector);
 
   // function to remove a document from a results object, used by edit and remove cases below
-  const removeFromResults = (results, document) => {
-    const listWithoutDocument = results[listResolverName].filter(doc => doc._id !== document._id);
-    const newResults = update(results, {
-      [listResolverName]: { $set: listWithoutDocument }, // ex: postsList
-      [totalResolverName]: { $set: results[totalResolverName] - 1 } // ex: postsListTotal
+  const removeFromResults = (data, document) => {
+    const listWithoutDocument = data[resolverName].results.filter(doc => doc._id !== document._id);
```
justinr1234: Is there a possibility of this being `null` on `results`?

packages/vulcan-core/lib/modules/default_mutations.js
```diff
@@ -82,164 +83,197 @@ export const getDefaultMutations = (collectionName, options = {}) => {
...
 
-      async mutation(root, { search, set, unset }, context) {
-        const collection = context[collectionName];
+      async mutation(root, { input }, context) {
+        const { selector, data } = input;
+        const collection = context[typeName];
 
         // check if document exists already
-        const existingDocument = await Connectors.get(collection, search, { fields: { _id: 1 } });
+        const existingDocument = await Connectors.get(collection, selector, { fields: { _id: 1 } });
```
justinr1234: Does this still return 1 document? We should make sure to query only 1 document.

justinr1234: Side note, we should probably make a setting for what the `_id` field is called in each schema. Since we are wanting to be truly database agnostic, eventually we need to start replacing all these default `_id` with `getSetting('idFieldName', '_id');`

packages/vulcan-core/lib/modules/default_mutations.js
```diff
@@ -82,164 +83,197 @@ export const getDefaultMutations = (collectionName, options = {}) => {
...
+        const { documentId } = selector;
+        const collection = context[typeName];
 
         const document = await Connectors.get(collection, documentId);
         Utils.performCheck(this.check, context.currentUser, document, context);
 
-        return await removeMutator({
-          collection, 
-          documentId: documentId, 
+        return await deleteMutator({
```
justinr1234: Similar comment as above, we should be passing the entire selector to the `deleteMutator` and let it handle what to do with it.

packages/vulcan-core/lib/modules/default_mutations.js
```diff
@@ -82,164 +83,197 @@ export const getDefaultMutations = (collectionName, options = {}) => {
...
+  return mutations;
 };
 
-
 const registerCollectionCallbacks = collectionName => {
-
   collectionName = collectionName.toLowerCase();
 
   registerCallback({
-    name: `${collectionName}.new.validate`,
```
justinr1234: Should leave the backwards compatible callback names

packages/vulcan-core/lib/modules/default_resolvers.js
```diff
@@ -11,18 +11,22 @@ const defaultOptions = {
   cacheMaxAge: 300,
 };
 
+// note: for some reason changing resolverOptions to "options" throws error
 export const getDefaultResolvers = (collectionName, resolverOptions = defaultOptions) => {
+  // TODO: find more reliable way to get type name from collection name
```
justinr1234: Probably make any TODO comments github issues so we can link them in the code? Should make for managing these types of things easier. Can use github labels as well.

packages/vulcan-core/lib/modules/components/App.jsx
```diff
@@ -50,7 +50,7 @@ class App extends PureComponent {
     this.props.cookies.set('locale', locale);
     // if user is logged in, change their `locale` profile property
     if (this.props.currentUser) {
-     await this.props.editMutation({ documentId: this.props.currentUser._id, set: { locale }});
+     await this.props.updateUser({ documentId: this.props.currentUser._id, set: { locale }});
```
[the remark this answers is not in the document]
SachaG: You're right, I'm sure I missed a couple places. But note that the new HoCs are all backwards-compatible anyway.

packages/vulcan-core/lib/modules/containers/withCreate.js
```diff
@@ -0,0 +1,59 @@
...
+import React, { Component } from 'react';
+import { graphql } from 'react-apollo';
+import gql from 'graphql-tag';
+import { getFragment, getFragmentName, getCollection, createClientTemplate } from 'meteor/vulcan:core';
+
+export default function withCreate(options) {
+
+  const { collectionName } = options;
+  // get options
+  const collection = options.collection || getCollection(collectionName),
```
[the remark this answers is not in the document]
SachaG: yeah good point.

packages/vulcan-core/lib/modules/containers/withMulti.js
```diff
@@ -76,89 +72,86 @@ const withList = (options) => {
...
-                loading = props.data.networkStatus === 1,
-                loadingMore = props.data.networkStatus === 2,
-                error = props.data.error,
-                propertyName = options.propertyName || 'results';
+            // results = Utils.convertDates(collection, props.data[listResolverName]),
+            results = props.data[resolverName] && props.data[resolverName].results,
+            totalCount = props.data[resolverName] && props.data[resolverName].totalCount,
+            networkStatus = props.data.networkStatus,
+            loadingInitial = props.data.networkStatus === 1,
+            loading = props.data.networkStatus === 1,
```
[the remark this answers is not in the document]
SachaG: If I'm not mistaken that `loading` would be `true` even when you click `load more`, which would wipe your already-loaded data. That's why we need to differentiate between the initial loading and subsequent loadings.

packages/vulcan-core/lib/modules/containers/withMulti.js
SachaG: Yes, that's probably a possibility.

packages/vulcan-core/lib/modules/default_resolvers.js
SachaG: Oh how do I do that? Just add the issue number and GitHub will link it up?

packages/vulcan-core/lib/modules/default_resolvers.js
```diff
@@ -58,24 +62,28 @@ export const getDefaultResolvers = (collectionName, resolverOptions = defaultOpt
 
         debug(`\x1b[33m=> ${restrictedDocs.length} documents returned\x1b[0m`);
         debugGroupEnd();
-        debug(`--------------- end \x1b[35m${collectionName} list\x1b[0m resolver ---------------`);
+        debug(`--------------- end \x1b[35m${typeName} list\x1b[0m resolver ---------------`);
         debug('');
 
+        // get total count of documents matching the selector
+        const totalCount = await Connectors.count(collection, selector);
```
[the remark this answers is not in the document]
SachaG: Good idea.

packages/vulcan-core/lib/modules/default_resolvers.js
```diff
@@ -11,18 +11,22 @@ const defaultOptions = {
...
-      description: `A list of ${collectionName} documents matching a set of query terms`,
+      async resolver(root, { input = {} }, context, { cacheControl }) {
+        const { terms = {}, enableCache = false } = input;
 
-      async resolver(root, { terms = {}, enableCache = false }, context, { cacheControl }) {
         debug('');
-        debugGroup(`--------------- start \x1b[35m${collectionName} list\x1b[0m resolver ---------------`);
+        debugGroup(`--------------- start \x1b[35m${typeName} list\x1b[0m resolver ---------------`);
         debug(`Options: ${JSON.stringify(resolverOptions)}`);
         debug(`Terms: ${JSON.stringify(terms)}`);
```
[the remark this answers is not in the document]
SachaG: Maybe you can submit a PR for that once we're done with the other stuff?

packages/vulcan-core/lib/modules/containers/withCreate.js
```diff
@@ -0,0 +1,59 @@
...
+  - createMovie(document)
+    
+*/
+
+import React, { Component } from 'react';
+import { graphql } from 'react-apollo';
+import gql from 'graphql-tag';
+import { getFragment, getFragmentName, getCollection, createClientTemplate } from 'meteor/vulcan:core';
+
+export default function withCreate(options) {
```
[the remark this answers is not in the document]
justinr1234: Functions are hoisted to the global scope and can hide circular dependencies. Load order should be controlled by the files themselves. By doing it as a variable function expression instead of a function declaration, you're creating a file-scoped variable that only exists when that line is loaded which prevents hoisting it outside of the file.

https://stackoverflow.com/a/39582542

Specifically: https://stackoverflow.com/a/336868

packages/vulcan-core/lib/modules/containers/withMulti.js
```diff
@@ -76,89 +72,86 @@ const withList = (options) => {
...
 
     // wrap component with graphql HoC
     graphql(
-
       query,
 
       {
-        alias: 'withList',
-        
+        alias: `with${typeName}s`,
```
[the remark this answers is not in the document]
justinr1234: What are your thoughts on making it customizable if needed and defaulting to `s` otherwise?

packages/vulcan-core/lib/modules/default_resolvers.js
justinr1234: It won't actually link it up, I'm saying when you have a TODO like this, make a github ticket about it and then put the link to the issue inside the TODO comment.

```
// TODO: find more reliable way to get type name from collection name
// https://github.com/VulcanJS/Vulcan/issues/1993
```

packages/vulcan-core/lib/modules/default_resolvers.js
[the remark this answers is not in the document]
justinr1234: https://github.com/VulcanJS/Vulcan/issues/1996

packages/vulcan-core/lib/modules/default_mutations.js
```diff
@@ -4,76 +4,77 @@ Default mutations
...
-
+export const getDefaultMutations = (typeName, options = { create: true, update: true, upsert: true, delete: true }) => {
   // register callbacks for documentation purposes
-  registerCollectionCallbacks(collectionName);
+  registerCollectionCallbacks(typeName);
 
-  return {
+  const mutations = {};
 
+  if (options.create) {
```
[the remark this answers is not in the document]
SachaG: Probably not necessary since enabling/disabling specific mutations is a new feature.

packages/vulcan-core/lib/modules/default_mutations.js
SachaG: Good point, ideally we'd want to be able to make any arbitrary field an index.

packages/vulcan-core/lib/modules/default_mutations.js
SachaG: The callbacks are still there, `registerCallback` is only for the documentation.

packages/vulcan-core/lib/modules/default_resolvers.js
SachaG: good point. Maybe someone can do a search across the repo and create issues for all TODOs sometimes.

packages/vulcan-lib/lib/modules/graphql_templates.js
```diff
@@ -0,0 +1,531 @@
+import { Utils } from './utils';
+
+export const convertToGraphQL = (fields, indentation) => {
+  return fields.length > 0 ? fields.map(f => fieldTemplate(f, indentation)).join(`\n`) : '';
```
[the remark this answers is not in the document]
SachaG: I like using angled quotes everywhere for consistency personally.

PR #2005, opened 2018-06-20
Allow passing multiple args to HOCs

mattblackdev:
I needed this for:
```
registerComponent('Component', Component, [withStyles, styles, { withTheme: true }], withCurrentUser);
```

where ```withStyles``` takes 2 args.

### DIFF
--- a/packages/vulcan-lib/lib/modules/components.js
+++ b/packages/vulcan-lib/lib/modules/components.js
@@ -57,7 +57,11 @@ export const getComponent = (name) => {
   if (!component) {
     throw new Error(`Component ${name} not registered.`)
   }
-  const hocs = component.hocs.map(hoc => Array.isArray(hoc) ? hoc[0](hoc[1]) : hoc);
+  const hocs = component.hocs.map(hoc => {
+    if(!Array.isArray(hoc)) return hoc;
+    const [actualHoc, ...args] = hoc;
+    return actualHoc(...args);
+  });
   return compose(...hocs)(component.rawComponent)
 };
 


SachaG:
Oh good idea. Now that we're changing the `registerComponent` syntax maybe we could also support this syntax?

```
registerComponent({
  name: 'Foo',
  component: Foo,
  hocs: [
    {
      hoc: withStyles,
      args: [styles, {withTheme: true}]
    }
  ]
})
```

Then again it might be best to avoid having too many different conflicting syntaxes. I guess we can keep what we have for now.

mattblackdev:
@SachaG 
I typically see apis preferring an optional ```options``` object for the non-required parameters. So, since the hocs are not required to register a component, my recommendation is this:

```
const compose = [
  [withStyles, styles, { withTheme: true }],
  withCurrentUser
]

registerComponent('Foo', Foo, { compose })
```

I think it might also be okay ditch the nested array and defer with a function. However my gut tells me the debugging package might not be able to get the names of hocs to list in the ```/components``` route:
```
const compose = () => [
  withStyles(styles, { withTheme: true }),
  withCurrentUser()
]

registerComponent('Foo', Foo, { compose })
```
After comparing, I prefer the nested array.

SachaG:
I've already implemented the syntax I described on the `devel` branch actually: 

https://github.com/VulcanJS/Vulcan/commit/ecc97f157e49d821adb9cfc5b55ae926d4273fd5

You never know how things might change in the future so I think making every API take a single object is both more consistent and more future-proof (since you can accept new "arguments" without affecting existing function calls).

Apollinaire:
Good PR,
On a side note, I'm guessing you're using Material-UI, you could actually do `registerComponent('CompName', CompObject, withTheme, [withStyles, styles], withCurrentUser)`

Issue #2002, opened 2018-06-14
Allow the passing of arrays with both functions and strings to the permission system 


Discordius:
**Copied from Slack:** 
Feature request for the permission system: Allow `viewableBy`, `editableBy` and `insertableBy` to take arrays of functions as well as arrays of group names
I have a pretty common use-case where I want certain fields to be accessible by the owner of the document, and by moderators
But right now there is no elegant way to make that happen
I can either make it accessible by moderators, by passing `['moderators']`
Or I can make it accessible by the owners, by passing `Users.owns`
But I can't do both, i.e. this doesn't work: `[Users.owns, 'moderators']`

Apollinaire:
Did the code for permissions change since the `opencrud` update ? If it did not, I could work on this, since it is useful to me too (I have the same use case as @Discordius) but I have not updated yet and won't before a couple of months

SachaG:
It did change a bit so I'd recommend working on the latest `devel` branch if you want to submit a PR for that.

Apollinaire:
Alrigth, I'll work on my older version first (1.9) and will do it after with the most recent version

Apollinaire:
Actually I don't see what changed on the way we treat `viewableBy, editableBy, insertableBy`, to me the functions [`Users.canViewField`](https://github.com/VulcanJS/Vulcan/blob/a2e3ca01cc6de29a6138aee429ef6c9598c8a023/packages/vulcan-users/lib/modules/permissions.js#L192), [`Users.canEditFields`](https://github.com/VulcanJS/Vulcan/blob/a2e3ca01cc6de29a6138aee429ef6c9598c8a023/packages/vulcan-users/lib/modules/permissions.js#L268), [`Users.canInsertFields`](https://github.com/VulcanJS/Vulcan/blob/a2e3ca01cc6de29a6138aee429ef6c9598c8a023/packages/vulcan-users/lib/modules/permissions.js#L256) did not change ? Am I missing something? 

If I'm not missing anything, i have made this that seems to work with v1.9.1 : 
```js
Users.canViewField = function (user, field, document) {
  if (field.viewableBy) {
    if (typeof field.viewableBy === 'function') {
      // if viewableBy is a function, execute it with user and document passed. it must return a boolean
      return field.viewableBy(user, document);
    } else if (typeof field.viewableBy === 'string') {
      // if viewableBy is just a string, we assume it's the name of a group and pass it to isMemberOf
[...]

PR #2006, opened 2018-06-21
Allow the passing of arrays with both functions and strings to the permission system

Apollinaire:
See #2002


Apollinaire:
ping @SachaG I think this is ready to ship. In my last commit I removed the backwards comp. for the function names as discussed in #2002, and changed the names of these functions everywhere I could find it in the code. I also tested inserting a new document, editing a document and viewing a document in example-forms. 
I searched in vulcan-starter and there seem to be no occurences of `Users.canView / Edit / InsertField` so there should be no additional work on these.

SachaG:
Great work. I'll be travelling tomorrow but I'll try to find some time soon to review and merge this.

PR #2010, opened 2018-06-28
Improved form errors i18n

ochicf:
Features implemented in this PR:

1. Customized errors for specific fields by collection
2. Translated field label
3. Error context

#### 1. Customized errors for specific fields by collection

I sometimes find the need to set a more concrete message for a particular field in a given collection. After thinking about this, the simplest solution it seemed to pass a `collection` property along the error when formatting the message, so from there errors can be filtered by collection and path.

For example, to set different messages for different fields:

```
'errors.required': `{name, select,
  image {Please add an image}
  other {Field "{name}" is required}
}`
```

And setting them by collections:

```
'errors.required': `{collection, select,
  user {{collection, select,
    image {You must provide an image so others can see you!}
    other {Field "{name}" is required.}
  }}
  other {Field "{name}" is required.}
}`
```

As I've stated, this is the simplest solution from the standpoint of the codebase, and it delegates the problem to the strings definition/translation.

**IMPORTANT**: note that this feature will only work as described in the examples when using [`react-intl`](http://docs.vulcanjs.org/internationalization.html#Using-react-intl)

#### 2. Translated field label

Pass as property the translated label of the field to format message. This is done in the new `FormError` component, that uses the `Form.getLabel` method -received by context- to retrieve the field label.

So now when defining the message the translated label can be retrieved:

```
'errors.required': `Field "{label}" is required.`
```

Note that `name` property is kept so it can be used to construct strings based on the field name (as explained in the previous feature).

#### 3. Error context

A `context` property is passed to format message. That way a different error message could be shown for inline errors -`field` context, named after the `FieldErrors` component- or the ones group for the entire form -`form` context, named after `FormErrors`.

Then the error string can be registered as:

```
'errors.required': `{context, select,
  field {Required}
  other {Field "{name}" is required.}
}`
```

**IMPORTANT**: note that this feature will only work as described in the example when using [`react-intl`](http://docs.vulcanjs.org/internationalization.html#Using-react-intl)


ochicf:
@SachaG as noted, to really take advantage of 2 of the features the developer would have to use the entire `react-intl` package. **Do you think it would be worth that Vulcan provides all the features of `formatMessage` by default?**

If so, I could try to add it to the PR (using [IntlMessageFormat](https://www.npmjs.com/package/intl-messageformat) and [intl-format-cache](https://www.npmjs.com/package/intl-format-cache)).

SachaG:
This looks great, thanks! 

> Do you think it would be worth that Vulcan provides all the features of formatMessage by default?

What additional feature of react-intl would we need to support exactly?

ochicf:
Right now in Vulcan a simplified `formatMessage` function is defined on the `vulcan:i18n` provider, so in order to have the fully featured `formatMessage` we would have to directly use [IntlMessageFormat](https://www.npmjs.com/package/intl-messageformat).

I've seen that `react-intl` uses [intl-format-cache](https://www.npmjs.com/package/intl-format-cache), and it seems that it is to speed up the formatting (you can read more about this on the package readme).

I understand that the decision of not including any i18n package was to keep the bundle size low, and I'm not sure of how much will it increase by adding this 2 packages... IMHO having all the [features](https://formatjs.io/guides/message-syntax/#selectordinal-format) is really worth it, but I also like approach of keeping Vulcan limited in i18n by default.

SachaG:
So the main feature we're missing is handling plurals/ordinals? I'm just not clear on what part of the PR exactly requires these extra features?

ochicf:
With Vulcan's default `formatMessage` you could not do the examples in points 1 and 3, because the missing `{select}` feature. That is why I was wondering whether to try to include the full `formatMessage` on Vulcan by default, but other than that this PR is fine.

Maybe I should have opened an issue to have that discussion, sorry.

SachaG:
btw: https://github.com/aldeed/simple-schema-js#custom-field-validation

PR #2011, opened 2018-06-28
FormComponent value handling improvements

ochicf:
In this PR I've added the possibility to inject some of the props to `FormComponent` so value handling can be customized, while keeping the same behavior as default.

Changes description:

- `isEmptyValue` is now a prop that defaults to the one defined in `forms` package
- added `emptyValue` prop and use it whenever `''` appeared. It still defaults to `''`
- extracted merge logic into functions: one checks whether value should be merged and the other performs the merge. They can be injected as props in order to customize the behavior of the field

I will enter more in depth to the merge logic. I was having issues with fields stored as objects, especially when updating a field that already had a value, and wanting to change it entirely. With the existing code, I was always getting the result of the merge of the current and the old value, but I only wanted the current (because it has changed entirely).

#### Example of current problem and solution

I have a field that is a range, so it is an object with 2 subfields: `initial` and `final`. Users can only change the `final` subfield, but I need the initial to perform some checks and to show it to them.

Now imagine that the user has already saved the document and the value of the field is:

```
{
  initial: 162,
  final: 170,
}
```

Now the problem arrises when the user canges the value of the input for `final`, because it will always be merged with de document value. Hence, when `final: null`, it will take again the value `170`.

t1aebpvljg

But with after the PR this can be fix with adding the following to the field schema:

```
form: {
  shouldMergeValue: () => stubFalse,
},
```


SachaG:
> I have a field that is a range, so it is an object with 2 subfields: initial and final. Users can only change the final subfield, but I need the initial to perform some checks and to show it to them.

> Now the problem arrises when the user changes the value of the input for final, because it will always be merged with de document value. Hence, when final: null, it will take again the value 170.

What if you update each subfield separately using the `path` system? so `updateCurrentValues({ 'range.final`: null })` instead of `updateCurrentValues({ 'range`: {initial: 162, final: null } })` (if that's what you're currently doing).

I think this way the `null` value would not get overwritten?

packages/vulcan-forms/lib/components/FormComponent.jsx
```diff
@@ -117,27 +117,20 @@ class FormComponent extends Component {
...
     const path = p.locale ? `${this.getPath(p)}.value` : this.getPath(p);
     const documentValue = get(document, path);
     const currentValue = get(currentValues, path);
     const isDeleted = p.deletedValues.includes(path);
 
     if (isDeleted) {
-      value = emptyValue;
+      value = '';
     } else {
       value = mergeValue({ currentValue, documentValue, ...p });
```
ochicf: @SachaG i've kept `mergeValue` as a function to keep it more encapsulated, but now it cannot be passed as prop.

SachaG:
Thanks for the changes! :)

PR #2013, opened 2018-06-30
Dynamic loader improvements

ochicf:
Fixes https://github.com/VulcanJS/Vulcan/issues/1997

- added `delayedComponent` as a simple component creator that will render the given registered component
- deprecate `getDynamicComponent` in favor of `renderDynamicComponent` so the name describes better what it actually does
- allow `dynamicLoader` to be used both with dynamic import callbacks and promises, and use `delayedComponent` so it can be used in contexts where `Components` have not yet been populated


SachaG:
I'm not using dynamically loaded components myself yet but I trust you on this one :)

PR #2019, opened 2018-07-18
Setup an example unit test with Mocha

eric-burel:
I did a few try with Meteor testing. My conclusion is the following:

- for testing packages, Meteor's Tinytest is the easiest solution. `practicalmeteor:mocha` and `meteortesting:mocha` simply did not work as expected. It is unit tests only, but it seems sufficient for testing a package.

- Tinytest assertion system is quite limited and no evolution seems to be planned, so we can use Jest's expect as a standalone for assertions

- Tinytest needs to manually import the tests. This is the most annoying thing on earth and does not allow to put the tests near the sources. But again I have no better solution

- It would be REALLY cool if we managed to use Jest, both for testing Vulcan applications and Vulcan packages. But that's a huge work.

Edit: I actually managed to use `meteortesting:mocha` instead of Tinytest. It is more consistent. Using Jest's expect is still a good idea, it is a more structured lib than Chai and provide a lot of functions.


eric-burel:
I actually got meteortesting:mocha to work but it is very ugly, TInyTest is still better though less consistent with the usual Meteor app testinf workflow.

eric-burel:
Client side tests are not yet working though, see [https://github.com/meteortesting/meteor-mocha/issues/71](https://github.com/meteortesting/meteor-mocha/issues/71)

PR #2023, opened 2018-07-25
[WIP] Feature/nested schema

eric-burel:
This PR tracks the support of nested schemas. Right now only arrays are supported, not simple objects.

Todo:
- [X] setup unit testing of React components
- [X] update the `FormNested` so that it handles both arrays and objects
- [X] update the `schema_utils` `getNestedSchema` to detect nested schemas for object
- [X] update other Form components where necessary
- [ ] check what need to be done server side when generating the graphQL schemas
- [ ] update the Starter with an object field : see WIP [https://github.com/VulcanJS/Vulcan-Starter/pull/68](https://github.com/VulcanJS/Vulcan-Starter/pull/68)
- [ ] manual verification, cleanup


eric-burel:
Thanks for the link I'll check what has been done already.

Yes one use case is reusing the same schema in multiple fields (e.g billing address, physical address, meeting address in the same collection). It will help creating cleaner and more structured schemas, and I think it will also help improve the GraphQL types/queries (if you don't want the address of the user you only remove one field  of your query for example) and MongoDB document structure.

In the Starter `example-forms`, I added a field `meetingPlace` to the `Customer` schema, in addition to the meeting date. This field reuse the `addressSchema`.

If some Googler find this PR and need a workaround: for the moment you can simply flatten your schema, e.g by creating fields named `meetingAdress.street`, `meetingAdress.zipCode` etc. instead of using only one Object field.

eric-burel:
#1992 goal seems to flatten the subschema automatically. This is the fastest to implement and it does the job. However I am not sure that it allows to have actualy nested documents in the database and graphQL types that reflects the nested structure.

Providing users an helper like `combineSchema(initialSchema, injectedSchema, prefix)` may help doing the same without needing to modify the schema management system.
E.g `combineSchema(customerSchema, 'address', adressSchema)` would generate `address.street`, `address.country` fields in the schema. The prefixing makes this reusable for multiple fields.
This is "false" nesting, but it may already fit some use cases.

I opted for supporting an object value, without flattening, so it is a bit more work in the frontend. I will now investigate the changes needed in `schema_utils` and other places.

SachaG:
Yeah I don't remember the details but that PR broke nested forms. I agree not flattening is better, this way we keep closer to the schema.

SachaG:
Can this be merged? I think the GraphQL part of handling nested schemas should probably be a separate PR anyway.

eric-burel:
Supporting `{ someObjectField : new SimpleSchema({ ...} ) }` in a schema seems done and working to me.
If it does not break stuff in your apps we can merge I think, I'll open a new PR for the next improvements (avoiding the need to call `new SimpleSchema` on the subfield, handling graphQL, a few more tests etc.).

SachaG:
I'll run my projects on this branch for a little while to make sure everything works and then I'll merge it.

SachaG:
Actually I think this breaks internationalized fields because they are treated the same way as Nested Object fields.

SachaG:
I ended up doing a lot of changes to the way form values are handled, hopefully it doesn't break what you did. I tested the current nested form arrays example and it seemed to work at least…

eric-burel:
Nice! I'll check the form system more thoroughly and write a few more tests asap (so in the weeks to come), I did not thought of i18n for example.

PR #2024, opened 2018-07-27
Pass opencrud field properties to the form

ochicf:
I was trying to use SmartForm with a collection only using the new OpenCrud API (`canRead` instead of `insertableBy`, etc) and I found that no fields were showing because the schema arriving to `Form.jsx` did not contain those properties.


SachaG:
Oh, good call! Thanks!

PR #2029, opened 2018-08-10
Warn when replacing a non-registered component and register it anyway

ochicf:
@SachaG feel free to improve the warning in order to make it more understandable.

Example of current message:

>Trying to replace non-registered component FormNested. The component is being registered. If you were trying to replace a component defined by another package, make sure that you haven't misspelled the name. Check also if the original component is still being registered or that it hasn't been renamed.


SachaG:
Yeah I guess that could be helpful. Although at some point we should standardize where we do `console.log`, `console.warn`, `debug`…

PR #2030, opened 2018-08-10
[Forms] Add currentDocument to clearForm

Apollinaire:
allows for the form fields to be correctly cleared when calling clearForm with `clearCurrentValues == true`


SachaG:
👍

PR #2031, opened 2018-08-10
Allow registering component with unique argument without hocs

Apollinaire:
Before: 
```js
registerComponent({name: 'MoviesItem', component: MoviesItem});
```
would crash the app with the message `cannot read map of undefined`  because `getComponent` tried to map on `hocs`. 

Now, `hocs` is defaulted to `[]` in `registerComponent` to avoid adding the `hocs` property when you don't need it

### DIFF
--- a/packages/vulcan-lib/lib/modules/components.js
+++ b/packages/vulcan-lib/lib/modules/components.js
@@ -105,19 +105,29 @@ export const populateComponentsApp = () => {
  * an empty array, and it's ok! 
  * See https://github.com/reactjs/redux/blob/master/src/compose.js#L13-L15
  */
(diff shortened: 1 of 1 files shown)


Apollinaire:
I added support for single argument on replaceComponent: `{name: 'MyComponent', component: NewComponent, hocs: [hoc1, [hoc2, option2]]}`
Might create conflicts with #2029 but they should be easy to resolve

Apollinaire:
Resolved merge conflicts

SachaG:
Sorry for not merging this earlier, somehow I missed it. But why is the same error duplicated?

SachaG:
I'll refactor this to keep it a little bit DRYer :)

PR #2033, opened 2018-08-13
[opencrud] update Users mutations to match {selector, data} args

Apollinaire:
The commit a1a3ffbb8ca33c7af04bdc59e1324104144721ce changed mutation arguments from `{input: {selector, data}}` to `{selector, data}`, which broke the users' package custom mutations.


SachaG:
Thanks!

PR #2043, opened 2018-08-30
When changing email address on an account, mark the new address as unverified

jimrandomh:
This closes a minor security hole where if you create an account, verify your email address, then change your email address, the new email address is marked as verified (even though you never verified it).


SachaG:
Thanks! By the way I merged this before I noticed but PRs should be made to the `devel` branch, not `master`.

PR #2045, opened 2018-09-03
[WIP] Improve Users creation/management workflows

eric-burel:
I wish to allow more complex workflow out-of-the box when it comes to user management, mainly targeted at companies. Here is a short summary on possible workflows, and the one I'd like to add:

- [X] Creating your own account, with email and password. Works perfectly fine and fits most B2C use cases.
- [X] As an admin, creating an User with an email and send him an enrollment email to set his password afterward. To create the User, simply use a SmartForm on the Users collection from `vulcan:users`, or use the `newMutation` if you want to create them programmatically. For the enrollment part, you can create a simple Meteor method that calls `Accounts.sendEnrollmentEmail`
- [ ] As an admin, create an User and specify his password. Right now it is doable but this would require a custom form that calls the `Accounts` method, like how it is done in the `AccountsLoginForm` on signup. Check the `vulcan:accounts` package if you need this absolutely.
 A solution could be to change the `onCreate`/`onUpdate` workflow of the password field, or maybe to tweak the Users creation mutations.

**In this workflow, the User should not need to have an email**. This makes sense for small companies whose employees do not necessarilly an email.
- [ ] As an admin, create an User and autogenerate a first password, that will be send to him by email. Cool when you need to create 6 000 users or change their passwords

All those workflow should of course allow password update with a similar system (allow admin to change the password directly, to reset it, or to trigger a reset email).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "Vulcan",
-  "version": "1.12.0",
+  "version": "1.12.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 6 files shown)


Apollinaire:
Even if I find these updates necesary, because they are the usual features most apps need, is it a good idea to implement **now** some features that are tightly related to Meteor when we are thinking about leaving it in say six months ? Most of the work done here will probably have to be done again, and will need extra work to stay backwards compatible (if that's even possible).
Again, I agree that we need these features, they are the most useful and reusable between apps so having them handled in the framework is a real plus.

eric-burel:
Hi,

I don't think it is actually so tightly bound with Meteor, because other auth services would have a similar API (sending enrollment mail, replacing password, etc.). My problematic is more about how to handle the `Users` as a normal collection. Currently it needs a lot of specific code, e.g the `AccountsLoginForm`, and does not allow to reuse the `SmartForm`.

My goal here is not to provide a solution out-of-the-box, but more to check if it is actually doable given the current schema system and to add small improvements that makes it easier to implement.

For example, one way to handle password creation while still using a `SmartForm` on the `Users` collection is to add a dummy password field. However this is not doable in a clean way right now, this would require to be able to add custom fields that are totally ignored by the server on a specific form.

SachaG:
Should some of these changes have their own PRs maybe? Having everything in one PR will make it harder to approve some of the more simple fixes.

eric-burel:
Yep definitely, I did not switch branches because I needed all these changes together to make my app work but I will split it in multiple PRs.

eric-burel:
I've split changes into multiple PRs. I've managed to set up a module that allow the admin to create user and set their password, I am in the process of finishing it. 

It did not require much change in the core eventually, Vulcan offers enough possibilities to implement the quoted workflows, using `addField`, `SmartForm` callbacks, and Meteor methods, so we can wait for the big incoming updates before integrating them out-of-the-box has proposed by Apollinaire.

You can take a look at this project for my implementation: [https://github.com/lbke/awesome-vulcan](https://github.com/lbke/awesome-vulcan) (`featuer/user-management` branch for the moment), especially the `UsersForm` of package `users-admin`.

PR #2047, opened 2018-09-03
fixed issue that hardcoded queries to work with single queries

PolGuixe:
I am using the VulcanEmail.build method to avoid repeating code

### DIFF
--- a/packages/vulcan-email/lib/server/routes.js
+++ b/packages/vulcan-email/lib/server/routes.js
@@ -1,67 +1,51 @@
-import { Picker } from 'meteor/meteorhacks:picker';
-import { runQuery, Strings, getSetting } from 'meteor/vulcan:lib';
-import VulcanEmail from '../namespace.js';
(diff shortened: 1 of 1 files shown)


PolGuixe:
@SachaG you had a TODO where I replaced the code. 

I guess that Vulcan-Starter examples will have to be updates as well.

SachaG:
Thanks! Is this backwards-compatible? If not what are the steps to make it work?

PolGuixe:
@SachaG for production apps yes. For development environments a small change would be required. See below:

```js
// before
{
  // ...
  testVariables: {
    documentId: "12345678asdfghjk"
  }
}

//now
{
  // ...
  testVariables: {
    input: {
      selector: {
         documentId: "12345678asdfghjk"
      }
    }
  }
}
```

PolGuixe:
Thx. Let me know if there are any issues 😉

PR #2049, opened 2018-09-04
Add only document once on withMulti query reducer

ochicf:
### Problem this PR solves

After a create mutation, the created document is added multiple times to lists using `withMulti` HOC.

### Cause

`withMulti` update reducer is executed once for every observed query. So, when performing a create mutation the document is added to the results set as many queries being observed.

### Reproduction of the problem

1. Render two components decorated with `withMulti` HOC for the same type (so, two lists that retrieve the same list of documents).
2. Perform a `createMutation` with a new document that should be added to those lists.
3. The document will be added twice will be added into every list
4. On the next polling, the list will be updated with server data and the duplicated document will disappear

### Solution

Only add the document to the results set if is not already there.


SachaG:
Oh that's awesome, it's been an issue for a long time. Nice job tracking it down!

PR #2050, opened 2018-09-05
Pass collection as property when running new API mutators' callbacks

ochicf:
Simple, just pass collection as a property so callbacks can easilly know to which collection the document being mutated belongs. This is specially useful in the global hooks (such as `*.create.before`) that cannot know it otherwise.

I've only made changes to the new API callbacks.

Since this PR is open, @SachaG do you think we could pass some more info, like `context` for example?


SachaG:
Yeah I don't see why not, passing `context` seems like a good idea :)

SachaG:
Oh but in mutators we don't have access to `context` I think. But yeah we can pass more info if we do need to, there's no real drawback.

PR #2051, opened 2018-09-05
Allow user to return nothing in submitCallback

eric-burel:
Useful when the callback is meant for side effect management, eg storing some element of the form in the state, and do not modify the document.


SachaG:
Writing `data = data` seems weird to me, but I guess in this case it should be fine. THanks!

PR #2053, opened 2018-09-05
Cleaned up the options management in hocs

eric-burel:
Will fix issues when not specifying a `fragmentName` or `fragment` in `withDelete`, now user only need to pass a `collection` and `collectionName` (default fragment will be used in this case). Also factorized code a bit to get a consistent behavior across hocs. 

I wrote test but could not run them, I have trouble with `@babel/runtime`. Also I did not found out how to mock package methods like `getCollection` during tests.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -34,6 +34,9 @@
     "no-await-in-loop": 1,
     "comma-dangle": 0,
     "key-spacing": 0,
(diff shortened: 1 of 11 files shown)


SachaG:
Good idea :) Is it ok to merge now?

SachaG:
:+1:

PR #2054, opened 2018-09-05
Merge the collection schema with the props schema

eric-burel:
User can now pass a custom schema in the `SmartForm` to override default props.
It won't work if fields are added, only if existing fields are modified. It is mostly meant for replacing labels, controls and input options.

I also wrote tests in this PR but couldn't run them yet.

### DIFF
--- a/packages/vulcan-forms/lib/components/Form.jsx
+++ b/packages/vulcan-forms/lib/components/Form.jsx
@@ -80,7 +80,11 @@ const getDefaultValues = convertedSchema => {
 
 const getInitialStateFromProps = (nextProps) => {
   const collection = nextProps.collection || getCollection(nextProps.collectionName);
-  const schema = collection.simpleSchema();
+  const collectionSchema = collection.simpleSchema()
+  // merge with schema passed in the props to allow local overrid of existing field
+  // will provoke errors if the user try to add a field however
+  const schema = nextProps.schema ? collectionSchema.extend(nextProps.schema) : collectionSchema;
+
   const convertedSchema = convertSchema(schema);
   const formType = nextProps.document ? 'edit' : 'new';
   // for new document forms, add default values to initial document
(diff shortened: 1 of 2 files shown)


SachaG:
Isn't there some ambiguity? If I pass a custom schema, I would probably expect that it would override the default one instead of being merged with it? If you really want to merge you should probably do it yourself before passing the schema in the first place I think?

eric-burel:
The problem is that there are also plenty of fields that you would never want to change, like the type field. My use case here is customization of the default schema, so in all the use cases I think of you would usually like to keep the default fields with few changes here and there. 
User can do it himself of course but merging a schema is not that trivial (need a call to `extend`, or to recreate a schema, which are not things user may be used to).
The only issue I see with my proposition is when you want to delete some information, but I don't see any use case in the frontend. If you don't want to use a specific prop on your control you would ignore it anyway and not delete it.

It's debatable indeed, there is no hurry to merge anyway, I ended up not using this prop in my own app.

SachaG:
The use case I was thinking of was if you want to show fewer fields on one form. So you'd specify a new schema with only 3 fields instead of 5, but since it would be merged with the "normal" schema you'd end up with 5 fields anyway. Unless I'm missing something?

eric-burel:
For this you can still pass the `fields` props, since merging happens at the top level before fields are sorted out, it seems to work in my app

SachaG:
That's true, but I was just trying to anticipate how someone might expect how the feature works. They might not realize you need to also specify `fields`. Anyway I don't have a strong opinion either way, it just seems to me like if we don't merge we leave the option of merging outside of the form's logic, but if we do merge there is no way to "unmerge". So that's my reason for preferring not merging.

eric-burel:
Actually I did not notice that it was already used in the `FormWrapper` to override the `collection.schema()`,  I'll check it works as expected (it seems rarely used), because it could already provide a working solution. 
Anyway I think you're right, the fact that we can't unmerge could be annoying.

PR #2056, opened 2018-09-07
vulcan-users - permissions - filter out array based fields

vincro:
When using the resolver for the resolveAs examples from the documentation:

http://docs.vulcanjs.org/field-resolvers.html

I found that the Users.getViewableFields helper was return array based field names, which obviously are not compatible with the selector/options objects passed through to collection.findOne.

I did a search to see where the function is called in the DEVEL codebase and see that it is used only once.  That being said, others may well be using it and require the array field names, so perhaps might help to create a wrapper function that strips them out to use instead, but for now I figured I would submit this for you.


SachaG:
Right, those array fields are not supposed to have `canCreate` or `canUpdate` properties but I suppose if they do have them it's good to filter them out.

SachaG:
Sorry in this case it should be `canRead`. But the point is the same :)

PR #2057, opened 2018-09-07
Support arrays with primitives

eric-burel:
Right now array support is limited to arrays of nested objects.
It is doable to create a custom control but not to automatically respect the schema definition of its child.

```js
myArray: {
type: Array
},
'myArray.$': {
type: Date,
control: 'date',
}
```

 `myArray.$` definition was lost and could not be used by the array input component to detect that it should render an input suited for dates.

 I changed the schema transformation process and cleaned it up a little, so that the `arrayField` and `arrayFieldSchema` are now available to the array input component, allowing it to display the correct input its item automatically.

Edit: I've just noticed that PR is unclean, I'll fix this before we can merge.

Thanks to this we could even support array of primitive values out-of-the-box, however it would require a few changes in the `FormNestedArray` component or even creating a new one, I may come back to this in another PR.


eric-burel:
I cleaned up the PR commits, it's now mergeable

SachaG:
I'm not sure if putting that logic in separate methods makes the code easier to read? Since those methods are only called from there I would probably just leave the code where it was? It's a small detail though so if you think this way is better I'll merge it.

eric-burel:
Hi,

Actually I needed to refactor because usual fields and array item fields have slightly different behaviour regarding to the global schema (e.g permissions of an array item are equal to its parent permission, sometimes you need to use the parent name/schema sometimes not etc.), so I now have 2 `createField` and `createArrayField` methods that both rely on the same functions but with different params.
I would rather have kept only one method but I did not manage to handle the params cleanly.

Also I've noticed that the `input` props was ignored for arrays and nested objects so you could not pick your own input, this is now fixed. 

However it is still ignored when `intlInput` is set because I am not sure of the expected behaviour.

SachaG:
Feel free to merge this when it's ready :)

eric-burel:
It's clean enough to merge. The only feature we lack is custom inputs for objects (eg I have a recipe with an array of ingredients, each ingredient being an object, and I want to create a nice ergonomic input for the ingredients while using the default FormNestedArray input, it won't work). It wont be hard to add but needs cleanup beforehand.

PR #2059, opened 2018-09-10
Minor changes and bug fixes in withMulti, single resolver, schema generation

ErikDakoda:
- Added collection.options.singleResolverName and .multiResolverName to stay DRY
- In withMulti, fixed how loadMoreInc passes variables to fetchMore
- Added the "allowNull" option to single resolver to return null instead of throwing a MissingDocumentError
- Added extended SimpleSchema options to support "unique" field property (field can be used as part of a selectorUnique when querying for data)
- When generating schema, populate fields.selector and fields.selectorUnique


SachaG:
Nice work, thanks!

Btw for next time can you make sure you don't include any formatting change in the PR? It just makes it harder to see what the real changes are.

PR #2060, opened 2018-09-10
only clear current values for new document's form

ochicf:
The current behavior is that forms are entirely cleared always, so when editing a document and submitting, the form appears empty. I don't know if this is the intended behavior, but feels like a bug to me.

This PR makes that form is cleared only partially on when editing a document, so the the current values stay in place.


ErikDakoda:
currentValues contains any changes since the last time the document was saved. So when the form is submitted those values are stored in the database, the document is reloaded, and currentValues can be cleared.

What is the behaviour you are seeing? Are you seeing and empty form after submitting a new document or an existing document that you edited?

ochicf:
Hi @ErikDakoda,

What I'm experiencing is that after submitting a form all of its fields become empty and stay that way.

Note that I've only tested this on a project with many things built on top on Vulcan, so maybe it is breaking due to something else; I'll perform more tests and I'll try to reproduce this issue in a new Vulcan project when I have a moment.

By the way, it makes sense that `currentValues` is cleared when you put it that way, that it only has the changes made in the form, so the solution (if indeed the problem exists) should be another one.

Thanks!

ochicf:
Hi @SachaG and @ErikDakoda, I've made some changes:

- on `new` mutation: everything is cleared, but the initial document is mantained
- on `edit` mutatiotion: current values are cleared, but the edited document is used as both `currentDocument` and `initialDocument`

I've made some quick tests and it seems to work, but I'd much appreciate a revision.

SachaG:
I think the use case that might currently not work is if you don't redirect or close the form after edit. I could also imagine a scenario where you want to implement auto-save and submit the form without clearing it. 

The changes look ok to me I think but the end result is a little confusing. Since those are all APIs internal to the component (I think?) we can probably rename them to be as clear as possible without worrying about backwards compatibility.

ochicf:
Hi @SachaG, I've made some changes in order to simplify `clearForm`'s signature.

Basically I've reduced it to 3 options:
- `{Boolean} clearErrors=true`: if should clear errors, same as before
- `{Boolean} clearValues=true`: if should clear form values, them being `currentValues`, `deletedValues` and `currentDocument`. This basically unites all the previous different options to make it simpler
- `{Object} initialDocument=undefined`: only will apply if `clearValues` is `true`, and this indicates that the form will have this document as its `initialDocument`

Important things to point out:
- `currentDocument` will be cleared  to `prevState.initialDocument` by default, or to the provided `initialDocument`. This makes that the form starts with the same document as the initial (useful to look for changes in fields, etc.)
- default behaviour has changed, as it will clear the entire form by default (instead of just errors as before)

Some examples of uses with the new signature:

```js
// clear form errors and resets values to their inital ones
clearForm();

// clear form errors
clearForm({ clearValues: false });

// reinitialize form with a new document, for example in edit mutation
const updatedDocument = {
  // ... some fields
};
clearForm({ initialDocument: updatedDocument });

// fully clears the form
clearForm({ initialDocument: {} });
```

Let me know what you think.

SachaG:
Why is `package-lock.json` committed? I assume that was a mistake?

Also `document : undefined` seems a bit ugly to me… Can we find a cleaner way?

> default behaviour has changed, as it will clear the entire form by default (instead of just errors as before)

Going back to the first comment in this thread, isn't that the behavior we wanted to avoid originally?

SachaG:
OK so I think I like what you've done but I'm not sure if we even need `clearErrors` and `clearValues` since we always want to do that. 

Basically I would suggest:

- `clearForm` always clears errors, current values, deleted values, etc.
- If you pass a `document` property it will set the form to that, else default to reverting it to the initial document (to make the "revert changes" button work). To erase the form contents, pass `{}`.

In other words we end up with just `clearForm({ document })` where `document` can be:

- create: `{}`
- update: the updated document
- revert: `undefined`

SachaG:
Just realized I haven't merged this yet. @ochicf any feedback on my last comment?

ochicf:
Ups, sorry @SachaG, I read your comments but I've been swamped so I forgot to answer.

Thanks for the feedback, it makes a lot of sense to simplify it more, so I'll make the changes you propose.

By the way:

>>default behaviour has changed, as it will clear the entire form by default (instead of just errors as before)
>
>Going back to the first comment in this thread, isn't that the behavior we wanted to avoid originally?

By that I meant the default behavior of `cleanForm` function, not about the overall behavior expected by changes made by the PR. By your following comment I think you already got that, but I want to clarify it just in case it wasn't clear enough.

Thanks!

SachaG:
Yes, makes sense :)

SachaG:
Nice! Thanks for adding detailed comments, too.

PR #2062, opened 2018-09-11
Added support for cc, bcc, and replyTo fields for email API

PolGuixe:
`VulcanEmail.send` and `VulcanEmail.buildAndSend` now accepts cc, bcc, and replyTo fields.

`VulcanEmail.buildAndSendHTML` doesn't not support cc, bcc, and replyTo in order to keep it backwards compatible.

### DIFF
--- a/packages/vulcan-email/lib/server/email.js
+++ b/packages/vulcan-email/lib/server/email.js
@@ -59,8 +59,7 @@ VulcanEmail.generateTextVersion = html => {
   });
 }
 
-VulcanEmail.send = (to, subject, html, text, throwErrors) => {
-
+VulcanEmail.send = (to, cc, bcc, replyTo, subject, html, text, throwErrors) => {
   // TODO: limit who can send emails
   // TODO: fix this error: Error: getaddrinfo ENOTFOUND
 
@@ -76,6 +75,9 @@ VulcanEmail.send = (to, subject, html, text, throwErrors) => {
   const email = {
     from: from,
     to: to,
+    cc: cc,
+    bcc: bcc,
+    replyTo: replyTo,
     subject: subject,
     text: text,
     html: html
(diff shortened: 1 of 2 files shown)


PolGuixe:
@SachaG you are right. Makes sense to add them to the end. 

There are two more arguments: `text`and `throwErrors` at the end of `VulcanEmail.send`. Then when calling it should we add `undefined` for those variables?

SachaG:
Yeah. Although ideally we would replace all arguments with a single object argument like we've been doing elsewhere, and bypass the issue entirely. You can even do it in a backwards-compatible way with the following pattern:

```
f = (a, b, c) => {
  if (typeof a === 'object') {
    var { a, b, c } = a;
  }
  console.log({ a, b, c });
}
```

PolGuixe:
@SachaG what do you think on the last changes. Now is backward compatible. 

I've discussed it with @ochicf and he suggested to add a `Utils` helpers to deal with this refactoring. See changes. ;)

SachaG:
Nice, but I'm not sure if we really need that helper? If your function is `VulcanEmail.send = (options) => {...}` then `options` is already your object?

Then you just need to handle the backwards-compatibility case with:

```
if (arguments.length === 1) {
  ({ foo, bar, baz } = options)l
} else {
  ([ foo, bar, baz ] = arguments);
}
```

At least I think this should work?

PolGuixe:
It is a different approach. By using the helpers, we avoid repeating the extraction logic for every single function that gets refactored. It was a way to implement DRY. 

You are the owner 😅, let me know if you prefer your way and I'll make the changes. 😉

SachaG:
I just think using the helper in this way makes the code overly complex. Keeping things DRY is a good idea but the downside is that it introduces one more layer of things that can go wrong. So in this case let's KISS instead of DRY ;)

PolGuixe:
Changes made

SachaG:
Nice, thanks!

PR #2065, opened 2018-09-12
Add async hook to RouterHook and provide props as argument

ochicf:
### Purpose of this PR

Pass entire props object to `router.onUpdate` hook.

### Modifications

- `runCallbacksAsync` now can be used on client: it uses promises to defer execution of callbacks instead of `Meteor.defer`
- add `router.onUpdate.async` to avoid breaking changes by changing signature of `router.onUpdate`

### Why I submitted this PR

Ok, this may seem some trivial changes, but I'll try to explain why I needed them and why I think they can be useful.

By default `router.onUpdate` hook is called everytime the **current Vulcan route** changes. Note the importance of it being the Vulcan route: it won't be called between route changes of the same route definition. For example, route changes between the same route but with different parameters won't trigger the `router.onUpdate`.

To make this to happen is quite straight forward (and didn't need any changes to the Vulcan's code): just replace the `RouterHook` component with itself and decorate it with the `withRouter` HOC:

```js
import { replaceComponent, getRawComponent } from 'meteor/vulcan:core';
import { withRouter } from 'react-router';

replaceComponent(
  'RouterHook',
  getRawComponent('RouterHook'),
  withRouter,
);
```

Now it will trigger for each actual route change, because `withRouter` will inject some props that will trigger a rerender.

The problem now is that we cannot access those props from `router.onUpdate` callbacks because they are not passed as arguments. And in my case, I wanted to acces the current route's parameters. With this PR, it can be done:

```js
import { addCallback } from 'meteor/vulcan:core';

addCallback(
  'router.onUpdate.async',
  // eslint-disable-next-line prefer-arrow-callback
  function callbackName({currentRoute, params, ...otherProps }) {
    // do something with `params`
  },
);
```

I hope I explained myself clearly and that you find that it makes sense.

### DIFF
--- a/packages/vulcan-core/lib/modules/callbacks.js
+++ b/packages/vulcan-core/lib/modules/callbacks.js
@@ -1,21 +1,18 @@
 import { addCallback, getActions } from 'meteor/vulcan:lib';
 
 /*
-  
-  Core callbacks 
-  
+
+  Core callbacks
+
 */
 
 /**
  * @summary Clear flash messages marked as seen when the route changes
- * @param {Object} Item needed by `runCallbacks` to iterate on, unused here
- * @param {Object} Redux store reference instantiated on the current connected client
- * @param {Object} Apollo Client reference instantiated on the current connected client
+ * @param {Object} props
+ * @param {Object} props.client Apollo Client reference instantiated on the current connected client
  */
-function RouterClearMessages(unusedItem, nextRoute, store, apolloClient) {
-  store.dispatch(getActions().messages.clearSeen());
-  
-  return unusedItem;
+function RouterClearMessages({ client }) {
+  client.store.dispatch(getActions().messages.clearSeen());
 }
 
-addCallback('router.onUpdate', RouterClearMessages);
+addCallback('router.onUpdate.async', RouterClearMessages);
--- a/packages/vulcan-core/lib/modules/components/RouterHook.jsx
+++ b/packages/vulcan-core/lib/modules/components/RouterHook.jsx
@@ -1,5 +1,5 @@
 import React, { PureComponent } from 'react';
-import { registerComponent, runCallbacks } from 'meteor/vulcan:lib';
+import { registerComponent, runCallbacks, runCallbacksAsync } from 'meteor/vulcan:lib';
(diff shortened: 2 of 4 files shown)


SachaG:
That makes sense, I think. But:

> runCallbacksAsync now can be used on client: it uses promises to defer execution of callbacks instead of Meteor.defer

Are we sure that this won't affect the behavior of current async callbacks on the server? Potentially it could be a change with a pretty big impact. I just don't know how Meteor.defer works behind the scenes compared to Promises. But if you've tried it and it works then that's probably ok?

SachaG:
> If you want, we could use Meteor.defer on the server and regular promises on the client just to be sure.

Yeah I think that's a safest way for now. Maybe add a `// TODO: find out if we can safely use promises on the server, too` comment in there and come back to it later.

ochicf:
Hi @SachaG! Sorry, I totally forgot about this.

Added a commit in order to use `Meteor.defer` as we discussed. Note that a Promise is being returned that resolves an array with every callback return: this is done to allow callbacks to optionally return a Promise themselves so the invoker of the hook can await them.

SachaG:
I forgot about this too… I think if you've tested this code we can go ahead and merge it, but just a note that routing will change a lot with the next update since we're moving to React Router 4.

ochicf:
Sure @SachaG, I'm using this in production with a quite complex app and it seems to work fine.

Anyway, the changes in those PR are completely opt-in:

- async callbacks can now return a promise so the caller code waits for them (so for example, you can run callbacks in paralel and the caller code will be waiting). If they don't return anything, then the callbacks will run in a detached way (caller code will continue without waiting for them, as it was until now)
- async router hook can be used to execute parallel computations when a route has changed (and having access to the route's props)

With that said, having this merged would be great, since it forces me and my team to have a local Vulcan versions with this code, which makes setup and mantainability a lot harder.

SachaG:
Sounds good, I'll merge it then!

Apollinaire:
Merged to `master` :/ I'm merging master into devel; 
Is there a way to avoid PRs to master automatically?

ochicf:
Sorry, my mistake... I did not see that I was creating the PR into master :\

Issue #2066, opened 2018-09-13
Datatable results count is incorrect when checkAccess is set


eric-burel:
Hi,

Currently the DataTable loadMore button is displayed when `count < totalCount` where count is the number of item retrieved, and totalCount the number of item in the database.

This method is incorrect when a `checkAccess` method is defined. The correct comparison would be `count + filteredOut < totalCount`, where `filteredOut` is the number of results eliminated when you call `checkAccess` (or `beforeFilterCount < total` where `beforeFilterCount` is the number of results before filtering).

For example, if you have 10 items but user can see only 8, all data are loaded when 8 results are printed and 2 results are filtered out. But the comparison is `8 < 10`, so the user thinks there are 2 more results to load while we already know that he actually can't access them.

You can get this count easily in the `default_resolver` and return it alongside the total. However I don't manage to add new fields into the `withMulti` hoc results in addition to `data.totalCount`.

SachaG:
If I understand correctly, I think this is fundamentally unfixable because it would require calling `checkAccess` on every document in the database… I think the best we can do is add an option to disable the document count altogether when you need to.

eric-burel:
You can't compute it in the db, however you have access of the number of filtered document when you call `checkAccess`. 
So basically you can't compute the total of accessible documents, but you can still remember the total of document your refused to the user, and send it to him. 
Hopefully that's enough for a loadMore button, because you know you are done the total of document you have + the total you filtered is equal to the total of document in the db. 
You don't need to filter all the document at once, you only filter the documents retrieved by the user, and you can still have a correct behaviour.

For pagination, it is true that you can't take the number of accessible documents into account when you compute the number of pages. So you may require 10 document per page but only get 4 (or zero) after filtering, and indeed there is not solution to this because you can't filter documents.

To sum it up the use of a custom JS function for document level permission management will never be fully compatible with pagination, but it could be compatible with a `loadMore`pattern.

meteorplus:
@eric-burel can you show me an example file where you use the checkAccess ?

thanks

eric-burel:
Hi @meteorplus,

[See this link](https://github.com/lbke/vulcan-validation-workflows/blob/master/lib/modules/setupDocumentValidation.js#L60-L78). This is an example function that make unvalidated data unavailable, except if you actually created those data. It works on any collection.

[...]

PR #2067, opened 2018-09-13
[WIP] trying to fix pagination/load more when `checkAccess` is defined

eric-burel:
See #2066 

I almost figured out how to add a `beforeAccessFilterCount` variable with the results. This count will allow to fix pagination/load more count (right now it does not consider filtered document, so "load more" will appear even when all data are loaded, with no way to fix it).


meteorplus:
@eric-burel why totalCount is not enough?

eric-burel:
Still relevant

PR #2068, opened 2018-09-13
Minor bug fixes

ErikDakoda:
- Fixed bug in single resolver (slug is undefined)
- Fixed bug in Utils.pluralize (pluralizing words ending in 's', for example address => addresses)


SachaG:
Thanks :)

PR #2069, opened 2018-09-13
[BugFix] remove `document` from `canCreateField`

Apollinaire:
When you run Users.CanCreateField with a function in `canCreate` field prop, it's trying to call a `document` arg that is undefined, because obviously there is no initial document when you try to create a document. So now it's only calling canCreate(user) as defined in the docs. I should have spotted this in #2006, sorry!


SachaG:
:+1:

PR #2070, opened 2018-09-15
fix bug preflight from apollo

meteorplus:
resetStore: Store reset while query was in flight.

https://github.com/apollographql/apollo-client/issues/2919#issuecomment-378790109


SachaG:
Oh nice! Glad we can finally get rid of that error!

meteorplus:
you're welcome 🥇 now pay me in crypto :P

SachaG:
I could give you half of all the Bitcoin I own, would that work?

meteorplus:
@SachaG that's very generous! Deal!

SachaG:
OK, here's zero Bitcoins :)

PR #2072, opened 2018-09-15
Pass locale as key to IntlProvider to force a rerender on locale change

ochicf:
See docs: https://github.com/yahoo/react-intl/wiki/Components#dynamic-language-selection

The main problem is that optimization with `shouldComponentUpdate` prevents nested components from updating through context, hence all components below a component that does not rerender won't rerender on locale change.


SachaG:
Thanks!

PR #2073, opened 2018-09-15
New SmartForm groupComponent property and small refactorings

ErikDakoda:
- Added new SmartForm property "groupComponent", which allows switching the layout of FormGroups (eg: line separator, box separator, tabbed interface)
- Refactored FormNestedArray to allow better customization of material-ui version; you can now specify the layout of the array header, array footer, and a divider between items
- Moved all references to bootstrap from vulcan-forms to vulcan-ui-bootstrap


SachaG:
> Added new SmartForm property "groupComponent", which allows switching the layout of FormGroups (eg: line separator, box separator, tabbed interface)

Shouldn't that be done with `replaceComponent`? If the idea is that you want to be able to customize the components of a specific form without changing them for *all* forms then I think we should implement a more general solution as its own PR. Maybe passing a `components` prop and then having `components.FormGroup`, `components.FormSubmit`, etc.

SachaG:
Otherwise the other changes seem good to me :)

ErikDakoda:
That sounds like a fine idea to me. I’ll make it so. BTW what’s the best way to split a pull request in two or remove some changes or commits from one? I still find git a bit confusing though I’m starting to get the hang of it.

SachaG:
Ideally I think you would just revert the commits locally and then make another, separate PR.

PR #2074, opened 2018-09-19
Add method to easily add middleware to collection resolved fields

ochicf:
This PR pretends to add a minimal and simple way to add middleware to existing field resolvers, based on the behavior described in [prisma/grapqh-middleware](https://github.com/prisma/graphql-middleware).

I've added a method to `Collection.prototype` named `addFieldMiddleware` that has a similar signature than `addField`.

### Use case

Add a middleware to an existing and specific (not all) field resolver.

### Without the PR

It can be done like this (imagine we want to add a middleware for `userId` field, that resolves to `user`):

```js
import collection from './path/to/collection';

// manually retrieve resolver
const resolver = collection.userId.resolveAs.resolver;

// manually set middleware
collection.userId.resolveAs.resolver = (document, arguments, ...args) => {
  // actual middleware code
  console.log('resolving user');
  return resolver(document, arguments, ...args);
};
```

Cons:
- code repetition
- manually retrieve resolver: could break if field or resolver does not exist (not tolerant to changes over time)
- middleware is applied statically: resolver is extracted and overwritten by a new function that calls it

### With the PR

```js
import collection from './path/to/collection';

collection.addFieldMiddleware({
  fieldName: 'userId',
  fieldMiddleware: (resolver, document, arguments, ...args) => {
    // actual middleware code
    console.log('resolving user');
    return resolver(document, arguments, ...args);
  },
});
```

Pros:
- more readable code
- middleware is a pure function
- won't fail if field or resolver does not exist
- abstracts how middleware is applied (more on this below)

### How middlewares are applied

Right now the implementation of `addFieldMiddleware` just replaces the resolver with a new one that applies the received middleware function. I think this is the more straight forward and simplest approach.

Another more complex approach would be to NOT replace the existing resolver and add the middleware to an array, and then apply them later. This would be more extensible, because it would allow to remove or reorder middleware functions. Of course it is much more complex and I don't know if it is necessary, so I think there is no need to implement it now. The good thing is that changing this in the future won't break any middleware addition because how they are applied is enclosed inside this function.

### DIFF
--- a/packages/vulcan-lib/lib/modules/collections.js
+++ b/packages/vulcan-lib/lib/modules/collections.js
@@ -1,5 +1,6 @@
 import { Mongo } from 'meteor/mongo';
 import SimpleSchema from 'simpl-schema';
+import get from 'lodash/get';
 import { addGraphQLCollection, addToGraphQLContext } from './graphql.js';
 import { Utils } from './utils.js';
 import { runCallbacks, runCallbacksAsync } from './callbacks.js';
@@ -72,6 +73,43 @@ Mongo.Collection.prototype.removeField = function (fieldName) {
 
   // add field schema to collection schema
   collection.attachSchema(new SimpleSchema(schema));
+};
+
+/**
+ * @summary Add middleware to a resolved field (or an array of middlewares) to a schema.
+ * @param {Object|Object[]} fieldOrFieldArray
+ */
+Mongo.Collection.prototype.addFieldMiddleware = function (fieldOrFieldArray) {
+
+  const collection = this;
+  const schema = collection.simpleSchema()._schema;
+
+  const fieldMiddlewares = Array.isArray(fieldOrFieldArray) ? fieldOrFieldArray : [fieldOrFieldArray];
+
+  collection.attachSchema(
+    fieldMiddlewares.reduce(
+      (newSchema, { fieldName, fieldMiddleware, middleware = fieldMiddleware }) => {
+        const resolver = get(schema, `${fieldName}.resolveAs.resolver`);
+        if (resolver) {
+          return {
+            ...newSchema,
+            [fieldName]: {
+              resolveAs: {
+                resolver: (...args) => middleware(resolver, ...args),
+              },
+            },
+          };
+        }
+        console.warn(
+          'Trying to add a middleware for the non-existing or non-resolved' +
+          `field ${fieldName} in ${collection.options.collectionName} ` +
+          'collection.',
+        );
+        return newSchema;
+      },
+      {},
+    ),
+  );
 };
 
 /**


SachaG:
Interesting. But I wonder how often you need to extend already existing resolvers. In the current state of things, beyond the `Users` collection which is part of core I can't see that many use cases? 

Also if we did add this, what about providing a function that takes in the resolver result as argument instead of the actual resolver function? Wouldn't it be better if you didn't need to call the resolver yourself inside the middleware function?

Finally another way we could implement something like this would be with callbacks. For example you could imagine a `post.user.read` callback that triggers every time that field is requested. That seems like a more "Vulcan-y" way to do it to me? It also doesn't require adding any new functions to Vulcan's API.

ochicf:
Hi @SachaG,

Thanks for the questions and comments, I'll try to respond as clearly as I can.

-----------

>Also if we did add this, what about providing a function that takes in the resolver result as argument instead of the actual resolver function? Wouldn't it be better if you didn't need to call the resolver yourself inside the middleware function?

I've used the approach of the package I linked above. My guess is that by having the resolver instead of the result allows you to call it with the parameters you want, so you have more control over it. Also, it gives you the ability to do something before the original resolver is called.

--------------

>Finally another way we could implement something like this would be with callbacks.

Yes, but by using callbacks you only have to behave as a reducer, so you only can modify the results. This is kind of related to the previous point.

If I imagine the execution of callbacks I see them as more horizontal, while middleware has some depth. In fact, from the [prisma/graphql-middleware docs](https://github.com/prisma/graphql-middleware#how-does-it-work):

>Execution of the middleware and resolver functions follow the "onion"-principle, meaning each middleware function adds a layer before and after the actual resolver invocation.
>graphql-middleware

-------------

>Interesting. But I wonder how often you need to extend already existing resolvers. In the current state of things, beyond the Users collection which is part of core I can't see that many use cases?

I was just trying to give an example. This would just be useful for anyone that wants to wrap/extend/limit/modify any resolver. It basically applies the open/close principle on all existing resolvers.

#### Example: restrict viewable fields

Now bear with me on this one, because it may seem a bit complex or over-engineered.

Every resolver that returns a document will probably (should) use `Users.restrictViewableFields` on the resolved document. Something like this:

```js
import Users from 'meteor/vulcan:users';
import { Categories } from './categories'; 

const schema = {
  categoryId: {
    // ... field options
    resolveAs: {
      fieldName: 'category',
      resolver: async (document, arguments, { currentUser }) => {
        const category = document.categoryId
          ? await Categories.loader.load(document.categoryId)
          : null;
          
          return category && Users.restrictViewableFields(currentUser, Categories, category);
      },
  },
};
```

Now to avoid that code replication and also to prevent ourselves from forgetting to add that restriction and filtering sensitive data, a package/utility could be implemented with the following purpose: always limit the viewable fields of the documents returned by a resolver.

With middlewares you could do that: iterate over every collection, take every field that resolves a typename registered by `createCollection`, and then wrap the resolver with a middleware that will apply `Users.restrictViewableFields` on the resolved document.

The described code above would look something like this:

```js
import { Collections } from 'meteor/vulcan:core'; 

// this function looks for fields that resolve a document or documents of type names
// created by `createCollection`. It also returns that collection
import { getFieldsThatResolveADocument } from './magic';

Collections.forEach(collection => {
  collection.addFieldMiddleware(
    getFieldsThatResolveADocument(collection).map(({ fieldName, resolvedCollection }) => ({
      fieldName,
      fieldMiddleware: async (resolve, document, args, context, info) => {
        const resolvedDocument = await resolve(document, args, context, info);
        return document && Users.restrictViewableFields(
          context.currentUser,
          resolvedCollection,
          resolvedDocument,
        );
      },
    }),
  );
});
```

Note that you will also have an extra gain: all the resolvers would need only to provide the document, so they would be simpler as they could resolve the document directly:

```js
import Users from 'meteor/vulcan:users';
import { Categories } from './categories'; 

const schema = {
  categoryId: {
    // ... field options
    resolveAs: {
      fieldName: 'category',
      resolver: document =>
        document.categoryId && Categories.loader.load(document.categoryId),
  },
};
```

------------

By the way and as a side note, I always doubt whether I should open a PR with new ideas or implementations like this or I should better open an issue in order to discuss it before.

SachaG:
> My guess is that by having the resolver instead of the result allows you to call it with the parameters you want, so you have more control over it. Also, it gives you the ability to do something before the original resolver is called.

That's true. My only worry is that we don't really have anything else that works like this in Vulcan (when using callbacks you're always taking in an already-computed input, you don't need to call a function yourself), so we're introducing a new pattern. 

> With middlewares you could do that: iterate over every collection, take every field that resolves a typename registered by createCollection, and then wrap the resolver with a middleware that will apply Users.restrictViewableFields on the resolved document.

That's a great example. I've thought before about automating `Users.restrictViewableFields` in some way. Although it would probably make more sense to include it in the `Collection.loader.load` call – but that's another discussion. 

> By the way and as a side note, I always doubt whether I should open a PR with new ideas or implementations like this or I should better open an issue in order to discuss it before.

I think generally it's safer to discuss it first, but for simple features like this I guess having the actual code available to look at can only help :)

---

Overall my feedback would be: this is a cool idea, but

1) I'm worried about adding a "new" pattern.
2) It seems like a very niche use case.
3) There are already workarounds to achieve the same thing.

I would like to get other people's input (@Apollinaire @eric-burel @ErikDakoda ?) before I decide if we should approve/reject/modify this. 

Maybe a good middle ground would be implementing this with callbacks like we discussed, you'd lose some of the flexibility but it wouldn't require increasing Vulcan's API surface.

eric-burel:
Hi,
It sounds like something that could easily be added as a 3rd party package? 1. and 2. worries from Sacha are more about including it in the core or not. Since this feature does not seem to require changes in the core, I like the pattern of having a 3rd party package for it, and then if you notice a strong usage we can simply copy-paste it to the core.

Small remark: if it replace the resolver, then you can call it `replaceFieldResolver`  and rename the `resolver` argument as `currentResolver` for example, it sounds clearer to me.

About 3., are the patterns strictly equivalent? In terms of available arguments for example? E.g do callbacks have access to the context? If not then to me its ok to have another pattern, if it's a niche pattern then we can simply keep it out of the core for the moment.

ochicf:
@eric-burel thanks for your comment.

About this feature not being part of core, I was actually thinking the same, so I'll create an external package. Maybe I'll try to implement the `restrictViewableFields` as a use case example.

>Small remark: if it replace the resolver, then you can call it replaceFieldResolver and rename the resolver argument as currentResolver for example, it sounds clearer to me.

Mmmh you have a point there, so either:
- rename function: do you think `applyFieldMiddleware`, `applyResolverMiddleware` or something like that makes sense? I like idea of using the `middleware` concept to keep consistency with the package stated above
- defer replacement of the resolver until the schema is generated, so middleware functions will be actually added to a list (hence the `addFieldMiddleware` name is more accurate)

>About 3., are the patterns strictly equivalent? In terms of available arguments for example? E.g do callbacks have access to the context? If not then to me its ok to have another pattern, if it's a niche pattern then we can simply keep it out of the core for the moment.

I think the main differences are:
- the way both pattern execute the consumer code: middleware functions need to call the NEXT one -having control over it-, while Vulcan callbacks don't -having less control but being more easy/direct to use
- the main input of that function: being the resolver itself for the middleware function, and the resolver's result for Vulcan callbacks

If the Vulcan callbacks were called with all of the resolver arguments `(document, args, context, info)`, then all of them would receive them, so that is not a problem.

eric-burel:
It makes sense, you could use it for fine tuned caching for example. Eg you call the original resolver only in case of cache miss, because it may be costly (say the result of a geospatial request on another collection whose documents change a lot over time), and you don't want to/cannot tweak the original schema. You may want to change the context to replace a collection by a legacy one before resolving (good luck finding a real life example lol but why not).

Those are definitely advanced use cases but why not, as Vulcan apps grow bigger tricky optimization issues will arise. `applyFieldMiddleware` sounds ok to me.

Apollinaire:
I like the idea of adding middleware to the resolvers! However, I still have a few questions : 
- Is this strictly restricted to the fields where you set `resolveAs`? If so, how could we have a united way of applying middleware to all our resolvers, including the ones we give in `createCollection`? Or even the mutations ? Remember that in graphQL a resolver is simply an answer to a query, so you can use prisma/graphql-middleware for mutating data too.
- It seems that this could indeed be done with callback hooks, like we already do in mutations, but it would mean we loose the 'onion' structure of graphql-middleware that I find better.

IMO it's not a good idea to have the two patterns included in Vulcan core (callback hooks and middleware) because they achieve the same goal. Also, the middleware pattern is a bit hard to grasp, so explaining both would probably be confusing. 

BTW we can have context in callback hooks if they are called from within the field resolver. Context is already passed to the callback hooks of mutations too.

Here is what I suggest :
- If we want to include this in the core, we should rethink the whole callback hooks use in the resolvers and mutations, and replace it with middleware. Give a library of default middleware (for these [use cases](https://github.com/prisma/graphql-middleware#graphql-middleware-use-cases) and some other vulcan-specific) and maybe rewrite the `getDefaultResolvers` and `getDefaultMutations` functions to use it? That would be a big rework of how we're handling data, but is relevant regarding the late openCRUD change, because it would mean we are getting closer to the 'usual' graphQL way of doing things.
- On the pure application thing, I don't mind replacing the resolver. It's similar to composing hocs I guess? And it makes things simpler so i'm all for it.


Since we're talking about the `resolveAs` in the schema, I never really liked it because I find it unsafe. I've always thought it would be great to be able to just specify a relation field : basically, instead of defining a new resolver for your `categoryId`, you just define that this field is an id or slug or whatever, and it relates to a certain type, and then you use the `single` resolver for that type.

It can already be done like this but it's heavy to setup : 
```js
resolveAs: {
      fieldName: 'category',
      resolver: async (document, arguments, context) => {
         const query = `
         query GetCategory ($_id : String) {
             category(_id: $_id) {
               ...fragment
         }
         `
         return await runGraphQL(query, { _id: document.categoryId }, context)
     },
  },
```

SachaG:
> Since we're talking about the resolveAs in the schema, I never really liked it because I find it unsafe. I've always thought it would be great to be able to just specify a relation field : basically, instead of defining a new resolver for your categoryId, you just define that this field is an id or slug or whatever, and it relates to a certain type, and then you use the single resolver for that type.

Yes, I think that's the natural evolution. Adding a way to define `hasOne`, `hasMany`, etc. relationship presets that use `resolveAs` behind the scenes but don't require you to manually define it. But that's a different matter I think. 

@ochicf how about we keep this as an external package for now and if this turns out to be a need that pops up a lot, we'll think about merging it into core again?

ochicf:
@SachaG perfect.

I'll write and publish the package and I'll let you know when it is done. I'm closing this PR.

Thank you everyone!

PR #2075, opened 2018-09-19
Users' slug is updated on displayname change

Apollinaire:
- The slug is updated when the displayname changes
- The `Utils.getUnusedSlug` helper function does not return empty slugs anymore (it's '0' if the `slug` argument is falsy)

### DIFF
--- a/packages/vulcan-lib/lib/modules/utils.js
+++ b/packages/vulcan-lib/lib/modules/utils.js
@@ -169,11 +169,7 @@ Utils.slugify = function (s) {
   return slug;
 };
 
-Utils.getUnusedSlug = function (collection, unSureSlug) {
-  
-  //if the unSureSlug is falsy, replace it with '0'. This avoids the creation of an empty slug, which is bad in the urls.
-  const slug = unSureSlug ? unSureSlug : '0';
-  
+Utils.getUnusedSlug = function (collection, slug) {
   let suffix = '';
   let index = 0;
 
--- a/packages/vulcan-lib/lib/modules/utils.js
+++ b/packages/vulcan-lib/lib/modules/utils.js
@@ -169,19 +169,6 @@ Utils.slugify = function (s) {
   return slug;
 };
 
-Utils.getUnusedSlug = function (collection, slug) {
-  let suffix = '';
-  let index = 0;
-
-  // test if slug is already in use
-  while (!!collection.findOne({slug: slug+suffix})) {
-    index++;
-    suffix = '-'+index;
-  }
-
-  return slug+suffix;
-};
-
 Utils.getUnusedSlugByCollectionName = function (collectionName, slug) {
   return Utils.getUnusedSlug(getCollection(collectionName), slug);
 };
--- a/packages/vulcan-lib/lib/modules/utils.js
+++ b/packages/vulcan-lib/lib/modules/utils.js
@@ -169,7 +169,11 @@ Utils.slugify = function (s) {
   return slug;
 };
 
-Utils.getUnusedSlug = function (collection, slug) {
+Utils.getUnusedSlug = function (collection, unSureSlug) {
+  
+  //if the unSureSlug is falsy, replace it with '0'. This avoids the creation of an empty slug, which is bad in the urls.
+  const slug = unSureSlug ? unSureSlug : '0';
+  
   let suffix = '';
   let index = 0;
 
--- a/packages/vulcan-lib/lib/server/utils.js
+++ b/packages/vulcan-lib/lib/server/utils.js
@@ -1,5 +1,5 @@
 import sanitizeHtml from 'sanitize-html';
-
+import { Connectors } from './connectors';
 import { Utils } from '../modules';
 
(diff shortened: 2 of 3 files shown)


SachaG:
You added an `onUpdate` for the `slug` field but the slug field is not editable? So I don't think this would ever trigger? Or maybe I'm missing something?

Also making it easier to change the slug can be a problem because it will break old links (there is no forwarding system to keep old slugs working currently). So we might want to think about that.

Apollinaire:
> You added an onUpdate for the slug field but the slug field is not editable? So I don't think this would ever trigger? Or maybe I'm missing something?

I did that just as a safety if someone wants to edit the users slug. Indeed that part won't trigger with the current definition of the schema, but I thought that would be nice to have this additional security in case they want to make it `canUpdate`.

About the old links, I agree that it's an issue. Turning slugs to an array containing the history of slugs could work but it might be a perf hit to the resolution of a query (I don't know how indexing in Mongo works, and if it works with the items of an array)

SachaG:
I think that makes things a little too confusing… Let's keep it simple :)

SachaG:
Also I don't get the `0` thing. Why would you want to set a slug to that?

Apollinaire:
> Also I don't get the 0 thing. Why would you want to set a slug to that?

The `0` is like the default value of the slug. You need a slug to be non-empty, so I picked `0` because it seemed like the most appropriate, but it could be any alpha-numerical character.

Without this PR, if you use `Utils.getUnusedSlug` with an empty or null String, or with only special characters (load example-simple, and sign up with `;;;;;;;;;;;;;;` as a username) the slug will be empty, then if you want to render a route `user/:slug` then for this user you won't be able to access it because his slug will be an empty string, so you'll just end up going to the route `user`. 

I think I should remove the `onUpdate` part of this PR because I won't take the time to make a good system for old slugs and it would obviously complicate stuff, but we should keep the safety addition to `Utils.getUnusedSlug` for the reason I explained above

SachaG:
But then the next user who signs up with `;;;;;;;;;;;;;;` as a username will also have `0` as a slug? That doesn't seem like it would work? If the goal is to avoid bad slugs then we should probably generate a random string or number or use the user's `_id` as a fallback?

I'm not sure this should be added to this function though, I think that whatever is calling `getUnusedSlug` should test to make sure the return value is not `null` and then handle that case instead.

Apollinaire:
The next ones will receive `0-0`, `0-1`, ..., `0-n` as slugs, since all I do is check the input and replace it with `0` if it's falsy, so the suffix is appended to it as it is with any other slug. Generating a random string would also work yes. 
Or we could leave people the choice of a fallback ? add an argument at the end to make people be able to provide a fallback (that most likely will be the _id of the object you're slugging). I actually think this is the less opinionated solution, and it's better than the `0` by default.

SachaG:
I think I would do something like

```
const displayName = createDisplayName(user);
const basicSlug = Utils.slugify(displayName);
return basicSlug ? Utils.getUnusedSlugByCollectionName('Users', basicSlug) : getRandomSlug();
```

SachaG:
Or even just:

```
return basicSlug ? Utils.getUnusedSlugByCollectionName('Users', basicSlug) : user._id;
```

Apollinaire:
> return basicSlug ? Utils.getUnusedSlugByCollectionName('Users', basicSlug) : user._id;

I agree about this for the `Users` collection. However, what about the function `getUnusedSlug` ? Should it ever return an empty string as a slug? We can't just raise a warning on this because it's going to be lost in the logs.

SachaG:
I don't think `getUnusedSlug`'s job should be to make up a slug. It should just be to see if the slug you give it is already in use, and if not add a suffix. Also we shouldn't use `findOne` but that's another matter…

Apollinaire:
What's about `findOne` ? Is it about performance ? 
Could this be better? 
```js
  // test if slug is already in use
  const regex = new RegExp('^' + slug + '-[0-9]+$');
  const results = collection.find({ slug: { $in: [slug, regex] } }, { fields: { slug: 1, _id: 0 } }).fetch();
  const usedSlugs = results.map(item => item.slug);

  while (usedSlugs.indexOf(slug + suffix) !== -1) {
    index++;
    suffix = '-' + index;
  }
  return slug + suffix;
```
regex explanation : https://regex101.com/r/4ECNt7/3
The only thing I don't really like here is that you run `indexOf`  so the complexity is O(n<sup>2</sup>), but I guess it's already better than calling the db each time in the loop.

SachaG:
Yeah getting rid of the loop would be a good start. But more than that we should just generally not use `findOne` in random helpers like that, I feel like it makes things hard to debug since you might not realize where that extra database call is coming from. In this case though I don't know if there's a way around it… Also `Connectors.find` would be better than `find` since it's (in theory) more database-agnostic.

Apollinaire:
I can't do `import { Connectors } from '../server/connectors'` because we are in modules so this breaks the client. This function belongs to the server only, so I moved it to `lib/server/utils.js` and it's now using the connectors.

PR #2077, opened 2018-09-20
SmartForm.getLabel() intl string fallback

ErikDakoda:
SmartForm.getLabel() now falls back to a "global" namespace when a collection-specific intl string is not found; then it falls back to using no namespace.

This is implemented to reduce duplication of translations, especially for field labels common to multiple collections.


ErikDakoda:
@SachaG do you mind accepting this? I have a bunch more code to commit - and I haven't figured out how to have multiple pull requests at the same time.

Apollinaire:
> I haven't figured out how to have multiple pull requests at the same time.

One branch = one PR

PR #2079, opened 2018-09-22
Override SmartForm Components

ErikDakoda:
- You can now override which SmartForm components to use for a given form; currently `FormGroup`, `FormErrors` & `FormSubmit` can be overridden - more to come
- The default components to use can be chosen in settings
- `submitLabel`, `cancelLabel`, and `revertLabel` now take a node (which is backwards compatible with string) so that you can include a FormattedMessage, an icon, etc.

### DIFF
--- a/packages/vulcan-forms/lib/components/Form.jsx
+++ b/packages/vulcan-forms/lib/components/Form.jsx
@@ -31,6 +31,7 @@ import {
   getSetting,
   Utils,
   isIntlField,
+  getComponent,
 } from 'meteor/vulcan:core';
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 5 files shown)


ErikDakoda:
@SachaG Here is the other half of the split pull request from before

SachaG:
> The default components to use can be chosen in settings

I don't think we need that since you can already replace the default components with your own (current behavior) if you want to replace *all* form components?

This change might be a little tricky so maybe it's better if I handle it? I feel like it requires putting in more thought into what the best API would be.

SachaG:
OK, let me know what you think. I'll close this for now, but maybe you can make another PR for the label thing?

ErikDakoda:
vulcan-material-ui ships with multiple FormGroup components: One that puts each group in a material ui "paper", one that just separates the group with a horizontal line, and one for a special use case with no group decoration. This allows the user of the package to simply select which look they want to use by default, and then override it on a per-form basis if they want. This is necessary for my use case, so it leaves me in a bind if you reject the PR.

SachaG:
You could still have the same settings in your own package and then call `replaceComponent` depending on which setting the user chose?

PR #2091, opened 2018-09-28
Add /i18n debug page

Apollinaire:
I created a debug page for i18n strings. The route for it is `/i18n`
The features are : 
- Tells your current locale
- Allows you to switch locale based on the registered locales
- Shows all the strings registered for every locale and their related Ids in a `Components.datatable`

Tell me if there are any additional features you can imagine and would like to be added. 

If I can add a proposition about the debug package : all the routes should be renamed to `/debug/i18n` or `debugi18n`, or `i18ndebug` (which is my favorite) to avoid conflicts with the routes set for the app. I don't think anyone will set a route called `/i18n`, but a `/settings` route is pretty common in any app. So if someone was to add it to their app, they would be unable to see the settings debug page. @SachaG if you like the idea I'll add this to the PR.

If this gets merged I'll update the docs

### DIFF
--- a/packages/vulcan-debug/lib/components/I18n.jsx
+++ b/packages/vulcan-debug/lib/components/I18n.jsx
@@ -1,0 +1,70 @@
+import React from 'react';
+import { registerComponent, Components, Strings, Locales } from 'meteor/vulcan:lib';
+import PropTypes from 'prop-types';
+import sortedUniq from 'lodash/sortedUniq';
+
+/**
+ * Internationalization debugging page
+ *
+ *
+ **/
+function LocaleSwitcher(props, context) {
+  return (
+    <div>
+      <span>Switch locales :</span>
+      {Locales.map(localeObj => (
+        <Components.Button key={localeObj.id} onClick={() => context.setLocale(localeObj.id)}>
+          {localeObj.label}
+        </Components.Button>
+      ))}
+    </div>
+  );
+}
+LocaleSwitcher.contextTypes = {
+  getLocale: PropTypes.func,
+  setLocale: PropTypes.func,
+};
+
+export const I18n = (props, context) => {
+  // translations holds all the translations ids
+  let translations = [];
+  let columns = [
+    {
+      name: 'id',
+      component: function({ document }) {
+        return document;
+      },
+    },
+  ];
+
+  // reunite all the ids in a single array (translations) and create the columns for each language
+  Object.keys(Strings).forEach(language => {
+    translations.push(...Object.keys(Strings[language]));
+    columns.push({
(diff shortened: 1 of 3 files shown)


SachaG:
This is great! Will be very useful for my projects, too :) And I think `/debug/i18n` would be best, this way `/debug` can link to all the debug dashboars.

Apollinaire:
Something like `addRedirectRoute({ name: 'redirectionSettings`, from: '/settings', to: '/debug/settings' })` and when you go to `/settings` it explicitely changes the url to `/debug/settings` ?

Apollinaire:
I'll try something : https://github.com/ReactTraining/react-router/blob/v3/docs/guides/RouteConfiguration.md#configuration-with-plain-routes

SachaG:
Yes that's what I had in mind. But maybe we can still use `addRoute`, just have a `redirect: '/debug/settings/` option

Apollinaire:
That makes sense. 
I tried doing this regarding the example in react-router's docs (link above):
```js
addRoute({ name: 'redirection', path: '/settings', onEnter: (nextState, replace, callback) => replace(`/debug/settings`)})
```
Because the way things are done, any property you add to the argument of `addRoute` is kept in the route object. The argument passes along fine and you can see it in the state of the `<Router >` component, but it does not work when navigating to the route. Any ideas about why it's not working ?
Also, maybe it's not worth doing this now since @eric-burel is working on integrating rr4. 

To come back to the changes I made in this PR: I made all the routes "side by side", but actually they should be parent/children right ? I could refactor maybe. I'm not using this personally and I don't really understand the point of having parent/child relation in a route, but I guess this could be a good example of implementation for this feature.

SachaG:
Yeah let's wait until RR4 and open a new issue to discuss this feature. As for parent/child routes I also don't really get it, so I think we can leave it for now.

eric-burel:
Hi, 
You can already take a look at the `apollo2` branch for routing, it should be working.

Usually parent child relationship usually means nesting levels of layout. Eg you have your app layout, and a children is displayed depending on the route. Then, this children could also display smth depending on the route. 
For example if you wanted to create a debug home page with a submenu it would be better to nest routes. In terms of React lifecycle I think the difference is that you provoke more mounts/unmounts than necessary when you don't use nested routes.
I think currently having a central route registry is not compatible with parent/children routes, because you have to include a `Switch` in your own component to do this. Instead you would register the "home" route of your package and then provide your own routing internally.

It will indeed need further discussion, I am not really familiar with RR4 (nor fond of it...) and it is quite complex. I'll check the server part of Apollo 2 soon.

PR #2092, opened 2018-09-30
SmartForm: use prop `schema`, if given

dimofte:
Prop `schema` exists in Form's `propTypes` but it wasn't used, the component was always using the schema attached to prop `collection`.

### DIFF
--- a/packages/vulcan-forms/lib/components/Form.jsx
+++ b/packages/vulcan-forms/lib/components/Form.jsx
@@ -33,6 +33,7 @@ import {
   isIntlField,
 } from 'meteor/vulcan:core';
 import React, { Component } from 'react';
+import SimpleSchema from 'simpl-schema';
 import PropTypes from 'prop-types';
 import { intlShape } from 'meteor/vulcan:i18n';
 import Formsy from 'formsy-react';
@@ -74,7 +75,7 @@ const getDefaultValues = convertedSchema => {
 
 const getInitialStateFromProps = (nextProps) => {
   const collection = nextProps.collection || getCollection(nextProps.collectionName);
-  const schema = collection.simpleSchema();
+  const schema = nextProps.schema ? new SimpleSchema(nextProps.schema) : collection.simpleSchema();
   const convertedSchema = convertSchema(schema);
   const formType = nextProps.document ? 'edit' : 'new';
   // for new document forms, add default values to initial document


dimofte:
this implementation relies on the existing faulty architecture of deriving state from props, I didn't want to refactor the whole component now

packages/vulcan-forms/lib/components/Form.jsx
```diff
@@ -168,7 +169,7 @@ class SmartForm extends Component {
   /*
 
   Get a list of all editable fields
-  
+
```
SachaG: It'd be great to only have the actual code changes in the PR if possible

packages/vulcan-forms/lib/components/Form.jsx
```diff
@@ -560,18 +561,29 @@ class SmartForm extends Component {
...
   // TODO: see https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html
-  
+
   */
   UNSAFE_componentWillReceiveProps(nextProps) {
     if (!isEqual(this.props, nextProps)) {
       this.setState(getInitialStateFromProps(nextProps));
     }
   }
+  componentWillReceiveProps(nextProps) {
```
SachaG: We probably don't want to use both `UNSAFE_componentWillReceiveProps` and `componentWillReceiveProps`?

Also wouldn't changing the schema already be covered by `!isEqual(this.props, nextProps)`?

packages/vulcan-forms/lib/components/Form.jsx
```diff
@@ -74,7 +75,7 @@ const getDefaultValues = convertedSchema => {
 
 const getInitialStateFromProps = (nextProps) => {
   const collection = nextProps.collection || getCollection(nextProps.collectionName);
-  const schema = collection.simpleSchema();
+  const schema = nextProps.schema ? new SimpleSchema(nextProps.schema) : collection.simpleSchema();
```
SachaG: Maybe the schema you pass should already be a SimpleSchema schema? It seems like it would simplify the component's logic.

SachaG:
Thanks! I think this is a good idea but I'm not sure about the implementation specifically.

packages/vulcan-forms/lib/components/Form.jsx
dimofte: done, sorry about the automatic linting

packages/vulcan-forms/lib/components/Form.jsx
dimofte: Wow, I somehow missed the big deprecation of lifecycle methods in React v16.3 😊 
Then I totally misread the code, I thought it's a creative way of disabling the method call, due to some reasons regarding the whole state recalculation. Of course deep comparing whole props is covering the schema case. I've removed the commit altogether

dimofte: However, note that this code does not work in the current `master` of Vulca-Starter, because both `react` and `react-dom` need to be at version ^16.3 (the reason why the state wasn't updated in my case: react-dom v16.2)
I opened a [pull request](https://github.com/VulcanJS/Vulcan-Starter/pull/96) in that repo

packages/vulcan-forms/lib/components/Form.jsx
dimofte: It would simplify this code, indeed, but it would be inconsistent with the way schemas are attached to the collections in Vulcan.
Unlike Meteor collections, where they are attached as `SimpleSchema` objects, in Vulcan they're passed as plain objects. 
So I think the API is more consistent if they stay plain objects everywhere

SachaG: Makes sense. Let's keep it like this then :)

PR #2094, opened 2018-10-02
[WIP] Apollo2 server

eric-burel:
I started the update to Apollo server v2. It is not functional but basic ideas are there.

Things to do:
- [X] Splitting the code to make it a bit clearer
- [X] First experiment with new patterns
- [X] Get a working version
- [X] Update imports everywhere needed, as Apollo-server exports a few common helpers (`makeExecutableSchema`, `gql` etc.)
- [X] Simplify the engine setup, it should now only need the correct options to work **-> the API have changed a lot, and I can't tell what setting meant what, so there might be regressions**
- [X] First test with GraphQL Playground: **we need to setup the Authorization header with `localStorage['Meteor.loginToken'], however GraphQL PlayGround (replacement to GraphiQL) does not posses a `passHeader` option as graphiql did**. It does work if you manually update the header with your token, but obvisouly this is tedious.
- [ ] Setup cookies
- [ ] Populate the `Authorization` header when using GraphQL PlayGround. See https://github.com/prisma/graphql-playground/issues/510 for global Headers (right now we have to create an initial tab to setup headers). The best solution seems to be using cookies. In GraphiQL, we used the `passHeader` option that allowed access to the localStorage, but it was awfully hacky. See https://github.com/prisma/graphql-playground/issues/849
- [x] Update starter package npm packages versions
- [X] Meteor integration (which is needed to actually serve the graphql endpoint) : **seems ok, we need to use meteor/WebApp to setup a 3rd party HTTP server (here our graphQL endpoint), and thus use `apollo-server-express` as well**
- [X] Take a look a t `meteor_patch.js`, is it still needed? **->Seems to work without it, I don't know which issue lead to using this however**
- [X] Investigate whether we should keep Express or replace its feature with other modules **-> right now this is the only way to connect with Meteor**
- [ ] Try merging with devel to catch up latest updates?
- [ ] SSR (will come later, needs both improvement on client and server)
- [ ] Better error management, right now they are caught by Apollo Server and displayed in the worst possible way
- [ ] Investigate using datasources instead of DataLoader to pass the collection to the resolvers (need help)

- It seems easier to get rid of Express, there are no example of Apollo + Meteor + WebApp and it is not trivial. However this means we must replace Express middlewares by smth else. I am investigating this.
- Resolvers should now accept a `dataSources` object in the context. We should access collections from this object instead of the bare context. We can maintain both the legacy behaviour (`context.MyCollection`) + the new syntax for the moment. However I don't know how we are expected to set this up or replace DataLoader. See https://www.apollographql.com/docs/apollo-server/v2/features/data-sources.html.
- Note that you now must separate options computation from context computation (only the context option can take the current request into consideration, which was already true in v1 despite a misleading API anyway).


SachaG:
Nice! The new `dataSources` feature is really nice but let's keep the legacy `context` implementation for now (unless setting up datasources is really trivial). Introducing caching into the mix will make it harder to know what's really going on, and even for Vulcan users in general I feel it's better if we introduce new concepts progressively instead of all at once in the same update.

eric-burel:
This is merged as I got a working solution, except for GraphQL Playground that we can still fix later.

PR #2098, opened 2018-10-05
Add graphql-voyager

Apollinaire:
available at `/voyager` in dev mode only
Disclaimer : I don't know if I'm breaking things here, since I don't know the advancement of the update to apollo v2. When I tested this, the client was not rendering at all because of some missing exports from `vulcan:core` (the ones 'redirected' from `vulcan:lib`), but the apollo server was running and also the `voyager`.

### DIFF
--- a/packages/vulcan-lib/lib/server/apollo-server/apollo_server2.js
+++ b/packages/vulcan-lib/lib/server/apollo-server/apollo_server2.js
@@ -19,6 +19,8 @@ import { ApolloServer } from 'apollo-server-express';
 //import { makeExecutableSchema } from 'graphql-tools';
 import { Meteor } from 'meteor/meteor';
 // import { Accounts } from 'meteor/accounts-base';
+import express from 'express';
+import { express as voyagerMiddleware } from 'graphql-voyager';
 
 import { GraphQLSchema } from '../../modules/graphql.js';
 import { WebApp } from 'meteor/webapp';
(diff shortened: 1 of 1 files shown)


Apollinaire:
ping @eric-burel

eric-burel:
Hi @Apollinaire, I've struggled a bit but got it working with a different pattern, thanks for your demo!

With your version, it does not find the route correctly :/ I can't really tell why, because theoretically it's a valid way to add an endpoint connected to the apollo server, and I think it is indeed correctly set server side.

Anyway take a look at the `apollo2` branch, it should work as expected. Note that there might be issues with versions, `react-apollo` 2.2 seems breaking, I forced `2.0.4` in the starter to get it working.

Can you please update the Starter packages if you get it working? My clone is not clean as I am testing stuffs. Let me know if there are more issues.

Also we will need to set up GraphiQL the same way, since we can't set GraphQL Playground headers yet (I may try to PR smth on their repo if I manage to)

Apollinaire:
Thanks for this ! I don't have the time rn to look into the starters but I will in a week or so. 
Closing this PR since the feature is done

eric-burel:
@Apollinaire Actually I got why it doesn't work, it's just that I misguided you with `applyMiddleware`. It actually just add middlewares to the `/graphQL` endpoint, so you pass it your Express server. Currently your endpoint was simply not declared known by Meteor, which returns a 404 instead of passing the request ot the server you instanced. Instead the only solution is to use `WebApp.connectHandlers.use(...)`.

Apollinaire:
what's weird is that I used `applyMiddleware` but I could reach the voyager at `/voyager` ? Anyway it makes more sense to pass it to meteor since it's the first "layer" of the server

PR #2102, opened 2018-10-09
Revert changes made to getUnusedSlug in #2075

Apollinaire:
In #2075 I made the mistake of thinking that `onCreate` functions were always exectued in async, which is [not the case for the `Users` collection](https://github.com/VulcanJS/Vulcan/blob/66b1aa467f30a0f3025f022f52bdac5981fbaed7/packages/vulcan-users/lib/server/on_create_user.js#L5). Thus, the slugs would be `[ Object object ]` because `getUnusedSlug` returned a pending promise.
So I reverted to the previous way we did things, using `collection.findOne`. I don't really understand why this is fine in terms of async, but at least it works. I let the version I implemented but it's commented, this way it's not lost and can be reused when we refactor or move out of meteor for example.

### DIFF
--- a/packages/vulcan-lib/lib/modules/utils.js
+++ b/packages/vulcan-lib/lib/modules/utils.js
@@ -168,6 +168,35 @@ Utils.slugify = function (s) {
 
   return slug;
 };
+Utils.getUnusedSlug = function (collection, slug) {
+  let suffix = "";
+  let index = 0;
+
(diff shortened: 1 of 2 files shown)


SachaG:
> I don't really understand why this is fine in terms of async

Because of Meteor magic (Fibers) ;)

PR #2106, opened 2018-10-19
Warn when no searchable field is set and terms.query is set

eric-burel:
Maybe there are use cases where you might want to use `terms.query` while not setting searchable to true?


SachaG:
Seems like a good idea. Btw please make separate commits for formatting changes if possible.

PR #2108, opened 2018-10-25
[WIP] Cleanup forms

eric-burel:
I've reviewed quickly the different Form components. 

Todo list:

- [X] Always use the same patterns for loading FormComponents, so that they are always defined. This allows to reuse some form subcomponents, eg NestedArrays
- [X] Defined "Layout" components, to make them truly replaceable. @Apollinaire and @ErikDakoda may be interested as it makes it easier to replace Bootstrap without rewriting all the logic of the component everytime. It also allows to change the layout locally, for example to display the "remove" icon of an array of inputs somewhere else, or to use Material UI `<Grid>` instead of div, etc. The same pattern could be applied to DataTable and other components.
Ideally you should only find `div` with bootstrap classes in those "Layout" components.

- [X] add a `changeCallback` to listen to SmartForm changes, for example to compute a price based on currently selected options

- [ ] add a `blurCallback`. There can be a lot of change events, so we could need an intermediate callback. Not sure how to implement it though.

- [ ] Split logic. `Form` and `FormComponents` have become...well... quite huge. I'd like to split them up a bit with a more functional logic. First, I think we could separate the "schema to fields" transformations (like `createField` method), because they are becoming quite complex with the support of arrays, arrays of items, etc. They could be handled in a pure JS file instead of belonging to a React component.

Then we can define containers for similar features. I think error management, current state management, submission management, fields computations, for example. They might be tied together so it's not trivial but I'm sure we can figure this out.

- [ ] Support custom input for arrays of objects. Right now the `input` prop is ignored (e.g `"address.$" :{ type: new SimpleSchema({...}), input: "ThisInputIsIgnored"}`). It needs small tweaks in the `schema_utils` file, and maybe to define a unified approach for `nestedFields` and `arrayField`.

- [ ] Support nested schema with identic keys. The schema flattening currently makes this impossible (eg if your "recipe" as a `name` field and a `ingredients` array, and your "ingredient" schema also have a `name` field, you will have trouble). Might be hard though.

- [ ] Review `FormComponentInner`, which belongs to the core. RIght now it's logic is duplicate in `vulcan:material-ui` and creates issue (can't override input of arrays)

- [ ] allow to pass a `document` prop instead of `documentId`. When passing a `document` prop, the form is filled as expected but edit fails, as there is no `updateWhatever` method. The `FormWrapper` seems to expect a `documentId` prop and ignore `document`. We can define `documentId` too but then the document is fetched again.


eric-burel:
@SachaG This should be mergeable. There seems to be a lot of changes but actually I mostly moved files around in the last commits to start drying the code. The biggest new feature is the `changeCalback`.

Let me know when you have found time to review it. It works in my apps and all tests are still passing, so that should be okay.

SachaG:
I won't have time to really review this for a while so I think you should go ahead and merge it, by now you know the forms code better than I do anyway :)

eric-burel:
Ok no problem, let me know if you notice any bug later on!

Issue #480, opened 2014-10-08
Password reset email sent from "Meteor Accounts"


nickdima:
The password reset email has as sender "Meteor Accounts" (no-reply@meteor.com).
I think it would be good to have it customised based on Telescope's settings, like the invite email is.

SachaG:
I need to take another look at this…

eric-burel:
Encountered this today, defaultEmail is not respected

PR #2110, opened 2018-10-26
set default email and siteName for Accounts related emails

Apollinaire:
Related to #480 
Meteor docs: https://docs.meteor.com/api/passwords.html#Accounts-emailTemplates
Set the email address for:
- resetPassword
- forgotPassword
- verifyEmail

Questions remaining: 
- Is it right to set the `siteName` from the `title` setting, and what is it used for ?
- Can `siteName` and `from` still be overwritten from other packages?


SachaG:
Nice improvement :)

PR #2112, opened 2018-10-29
single resolve documentId undefined -> now returns null

eric-burel:
Hi,

All in the title. @Apollinaire @SachaG let me know if it is okay. Here is how I intend to update the doc:

### Single Resolver

"The `single` resolver takes a `documentId` argument, and should return a single document. If no `documentId` is provided or the `id` does not match any existing document, it should either return `null` or throw an error depending on your use case.

If you need to get only one element but you don't know its id, for example to display the last article on your website, the preferred pattern is to use `withMulti` with a limit of 1."

Edit: I tend to have ugly merge commit, any idea how to avoid this?


SachaG:
Looks good to me!

Issue #2090, opened 2018-09-27
withMulti results become undefined on terms change


eric-burel:
Hi, all is in the title. I setup a Map component that rely on withMulti. The architecture is:
- The parent component wraps its children with `withMulti` (as the Datatable does with DatatableContent)
- The children display results on map
- There is an intermediate component that manage `terms` update when the user moves

Terms are correctly updated. But if I move the Map, the marker disappear, as results become `undefined` (ie `graphql` `props.data[resolverName]` is undefined). No query is triggered according to network. A call to `refetch` fixes but provokes blinking and sounds like a palliative.

However, when the next polling is triggered, the correct data are retrieved. It's just changing `terms` that does not behave as expected.

This issue seems to have appeared to other people, but not all the time, so it seems related to how we define the React component and how we use `withMulti`.
I can reproduce even with `apollo-client` and `apollo-react` version 2 as well as the current Vulcan version. Maybe smth is failing silently ?

I will try to setup a minimal repro, but in the meantime here is my component code, in case someone  understand this issue (or want to setup a Leaflet Map that almost work :)):

```jsx
import React, { Component } from 'react';
import { Components, registerComponent, withMulti } from 'meteor/vulcan:core';

// the parent, it creates the component wrapped with `withMulti` and 
// handle the initial render
class GeoSearchPage extends Component {
  render() {
    const { collection, defaultTerms, ...otherProps } = this.props;
    const withMultiOptions = {
      collection,
      terms: defaultTerms
    };
    const GeoSearchPageInnerWithMulti = withMulti(withMultiOptions)(
      Components.GeoSearchPageInner
    );
    return (
      <ViewportProvider
        {...otherProps}
        defaultTerms={defaultTerms}
        C={GeoSearchPageInnerWithMulti}
      />
    );
  }
}

// manage state, is in charge of updating terms and passing to its child,
// which is a component wrapped with withMulti
class ViewportProvider extends Component {
  constructor(props) {
    super(props);
    this.state = {
      currentTerms: props.defaultTerms,
      currentViewport: null
    };
  }
  isValidViewport = viewport => {
    return (
      typeof viewport.center[0] === 'number' &&
      typeof viewport.center[1] === 'number' &&
      viewport.zoom < Infinity
    );
  };
  shouldUpdateTerms = newViewport => {
    const { currentViewport } = this.state;
    if (!this.isValidViewport(newViewport)) return false;
    if (!currentViewport && newViewport) return true;
    return (
[...]

PR #2113, opened 2018-10-29
Fix datatable bug when sorting a column

Apollinaire:
see https://github.com/VulcanJS/Vulcan/issues/2090#issuecomment-433860782

Previously : when you clicked on a sortable column, the results of the fired query would be null, and the next polling would make it right. You had a long time with no results displayed.

Now : The query fired when you click to sort a column returns the right result, with the props updating correctly, so no need to wait for the polling to update the results.

See #2090 for an in-depth explanation


SachaG:
Thanks for working this out! I noticed the issue but couldn't figure out the root cause,

SachaG:
Feel free to merge this yourself if you have rights, otherwise let me know and I'll do it and/or add you to the core team.

PR #2117, opened 2018-10-31
Respect user.isAdmin

eric-burel:
`makeUsersAdmin` callback was overriding user provided `isAdmin` prop. Added tests too


SachaG:
Thanks, but could you filter out cosmetic changes? It seems like a lot of this is just formatting changes?

eric-burel:
closed in favour of #2122

SachaG:
Yes we should run Prettier everywhere, but there's some files where it would hurt readability (for example, sometimes you repeat the same pattern 3 times with different variables, and then the 3 lines get formatted slightly differently because of line lengths) so I've refrained from doing it until I can figure out better Prettier settings.

PR #2119, opened 2018-11-02
Unit testable callbacks - Oops

jimrandomh:
Last commit was somewhat messed up (I made a change to prevent running out of IDs, but it had an error of the sort lint would've caught, except we don't lint Vulcan. Oops)


jimrandomh:
Oops, chose the wrong base for this PR. Should be against the LW2 fork.

PR #2122, opened 2018-11-05
Respect user.isAdmin on creation

eric-burel:


eric-burel:
I'll merge since its fully tested and I have cleaned up linting, there should be no regression

PR #2123, opened 2018-11-05
rollback callbacks test

eric-burel:


eric-burel:
see https://github.com/VulcanJS/Vulcan/issues/2118

PR #2125, opened 2018-11-13
Export mergeWithComponents

justinr1234:


eric-burel:
Hi,
What is your use case? Maybe I can create an HOC to wrap this instead?

justinr1234:
Replacing form components that use this function as I need to rearrange how the component is put together for a different UI

eric-burel:
Take a look at https://github.com/VulcanJS/Vulcan/pull/2126
I've setup a `withComponent` hoc. You can copy paste this an `mergeWithComponents` until we merge/publicly expose it. You would rather need the HOC instead of using the function directly.

PR #2127, opened 2018-11-15
Don't reset forms to default with spurious prop changes

jimrandomh:
Currently, when any props of a Form change, it reinitializes the form, resetting all values to defaults. Unfortunately, it's pretty easy to accidentally end up with some spurious property changes. Due to some ugly hacks in the LessWrong codebase's form components, most fields ended up ignoring the change, leaving a subtle bug in those components which didn't.

### DIFF
--- a/packages/vulcan-forms/lib/components/Form.jsx
+++ b/packages/vulcan-forms/lib/components/Form.jsx
@@ -635,15 +635,13 @@ class SmartForm extends Component {
   // --------------------------------------------------------------------- //
 
   /*
-  
-  When props change, reinitialize state
-  
-  // TODO: only need to check nextProps.prefilledProps?
-  // TODO: see https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html
-   
+
+  When prefilled props change, reinitialize state, resetting form contents
+  to default
+
   */
   UNSAFE_componentWillReceiveProps(nextProps) {
-    if (!isEqual(this.props, nextProps)) {
+    if (!isEqual(this.props.prefilledProps, nextProps.prefilledProps)) {
       this.setState(getInitialStateFromProps(nextProps));
     }
   }


SachaG:
@eric-burel what do you think?

justinr1234:
It will break behavior in my app if we make this change. If the root cause is resettting form values accidentally, which I agree is a bug, then the getInitisl values should respect existing values.

eric-burel:
I'll take a closer look but basically, such issues are typical when computing state based on props. This is the relevant ressource as mentioned in the comments: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

@justinr1234 what do you call the `currentDocument`? You mean the current state of the form? Can you also develop a bit more what behaviour it breaks in your app (form in stale state after a prop change I guess?).

In any case `getInitialStateFromProps` should stay pure so it can't rely on the current state.

First I'd suggest to write one or more unit tests for the `getInitialStateFropProps` function. Since it's a pure function it should not be hard to write down a few working case and maybe a failing case.

My question is: when do we actually need to reset the form? Because that seems the underlying issue here.

justinr1234:
When you are in a forum view and switch pages React doesn’t always unmount components if the tree doesn’t change. If you, for example, have a split pane view where a list of emails Is on one side and the actual view of the specific email is on the other side, then if you navigate the URL prop with postId and pass that down into the email view side,  react well never on mount the view components and will instead just change the props. Unless you recompute  that derive state on the change you will get steel values with in the form. And now that I think about it we probably should ignore this change all together as changing the props should cause a re-render and you should not be passing down props that are extraneous

eric-burel:
Changing the props should indeed cause a re-render but here the issue is about whether we keep the state or not. Resetting state on any prop change is definitely not optimal. @jimrandomh request is legitimate imho, we can just improve the PR by doing a better selection of props that should/should not trigger a reset.

jimrandomh:
When the form is hooked up to Apollo database operations, some of its props are Apollo mutators that, in practice, will change every render no matter what.

justinr1234:
Do you have a code sample?

eric-burel:
I am writing a PR to test this better.
Here are my idea about when state should be reset or not:

Props that should trigger a state reset when modified: `'collection', 'collectionName', 'typeName', 'document', 'schema', 'currentUser'`.
**Props I'd specifically like to have your opinion about**: `'fields', 'removeFields', 'prefilledProps', 'client'`. I think fields should reset, prefilledProps should not, and I don't know about the others.

Other props should NOT reset the state, this includes graphQL mutation, prefilledProps (would be merged), callbacks, and visual/ui config.

eric-burel:
Take a look a #2131. Tests are passing but @justinr1234 @jimrandomh it would be nice if you could give this a try on your apps or at least give me feedback on the code.
I defined a `RESET_PROPS` array that lists all props that should clear the form (collection, document, fields, collectionName and so on). Other props won't trigger a form clear when changed.

justinr1234:
I believe we should be changing on any known props changing. Specifically the use case i mentioned above about changing routes but components not being unmounted due to identical trees with different props. If you want to ignore changes for unknown props, then I think that makes sense. In this case maybe `Object.keys(FormComponent.propTypes)` gets you the list of props.

eric-burel:
I am not sure it is related to mounting, componentWillReceiveProps should be called on any prop change, it's not a problem if your component is not unmounted.
Using `Object.keys(Form.propTypes)` won't solve the issue you mentioned, as it will simply track callbacks that are irrelevant to the state (the one that seems to bother @jimrandomh in this precise case). You can't rely on those callbacks values to reset your form.

However you are right that if you keep the same props but change the route, e.g because you have very same form (same collection and all) but on different routes, the state might stay the same and the form would not be cleared. But that might be the expected behaviour. You can still play on `prefilledProps` to force a reset. 
Note that there might underlying issues or I might have misunderstood smth.

justinr1234:
I'm saying specifically, all known component props should cause a reset. If props are being passed down (via an HoC) that aren't relevant, then we shouldn't reset. This is similar to the UI components where you have `inputProperties` to avoid passing down props that shouldn't be passed from the parent component.

In this case, we can't control someone passing in HOC props that they want to pass. But we can avoid resetting for those props. But we shouldn't pick and choose resetting on other known props as there are different use cases.

SachaG:
I didn't really follow the discussion here, but what's the status on this? Can we settle on something for now even if it's not the perfect solution?

eric-burel:
Didn't have feedback yet but it should be solved by #2131 , I'll merge next and stay available if it actually provokes regressions (it's just 3 lines of code anyway, my PR is mostly a unit test).

eric-burel:
Closing for now, should be solved by PR #2131 
Let me know if I should reoen

PR #2129, opened 2018-11-22
Error when submitting a SmartForm with intl fields two times

m-kreis:
When picking fields in `Form.jsx` `getData` the `__typename` property is omitted. But on (nested) intl fields the `__typename` property is still there and graphQL throws an error.

Like on `getInitialStateFromProps` i added `Utils.removeProperty(initialDocument, '__typename');` which gets rid of `__typename`.


SachaG:
Can you give an example of a schema that would trigger that error?

m-kreis:
You can have a look at https://github.com/m-kreis/VulcanPR2129/blob/master/packages/em--event-modules/lib/modules/event-modules/schema.js
If you edit the fields twice you get the error. 
If you comment out the replacement of the SmartForm Default Component then you can see the behavior of formsy replacing the french value with the german one on save.

PR #2130, opened 2018-11-26
Add Prettier and Husky

nnennajohn:
This commit is to add Prettier to Vulcan's commit pipeline. 

Essentially, before each push, we run a script to see if changed files already have prettier run, if not, we show a message to the user to run ```prettier-all```. We could just skip this check and run ```prettier-all```, which is what I typically do. But I followed what the official react package does and used their own script. 

Also added Husky to do the actually running of the script. Not sure how React team does it, but they have a few script run that does not apply to us. And Husky is what I've used in the past. I am not particularly tied to it, and we could swap this for something else like "git-pre-commit", if a need arises.

We run ```lint``` on pre-commit and ```prettier``` on pre-push.

If this looks good, I can add this to the Vulcan-Starter as well. 

Let me know your thoughts. 

Thank you.

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,0 +1,21 @@
+'use strict';
+
+const {esNextPaths} = require('./.vulcan/shared/pathsByLanguageVersion');
+
+module.exports = {
+  bracketSpacing: false,
+  singleQuote: true,
+  jsxBracketSameLine: true,
+  trailingComma: 'es5',
+  printWidth: 80,
+  parser: 'babylon',
+
+  overrides: [
+    {
+      files: esNextPaths,
+      options: {
+        trailingComma: 'all',
+      },
+    },
+  ],
+};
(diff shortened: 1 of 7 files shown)


SachaG:
Sorry for not merging this yet, I haven't had time to test it out but I'll do it soon.

nnennajohn:
Hey @SachaG No worries at all. Will be on the lookout for any feedback. Thanks.

SachaG:
Seems to work well! Thanks so much! Next step is reviewing/documenting all the different linting config files we have in the repo. Someone should write a blog post about that actually…

SachaG:
Actually if you have ever have some extra time, I think adding the same thing to the [Starter Repo](https://github.com/VulcanJS/Vulcan-Starter/) could be super handy :)

nnennajohn:
@SachaG Ok. Will work on adding it to the starter kit and also perhaps update the docs. Thanks.

PR #2131, opened 2018-11-26
Feature/smart form reset

eric-burel:
See https://github.com/VulcanJS/Vulcan/pull/2127


SachaG:
I didn't really follow that thread myself so I'll let you merge it, you're more familiar with this issue than I am.

eric-burel:
I'd be glad to get feedback from @justinr1234 and @jimrandomh on this. This should work as-is in theory (and according to unit tests) but I do not have an example app complex enough to test this correctly in real life context.

justinr1234:
Looks good

PR #2132, opened 2018-11-26
wrap source_version git cmd in a try catch

eric-burel:
@see https://github.com/zodern/meteor-up/issues/807
On my local machine (Ubuntu 18, terminator shell) the command fail because it does not seem to be triggered in the correct folder when running tests. This is just a quickfix to avoid failure


SachaG:
Oh sorry I didn't see this and fixed it myself :(

PR #2137, opened 2018-12-05
add refetch to props on withSingle

Neobii:
This adds `refetch` on the props that are passed when using the `withSingle` HOC


SachaG:
Good idea.

Issue #2138, opened 2018-12-05
add minCount and maxCount for smartForm


Neobii:
This will add minCount and maxCount to the schema arrays to control form entries on SmartForm [simple schema](https://github.com/aldeed/simple-schema-js/blob/master/README.md#mincountmaxcount)

- [x] minCount should automatically generate a minimum number of subfields
- [x] maxCount will hide the add more button when max entries are reached
- [x] hide remove button if entries are at minCount

Neobii:
[New Branch](https://github.com/Neobii/Vulcan/tree/countRestraint)

PR #2139, opened 2018-12-06
Min max count

Neobii:
Adds the ability to supply `minCount` and `maxCount` on your collection schema.  #2138 [WIP]


Neobii:
I have a cleaner pr coming up.

PR #2141, opened 2018-12-07
add minCount and maxCount to SmartForm nested arrays

Neobii:
This allows you to supply minCount and maxCount on the form fields.

Supplying min count will automatically add nested form fields.

Hides ui when max and min are reached.

#2138

### DIFF
--- a/packages/vulcan-forms/lib/components/Form.jsx
+++ b/packages/vulcan-forms/lib/components/Form.jsx
@@ -97,6 +97,20 @@ const getInitialStateFromProps = nextProps => {
     nextProps.prefilledProps,
     nextProps.document
   );
+  
+  //if minCount is specified, go ahead and create empty nested documents
+  Object.keys(convertedSchema).forEach(key => {
+    let minCount = convertedSchema[key].minCount;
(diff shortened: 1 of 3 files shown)


SachaG:
Thanks! That looks like a good implementation. I assume validation is already taken care of by SimpleSchema on the server?

SachaG:
My only worry is whether creating empty documents can somehow mess things up. And edge cases like what would happen if a field is not required, but does have a `minCount`. But we can probably deal with that if/when it becomes an issue. 

Also I wonder if maybe we should display a counter (e.g. `2/3`) somewhere so that people understand why the "add item" button disappears when you reach the max. That can be a separate PR though.

Neobii:
I can work on that edge case, as for the counter do you think it should be optional to display?

Neobii:
The counter might not make sense when you specify minCount because that will populate 2 empty fields and the count would already show `minCount/maxCount`

SachaG:
Yeah I'm not sure about the counter, I haven't needed this feature yet myself so I'm not sure what makes the most sense. Let's use this as a start and improve it later :)

Issue #2144, opened 2018-12-09
Collection Creation Hooks


Neobii:
Add hooks to creation collection like [Intl](https://github.com/VulcanJS/Vulcan/blob/devel/packages/vulcan-lib/lib/modules/collections.js#L155-L188).  Callbacks almost need to be ensured to run in order to ensure callbacks like intl run last.

SachaG:
PR welcome!

Neobii:
I will work on it this week, we just have to agree on callback name you mentioned `{collectionName}.collection.sync` but maybe this could be `{collectionName}.schema`.

PR #2146, opened 2018-12-10
add schema modifier hook on collection creation

Neobii:
#2144 this adds schema creation hooks, adds the intl fields callback last and the runs all the hooks on collection creation.

### DIFF
--- a/packages/vulcan-lib/lib/modules/collections.js
+++ b/packages/vulcan-lib/lib/modules/collections.js
@@ -2,7 +2,7 @@ import { Mongo } from 'meteor/mongo';
 import SimpleSchema from 'simpl-schema';
 import { addGraphQLCollection, addToGraphQLContext } from './graphql.js';
 import { Utils } from './utils.js';
-import { runCallbacks, runCallbacksAsync } from './callbacks.js';
+import { runCallbacks, runCallbacksAsync, registerCallback, addCallback } from './callbacks.js';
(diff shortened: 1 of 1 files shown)


SachaG:
Nice! My main feedback would be to move `registerCallback` out of the `createCollection` function (move it to the very end of the file for example) as it's purely a documentation function. 

As for the `addIntlFields` callback, it's also weird to be adding it inside the function that runs it even though I see why you did it this way… I would probably either leave it like before (i.e. not in a callback) until we add an option for running callbacks last; or else put the `addCallback` in this same file but in a `Meteor.startup()` block at the end, which would ensure it gets added near the end of the callback stack. 

Oh and also we should add a `*` callback that would run for every collection, which is what we'd use for the intl callback. So the naming I suggest would be:

- `*.collection`
- `${collectionName}.collection`
- `*.collection.async` (currently named '*.collection')
- `${collectionName}.collection.async` (currently named `${collectionName}.collection`)

Hopefully those changes can also help clean up the Git diff, because the way it's done now it appears a lot more code has changed than actually has.

Neobii:
Yeah, I didn't like what the git diff was giving me and I'll put the addCallback in the Meteor.startup() area for now.  I wanted to show that the intl fields were using this callback and how it presents two new issues: callback run order and naming convention for the new schema field name added ie [originalField]_draft_intl.

PR #2157, opened 2018-12-21
Apollo2 finalization (backward compatibility)

eric-burel:
We are close!

Apollo2 is mostly working, now the main goal is backward compatibility. See the last part of my comment: https://github.com/VulcanJS/Vulcan/issues/2081#issuecomment-427384392


SachaG:
Maybe you've already addressed that somewhere but what should we do with `webAppConnectHandlersUse`? What's the replacement? Or do we still need it?

SachaG:
Also you disabled `currentRoute`? It was a bit of a hack but it was useful to get the Vulcan route name. It's probably not widely used though.

SachaG:
oh sorry I'll comment on the other PR.

PR #2172, opened 2019-01-04
input-text counter covers large text

calinn:
max counter in a new or edit form covers text on the input text right, so, its better this way.


SachaG:
Sorry, looks like there are some conflicts? Also could you include screenshots of the issue and the fix?

calinn:
schnauzer1
schanuzer2
but in a textarea input the counter is on the middle right side.

Apollinaire:
About `update to React Bootstrap 4` and `fix close button flash message` : could you revert these commits? I think they belong in another PR.
Also, don't remove props/features from core components like `Flash`. You can add some, but removing them is a breaking change. Instead, you could make this change in ui components like `Alert` in `vulcan:ui-bootstrap`. This way your update won't break other ui packages :)

calinn:
Yes, I can. I reverted it. Sorry for the mess.

Apollinaire:
@SachaG I don't use ui-bootstrap, could you take care of merging this one?

SachaG:
I think it's better to create a new, clean PR actually. I'll close this one for now, sorry…

PR #2177, opened 2019-01-08
Debug & Admin layouts

Apollinaire:
- Added a new column called `layoutName` to the routes' debug dashboard in `vulcan:debug`.
- Renamed the layout called `AdminLayout` to `DebugLayout` in `vulcan:debug`. I chose to do this because i thought the name `AdminLayout` was misleading and could lead to people thinking that this was a layout only accessible to admin users.
- Created an `AdminLayout` in the `vulcan:admin` package to serve as an example for a Layout that is only accessible to admins.

What do you think about the second and third changes?

Also I noticed that the `vulcan:payments` package registers routes using an `AdminLayout`, but it does not register it and it's not mentioned in its readme or docs.


Apollinaire:
I forgot about `withAccess` ! I will rewrite the AdminLayout.

Apollinaire:
Why does `withAcces` not have a fallback component like `Components.ShowIf` ?

SachaG:
Probably just an oversight on my part, I'm surprised it doesn't have one actually. 

To be honest a lot of things in Vulcan are "starting point" features that I set up so others could come along and improve them later on :)

Apollinaire:
Maybe I'll improve it later then ;)

Issue #2140, opened 2018-12-06
Create a getLabel function


Apollinaire:
There are mutliple components that use the label of a field : Card, Datatable, SmartForm. They all have their own way of getting the label. I think it could be good to have a function that unifies this pattern, to avoid inconsistency and enforce DRY. 

## Behaviour of each function

- Card: look for `${collection._name}.${fieldName}` in i18n, fallback to `schema[fieldName].label`
- Datatable: look for `${collection._name}.${columnName}` in i18n, look for `schema[columnName].label`, fallback to raw column name
- SmartForm: look for `${collectionName}.${fieldName}`, `global.fieldName`, `fieldName` in i18n, fallback to schema label, fallback to raw fieldname

### Card

```js
const getLabel = (field, fieldName, collection, intl) => {
  const schema = collection && collection.simpleSchema()._schema;
  const fieldSchema = schema && schema[fieldName];
  if (fieldSchema) {
    return intl.formatMessage({id: `${collection._name}.${fieldName}`, defaultMessage: fieldSchema.label});
  } else {
    return fieldName;
  }
}
```

### Datatable

```js
const columnName = typeof column === 'string' ? column : column.label || column.name;
  
  if (collection) {
    const schema = collection.simpleSchema()._schema;

    /*

    use either:

    1. the column name translation
    2. the column name label in the schema (if the column name matches a schema field)
    3. the raw column name.

    */
    const defaultMessage = schema[columnName] ? schema[columnName].label : Utils.camelToSpaces(columnName);
    const formattedLabel = intl.formatMessage({ id: `${collection._name}.${columnName}`, defaultMessage });

    // if sortable is a string, use it as the name of the property to sort by. If it's just `true`, use column.name
    const sortPropertyName = typeof column.sortable === 'string' ? column.sortable : column.name;
    return column.sortable ? <Components.DatatableSorter name={sortPropertyName} label={formattedLabel} toggleSort={toggleSort} currentSort={currentSort} sortable={column.sortable}/> : <th>{formattedLabel}</th>;

  } else {

    const formattedLabel = intl.formatMessage({ id: columnName, defaultMessage: columnName });
    return <th className={`datatable-th-${columnName.toLowerCase().replace(/\s/g,'-')}`}>{formattedLabel}</th>;

  }
```

### SmartForm

```js
  getLabel = (fieldName, fieldLocale) => {
    const collectionName = this.getCollection().options.collectionName.toLowerCase();
[...]

PR #2184, opened 2019-01-21
[WIP] I18n: formatLabel

Apollinaire:
This is related to issue #2140 : creating a function to centralize how we generate a label from the schema & the i18n fields. 
I've written a function `formatLabel` that lives in `vulcan:i18n`, is available from the i18n context and has the following API:

```js
formatLabel: function({ fieldName, collectionName, schema }, values) => String
```
NB: i've let collectionName because it was used in Forms, so I did not want to introduce a breaking change. However, in regards to openCRUD, it would be better to use the typeName instead. Should we change this? Should we do both to keep backwards compatibility?

The `values` are passed to `formatMessage` to be able to add values in the i18n text fields.

Here is the logic behind label generation: for example a field `year` in example-movies, collection `Movies`.

```
i18n('movies.year') > i18n('global.year') > i18n('year') > schema.year.label > "Year"
```

If we agree to make these changes, I'll rework the parts related to labels in SmartForm, Datatable, and Card. 

Advancement: 

- [X] write a `formatLabel` function
- [x] rework `getLabel` in Form.jsx
- [x] rework the labels generation in DatatableHeader
- [x] rework `getLabel` in Card.jsx
- [ ] add the function to the Internationalization page of the docs
- [x] write a test on the `formatLabel` function @Neobii

### DIFF
--- a/packages/vulcan-core/lib/modules/components/Card.jsx
+++ b/packages/vulcan-core/lib/modules/components/Card.jsx
@@ -8,12 +8,7 @@ import { Link } from 'react-router';
 
 const getLabel = (field, fieldName, collection, intl) => {
   const schema = collection && collection.simpleSchema()._schema;
-  const fieldSchema = schema && schema[fieldName];
-  if (fieldSchema) {
-    return intl.formatMessage({id: `${collection._name}.${fieldName}`, defaultMessage: fieldSchema.label});
-  } else {
-    return fieldName;
-  }
+  return intl.formatLabel({ fieldName: fieldName, collectionName: collection._name, schema: schema });
 };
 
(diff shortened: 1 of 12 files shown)


Apollinaire:
Closing this since the `devel` branch is Apollo2 now. I'll merge it in `devel-apollo1`.
Note for the future @SachaG  @eric-burel : we shouldn't merge big updates such as Apollo2 directly on devel, instead rename devel to apollo1-devel, then rename apollo2 to devel and make it the default branch. I think this way it does not break the pending PRs like that.

SachaG:
Good point. Let's be more careful next time we merge in a big update.

PR #2192, opened 2019-02-01
Default view sort doesn't take precedence over selected view

jimrandomh:
If a default view on a collection has an `options.sort` set, Vulcan currently deep-merges this with the `options.sort` on a view, which may also include a sort. Unfortunately, this merge isn't key-order-aware (and is in fact backwards), so this means that a default view can't have a sort specified without overriding the sort on all specific views.

Add a special case for `view.options.sort`, so tahat they don't merge.


SachaG:
Thanks! Yeah that makes sense I think.

PR #2194, opened 2019-02-08
vulcan-events: fix TypeError: Cannot redefine property: name

calinn:
google analitycs and vulcan-events get error so google bot from google search console can't load rest resource to render web page

### DIFF
--- a/packages/vulcan-events/lib/modules/events.js
+++ b/packages/vulcan-events/lib/modules/events.js
@@ -37,12 +37,15 @@ export const addIdentifyFunction = f => {
 };
 
 export const addPageFunction = f => {
-  const f2 = ({ currentRoute }) => f(currentRoute);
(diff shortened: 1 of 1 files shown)


calinn:
I think it resolve it:
https://github.com/emberjs/ember.js/issues/4567#issuecomment-38351180

SachaG:
`f2 = {...f2};` seems a bit hacky? Maybe you could use lodash's `clone`, or at least add some comments to make it clearer what's going on?

calinn:
I'm cloning it to make a copy of f2, because f2 already has a 'name' attribute from f

SachaG:
But f2 is a function? What does `{...f2}` mean for a function?

calinn:
For functions {...f2} is a null object {}. f2={}. you're right it's wrong

PR #2200, opened 2019-02-14
vulcan-events: TypeError: Cannot redefine property: name

calinn:
handling error for Object.defineProperty()


SachaG:
I still feel like that doesn't really solve the issue? Is there no way to fix the root cause instead of just catching the error?

eric-burel:
Closing as stale

PR #2204, opened 2019-02-16
add rootElement check on vulcan-core/lib/client/start.jsx

darwis059:
in my app i got two rootElement on client. rootElement already exist from SSR, but start.jsx  append it again to body. i don't know why this doesn't happen before upgrade to Apollo 2

### DIFF
--- a/packages/vulcan-core/lib/client/start.jsx
+++ b/packages/vulcan-core/lib/client/start.jsx
@@ -18,10 +18,12 @@ Meteor.startup(() => {
   const apolloClient = createApolloClient();
 
   // Create the root element
-  const rootElement = document.createElement('div');
-  rootElement.id = 'react-app';
-  document.body.appendChild(rootElement);
-
+  if (!document.getElementById('react-app')) {
+    const rootElement = document.createElement('div');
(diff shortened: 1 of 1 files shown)


SachaG:
Thanks but maybe we should figure out why this is happening in the first place? That fix might just be hiding a deeper issue?

eric-burel:
Thanks for the PR. Yes I am not 100% on how it's supposed to work, not sure indeed there is not another underlying mistake.

eric-burel:
@loustak @yannick-mayeur if you have some spare time, this may be a similar bug to the double `<head>` tag thing, maybe we could take a look at it?

yannick-mayeur:
We are looking into it.

yannick-mayeur:
This is done in the renderPage.js file:

```
const wrappedHtmlContent = `<div id="react-app">${htmlContent}</div>`;
```

It should probably be removed. I am doing a PR to do this.

yannick-mayeur:
See: https://github.com/VulcanJS/Vulcan/pull/2511

eric-burel:
Was solved in previous update.
Thanks for pointing the issue.

PR #2212, opened 2019-02-26
Do not wrap the modal trigger

vale981:
Wrapping the trigger is inconvenient. (e.g. in Tables)


SachaG:
This might break some apps if people rely on the `modal-trigger` classname though… Can you give an example of it being inconvenient?

vale981:
If you build lists of triggers, you want the triggers to be `li` elements, not `div`.

SachaG:
You could put the `div`s inside the `li`s?

vale981:
Right. Bad example. Another issue is a `tr` with `td` etc...

SachaG:
I still don't see why you can't just put the `div` inside the `td`. I agree that not having a div would be better, but in this case it just doesn't seem worth it to potentially break existing apps.

PR #2215, opened 2019-02-27
Add classname for required fields.

vale981:


SachaG:
If that's useful to you I can merge it, but note that the whole accounts package will be refactored soon.

vale981:
That's good, because it is a mess :)

SachaG:
Yeah I agree… for what it's worth it wasn't originally part of Vulcan, it was written by somebody else and then we used it because there was no other React-compatible accounts package for Meteor,.

PR #2226, opened 2019-03-04
Vulcan:users Add function createGroup

enzolutions:


SachaG:
I don't really understand what `parentRouteName` does or what routes have to do with groups?

SachaG:
Also it'd be better to accept an object (`createGroup({ name, description, label }))`).

enzolutions:
@SachaG My fault I copy some code from other places, I clean up, but I still have the error.

SachaG:
What error is that? Also I think  `Users.createGroup` should just be an alias for `createGroup`. In other words they should be the same function with the same API. It could be very confusing if one accepts arrays but the other doesn't.

enzolutions:
Sorry, @SachaG I understand what you mean, but I don't know how to do that.

The error I got before is **TypeError: createGroup is not a function** but with the alias, approach looks like that everything changes, so we could ignore that.

SachaG:
I think something like this would work?

```js
export function createGroup = (groupOrGroupName) => {
  const group = typeof groupOrGroupName === 'string' ? { name: groupOrGroupName} : groupOrGroupName;
// ...
}
```

Of course if every group is now an object, we probably also need to tweak any code that makes use of groups to check permissions, assign to users, etc. 

Also if `createGroup` needs to support arrays that makes the logic a little bit more complicated.

enzolutions:
@SachaG I tried to create the alias as you recommended, but when I tried to include in this way

```
import {Users, createGroup } from 'meteor/vulcan:users';
```

I got the following error

```
TypeError: createGroup is not a function
    at permissions.js (packages/keeptrack/lib/modules/permissions.js:40:1)
    at fileEvaluate (packages/modules-runtime.js:336:7)
    at Module.require (packages/modules-runtime.js:238:14)
```

Any recommendation? thanks in advance.

SachaG:
Maybe `createGroup` is not exported properly out of the package? Although `export * from './permissions.js';` should be taking care of that… Is it `undefined`? Or something else?

Also btw I think it should be `import Users, { createGroup }` since `Users` is the default export.

SachaG:
Or maybe it's the `export const createGroup = Users.createGroup =...` syntax that's messing with the export somehow?

enzolutions:
Hi @SachaG 

Yes I also tried `import Users, { createGroup }`, but either way, the **createGroup** is always **undefined**

PR #2228, opened 2019-03-05
move throw error to server

### LINKED ISSUE #2224: throwError is not defined
Throw error is defined in `/packages/vulcan-lib/lib/server/errors.js` but importet by `/packages/vulcan-lib/lib/modules/utils.js` from `/packages/vulcan-lib/lib/modules/errors.js`.

I am opening this as issue for I don't know what to make of it. The only solution i can think of is to shift all code using throwError to the server, or to rewrite `throwError` to work on the client.

vale981:
Fixes #2224


SachaG:
Thanks, but I actually ended up fixing it myself! Sorry about that…

PR #2229, opened 2019-03-06
Fix nested field specifiers for forms.

vale981:
Specifying fields for nested schemas did not work.
This PR should fix the issue relatively elegantly.

### Examples:
```jsx
 <Components.SmartForm
                 collectionName="Users"
                 fields={['userProfile', 'userProfile.title']}
                 documentId={this.props.currentUser._id} />;
```
Selects the `userPofile.title` field to be displayed in the form. 

```jsx
 <Components.SmartForm
                 collectionName="Users"
                 fields={['userProfile']}
                 documentId={this.props.currentUser._id} />;
```
Renders all the subfields of `userProfile`.


SachaG:
Thanks, that's a good idea! But to clarify, this means "show all subfields":

```
fields={['userProfile']}
```

And this means "show only userProfile.title":

```
fields={['userProfile.title']}
```

But what about this?

```
fields={['userProfile', 'userProfile.title']}
```

It seems like it could be interpreted either way?

vale981:
`fields={['userProfile.title']}` would be invalid with this code

SachaG:
I'm not sure, what makes the most sense? I think personally I would expect `fields={['userProfile.title']}` to work and show the `title` field, and `fields={['userProfile', 'userProfile.title']}` to show all subfields (so basically ignore `userProfile.title`). 

Also since this might potentially break how forms work and I'll need some time to test it out I might delay merging it until the current `devel` is merged and release, just so you know.

vale981:
i am going to create a new PR with this functionality

vale981:
See #2260

PR #2230, opened 2019-03-06
Temporarely fix apollo error bug.

vale981:
See https://github.com/apollographql/apollo-server/issues/2155


SachaG:
That seems like a regression? Can you test the current code with the latest version of `graphql-tools` first?

vale981:
Amen. Updating `graphql-tools` does the trick!

PR #2232, opened 2019-03-07
Upgrade react-router

vale981:
This upgrades `react-router` to version 4.x.

Needs further testing.

Breaking changes: 
 - `props.params` is now `props.match.params`


SachaG:
Thanks! Note that the `package.json` in the core Vulcan repo isn't actually ever used so we might want to get rid of it… But as long as it's here it's good that it's up to date.

justinr1234:
I think we should keep it, it is used if you clone the repo directly as a baseline for your app instead of relying on Atmosphere.

Apollinaire:
Package.json & node_modules for this repo are used by the tests too maybe ?

eric-burel:
Indeed we need it for tests and now for Storybook too. It's a bit hard to keep both package.json in sync, the update-package.js script does not work yet when some versions are locked in the package.json (the issue comes from a 3rd party lib https://github.com/userfrosting/merge-package-dependencies).

PR #2234, opened 2019-03-07
Add default locale to getString.

vale981:
Adds a default locale export for intl and default functionality for getString.

Takes the locale from the settings (which in turn defaults to 'en').


SachaG:
Thanks!

PR #2235, opened 2019-03-07
Add wrapperless Emails.

vale981:
There ain't no wrapper anywhere, so don't require one.

(Sth. to add to the docs?)


SachaG:
Yeah maybe. The whole email package is probably in need of a fresh look to be honest.

PR #2236, opened 2019-03-07
fix values property for handlebars helper, use context as default

vale981:


SachaG:
Thanks for all the fixes :)

PR #2238, opened 2019-03-09
Add debug messages for missing strings and default locale on per-message level.

vale981:
The previous PR used the default locales only if the chosen locale wasn't defined.
Now the fallback happens on a per-message level.

I've added debug logging for missing locale strings too, as they are hard to spot.
We could modify the `FormattedMessage` component to show a red error message if no string is found...


SachaG:
Debug messages are a good idea! Btw what's `O_CREAT`?

vale981:
I have no idea where that came from!

vale981:
I have removed the import and squashed my commits.

SachaG:
Thanks! Sometimes autocomplete can add some weird stuff if you're not careful…

PR #2240, opened 2019-03-09
Add locale on signup.

vale981:
Adds the current locale to the user profile on signup.
This is useful for localizing the email confirmation.


SachaG:
Good idea. The signup form will probably change soon but in the meantime this is a good improvement.

PR #2241, opened 2019-03-10
Fix error on invalid form submit.

vale981:
When a smart form had errors and was submitted again it let to a problem with accessing `mutation.result.data[mutationName].data`.

This PR adds a little tweak and now it seems to work alright.

### DIFF
--- a/packages/vulcan-core/lib/modules/default_mutations.js
+++ b/packages/vulcan-core/lib/modules/default_mutations.js
@@ -24,6 +24,7 @@ import {
 } from 'meteor/vulcan:lib';
 import Users from 'meteor/vulcan:users';
 import isEmpty from 'lodash/isEmpty';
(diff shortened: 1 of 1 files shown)


SachaG:
Thanks! But maybe using https://lodash.com/docs/4.17.10#get would make it cleaner? Not a big deal though, I can merge this as it is otherwise.

vale981:
Yeah. I did not know about this ^^.

SachaG:
Shouldn't it be

```
get(mutation, `result.data[${mutationName}].data`, {})
```

rather than

```
get(mutation, `result.data['${mutationName}'.data]`, {})
```

?

Also I think the `'` inside the path are unnnecessary.

vale981:
This has become irrelevant if i am not mistaken.

PR #2242, opened 2019-03-10
Clean document/data in default mutators.

vale981:
Makes submission of dates possible, as the clean function converts types!

I am not entirely sure if this is the right place to put it, but hey, it works and I didn't want to chain it in more centrally, as checking is not always desired.

(Maye we should make checking optional?)

### DIFF
--- a/packages/vulcan-lib/lib/server/mutators.js
+++ b/packages/vulcan-lib/lib/server/mutators.js
@@ -65,6 +65,9 @@ export const createMutator = async ({
   // we don't want to modify the original document
   document = data || document;
 
+  // clean document
+  document = collection.simpleSchema().clean(document);
+
   const { collectionName, typeName } = collection.options;
   const schema = collection.simpleSchema()._schema;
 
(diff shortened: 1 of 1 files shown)


SachaG:
Oh so you mean if you submit a date string for a `Date` field for example? I suspect in that case the GraphQL layer would still throw an error though… 

Generally I'm a bit hesitant to add any "magic" cleaning step that might modify documents into something the user doesn't necessarily expect so I'm not 100% sure about that to be honest. What if we just let user add it in a callback instead? Maybe make it easier by passing the schema as a callback option?

vale981:
Graphql doesn't complain. I did open this PR because I need it for submitting dates with autoform and i didn't want to write a helper for every date field.

vale981:
Good idea with that callback. The schema already is part of the context if I am not mistaken.

SachaG:
Actually it wasn't, but I just added it.

vale981:
Is this still relevant?

SachaG:
No, I think we can close it.

PR #2243, opened 2019-03-10
Fix labeling in select component.

vale981:
Shouldn't it be this way around?


SachaG:
Oh good catch, thanks!

PR #2247, opened 2019-03-16
Re-enable Search Input in DataTable

enzolutions:
After trying to use the DataTable component with the new Material UI Implementation, the DataTable wasn't rendered at all  get the following error and

```
Uncaught (in promise) Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```

After some debugging, I determine the error was related with the following component inside DataTable render

```
{
  showSearch &&
  
  <Components.SearchInput value={this.state.query}
                          updateQuery={this.updateQuery}
                          className={classes.search}
  />
}
```

Now, this is how looks in my Datatable.

Screen Shot 2019-03-17 at 8 12 00 am

Also, I fix some issues related to `ErrorCatcher` components, which is part of Vulcan errors packages, I added that package as a dependency.

Thanks to @digitalhydra who help me to do the re-implementation.


eric-burel:
Nice! You also updated the ModalTrigger?

enzolutions:
@eric-burel yes but I think was only because we were testing some other issues and our editor "fix" the indentation. Should I remove that?

I am checking some issues with `<Components.ErrorCatcher>` that don't allow the Modal to render the form.

enzolutions:
Also, I fix some issues related to `ErrorCatcher` components, which is part of Vulcan errors packages, I added that package as a dependency.

let me know if I have to split into two PR.

SachaG:
@eric-burel I'll let you merge this one and all Material UI stuff generally since I don't use it myself.

eric-burel:
@SachaG Yep no problem I'll handle them
@enzolutions It's ok to keep one PR, just double-check that it does not contain unexpected changes, and indeed try to cancel them if possible. 
We are close to get Storybook to work (it works for bootstrap but I just need to figure out what needs to be specifically imported for Material UI) but until then it's a bit tedious to test the UI, so I'll trust you on those changes and test them later. Just ping me when you're done so I can review and merge.

eric-burel:
And again thanks for the help, this is very welcome !

enzolutions:
Perfect.

I will try to review the select issue.

eric-burel:
Nice fixes, it already solves a few issues.
I've noticed that error messages are not correctly displayed, I'll check way.
As you've already noticed there's a small change in the `onChange` function API. I don't know if it comes from Vulcan or Material UI itself.

enzolutions:
let me know if I can help you with the error messages.

PR #2253, opened 2019-03-21
Fix denormalization when bio set to empty

jimrandomh:
Currently, when a user sets their "bio" to an empty string, the denormalized `htmlBio` field doesn't get updated, because empty-string is falsy. Fix that.


SachaG:
Thanks! We'll probably remove this field from core eventually but it's a good fix for now.

PR #2254, opened 2019-03-22
Add MuiPicker to be used in Date fields Material-UI

enzolutions:
I was having issues with the current implementation of Date fields because the MultiInput uses a text field to add the Date Picker.

To solve that I propose to use Material UI Picker as you can see in the PR

### DIFF
--- a/packages/vulcan-ui-material/lib/components/forms/base-controls/MuiPicker.jsx
+++ b/packages/vulcan-ui-material/lib/components/forms/base-controls/MuiPicker.jsx
@@ -1,0 +1,84 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import createReactClass from 'create-react-class';
+import withStyles from '@material-ui/core/styles/withStyles';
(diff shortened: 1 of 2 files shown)


eric-burel:
Merged manually, seem to work fine

enzolutions:
Awesome, @eric-burel  we still have the issue of timezone and enable custom format ( but I am not sure if this part is require to be handled by users)

eric-burel:
I temporarily removed moment-timezone indeed. I think that's something you should load in your own app, because you don't necessarily need it, except if I am wrong?
What is you use case for moment-timezone?

enzolutions:
Well in meteor (standalone) I handle that setting the server side timezone and storing the data with timezone applying that in client side, to avoid any issues with dates.

But due the forms are auto-generated in Vulcan, I don't understand how to handle a record edited and saved by a user in Australia and how is read it by a user in the USA.

Any ideas?

PR #2255, opened 2019-03-22
Re-enable Search Input in DataTable

enzolutions:
After trying to use the DataTable component with the new Material UI Implementation, the DataTable wasn't rendered at all get the following error and

Uncaught (in promise) Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
After some debugging, I determine the error was related with the following component inside DataTable render

```
{
  showSearch &&
  
  <Components.SearchInput value={this.state.query}
                          updateQuery={this.updateQuery}
                          className={classes.search}
  />
}
```

### DIFF
--- a/packages/vulcan-ui-material/lib/components/bonus/SearchInput.jsx
+++ b/packages/vulcan-ui-material/lib/components/bonus/SearchInput.jsx
@@ -15,15 +15,15 @@ const styles = theme => ({
 
   '@global': {
     'input[type=text]::-ms-clear, input[type=text]::-ms-reveal':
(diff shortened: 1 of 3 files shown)


eric-burel:
Merged manually, thanks! Please think to squash your commits beforehand with `git rebase -i HEAD^XX` to avoid having multiple identical commit

enzolutions:
Awesome

PR #2262, opened 2019-04-04
Additional tests

eric-burel:
Hi,

I've added a few tests and cleaned up GraphQL code generation. I've gathered a few remarks:

- `after` callbacks of the createMutation does not seem to be respected. This line `document = await Connectors.get(collection, document._id);` basically overrides them by fetching the document from the database. It thus ignore previously run callbacks. It should be placed before `after` callback are run, while now it is placed after.

- I splitted the code of the `graphql` generation, but actual modifications are minors. I just needed to be able to reset the state in order to write tests and I needed to refactor things a little to wrap my head around existing code.

- I could not drop a collection automatically. I tried to write an `isoCreateCollection` helpers for tests that deletes a collection if it exists already, before creating it. This way we can safely call all our test collections "Foo". More broadly even without this helper dropping collections is useful for cleaning up integration tests without dropping the whole db.

- IntlProvider tests are broken, it seems to be due to some API changes?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "Vulcan",
-  "version": "1.12.17",
+  "version": "1.13.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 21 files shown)


eric-burel:
@SachaG I merge this directly because there are no actual API change in those commits, I'll open separate issues for my remarks

PR #2263, opened 2019-04-09
Update synced-cron across a maintainer change

jimrandomh:
The `synced-cron` package was stuck on an old version (1.1.0, late 2014) because when it changed maintainers, it also changed package name. The newer versions include some somewhat important bug fixes; so, upgrade.

This is a breaking change! If you're using Vulcan, you will need to replace imports of `percolatestudio:synced-cron` with imports of `littledata:synced-cron`.

### DIFF
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -60,7 +60,7 @@ mongo-id@1.0.7
 npm-bcrypt@0.9.3
 npm-mongo@3.1.1
 ordered-dict@1.1.0
-percolatestudio:synced-cron@1.1.0
+littledata:synced-cron@1.3.1
 promise@0.11.1
 random@1.1.0
 rate-limit@1.0.9
--- a/packages/vulcan-core/lib/server/start.js
+++ b/packages/vulcan-core/lib/server/start.js
@@ -1,5 +1,5 @@
 import {Inject} from 'meteor/meteorhacks:inject-initial';
-import { SyncedCron } from 'meteor/percolatestudio:synced-cron';
+import { SyncedCron } from 'meteor/littledata:synced-cron';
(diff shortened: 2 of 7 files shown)


SachaG:
Thanks! Didn't know about this change.

SachaG:
But actually I now just use https://www.npmjs.com/package/cron. Are there good reasons to use a Meteor package for this? I guess the ability to rely on the db to avoid running a job twice, but I don't know how useful that is…?

jimrandomh:
Yes, LW runs on multiple servers and needs to avoid running the same job twice. Or, more precisely speaking: it needs to avoid starting the same job twice *simultaneously*, because that risks creating lots of subtle race conditions even in things that would theoretically be idempotent.

eric-burel:
Just to give some feedback on my experience on crons+Vulcan.

I now tend to separate cron jobs from my apps because they feel like very separate technical problem. For example, I tried to run a database backup cron, and I couldn't because the most common Meteor images won't let you install `mongodump` due to some issues in Debian. I ended up writing a shell script.
Also when the cron run in the Vulcan app, it may affect the app performance as far as I understand, since they "pollute" the Node thread loop (except if they run in a separate process or smth)?

Seperating crons from apps makes them annoying to write because I can't access Vulcan features, or even write them in JS (I could but that mean a more complex install and I did not find existing comprehensive solution for this). 
So for example sending a newsletter would be very difficult outside of Vulcan.

Regarding the database connection, I guess we could move toward using the npm package without breaking changes, we would simply need to improve the Vulcan provided crons with the same database/race conditions functionalities.

Also just a note but when updating, we need to remove `percolatestudo:` version from `.meteor/package`, otherwise it stays active and tries to create a `cronHistory` collection confliting with `littledata:` version

PR #2266, opened 2019-04-09
Add skipConversion option to mongo connectors

jimrandomh:
As part of the Vulcan 1.13/Apollo2 upgrade, I've been going through the places where LessWrong has diverged from mainline Vulcan and assessing/upstreaming things. I was on the fence about whether to upstream this one, and am on the fence about whether it should be merged, but, well, here it is.

An earlier Vulcan upgrade (1.12 I think) made it so that `documentId` had special meaning in queries, and would be replaced with `_id`. But LessWrong was already using `documentId` as a field name in a number of collections, identifying what object things refer to in some other table. For example, vote._id identifies a vote, vote.documentId identifies the thing voted on. (`vulcan-voting` still has this name conflict, though we have diverged from it heavily otherwise.)

Now that ID conversion is added, other projects may be relying on it so it may be hard to remove. So we added a safety valve, a `skipConversion` option on each of the Mongo connector operations, which makes `documentId` work as a regular field name, as it originally did.

### DIFF
--- a/packages/vulcan-lib/lib/server/connectors/mongo.js
+++ b/packages/vulcan-lib/lib/server/connectors/mongo.js
@@ -15,22 +15,27 @@ const convertUniqueSelector = selector => {
 };
 
 DatabaseConnectors.mongo = {
-  get: async (collection, selector = {}, options = {}) => {
-    return await collection.findOne(convertUniqueSelector(selector), options);
+  get: async (collection, selector = {}, options = {}, skipConversion) => {
+    const convertedSelector = skipConversion ? selector : convertUniqueSelector(selector)
+    return await collection.findOne(convertedSelector, options);
   },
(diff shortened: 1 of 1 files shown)


SachaG:
Yeah I want to fix that issue at some point, so I don't know if adding that band-aid is the best solution. I see two main ways of dealing with it:

1. Change the voting package, LessWrong, and any other codebase to use `itemId` instead of `documentId`. It's a simple change, but would require a db migration. 
2. Rethink how documents are selected in Vulcan so that neither `_id`, `documentId`, `slug`, etc. are hardcoded into core as special names. And drop the hacky conversion step altogether. 

Obviously 2. is the "real" solution but I'm not entirely sure how to do it yet. So in the meantime I'd probably suggest 1?

PR #2267, opened 2019-04-09
Mutation HoCs give equality-comparison-stable props

jimrandomh:
As part of upgrading to Vulcan 1.13/Apollo2, I'm going throw places where LW has diverged and upstreaming changes.

React decides which parts of the tree need to be rerendered by comparing component props, assuming that if a component's old and new props don't compare as equal (===), the component needs to be rerendered. If a prop is a callback or an object, and hasn't truly changed but has been reconstructed, it will compare as not-equal and cause a rerender. Vulcan's mutator HoCs add some callbacks to props, which you can call to perform mutations, but they get reconstructed - and, therefore, compare as not-equal - on every render. This caused some performance problems on LW.

This shuffles the construction of objects and callbacks around a bit, to prevent spurious rerenders caused by some Vulcan mutation HoCs. It does not cover every HoC. More importantly, this has (so far) only been  tested on Vulcan 1.12/Apollo1, and it's possible that some subtleties of the `graphql` function have changed, so it should be retested (both for not-breaking-anything and for not-rerendering) on Vulcan 1.13.

### DIFF
--- a/packages/vulcan-core/lib/modules/containers/withCreate.js
+++ b/packages/vulcan-core/lib/modules/containers/withCreate.js
@@ -31,6 +31,7 @@ import { graphql } from 'react-apollo';
 import gql from 'graphql-tag';
 import { createClientTemplate } from 'meteor/vulcan:core';
 import { extractCollectionInfo, extractFragmentInfo } from 'meteor/vulcan:lib';
+import { compose, withHandlers } from 'recompose';
 
 const withCreate = options => {
   const { collectionName, collection } = extractCollectionInfo(options);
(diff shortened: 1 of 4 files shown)


SachaG:
Sounds like a good change, but maybe we could wait until you've tested it with the new version for a while before merging?

ErikDakoda:
@jimrandomh I applied this pull request to my own fork of Vulcan to test it, and I ran into the following problem: The mutation returned by the HoC does not have a name. 

Before the change a mutation returned by withUpdate would look like this:

before

After the change the mutation returned looks like this:

after

I think this should be easy to remedy.

PR #2268, opened 2019-04-09
Fix a crash when using addFields together with a recursive schema

jimrandomh:
Forms have an `addField` prop, which can be used to override the defaults of which fields are visible in a form. We use this in combination with a custom submit button to make comment boxes show for logged out users (who don't have write/create access to the relevant fields). But it turns out that if you have a nested schema, it tries to add the extra fields into the subschema, not just the root schema, resulting in a crash.


SachaG:
Thanks!

PR #2269, opened 2019-04-09
make Datatable edit modal title customizable

carloslfu:
As well as other modal properties

### DIFF
--- a/packages/vulcan-core/lib/modules/components/Datatable.jsx
+++ b/packages/vulcan-core/lib/modules/components/Datatable.jsx
@@ -384,11 +384,13 @@ DatatableRow Component
 const DatatableRow = (props, { intl }) => {
 
   const { collection, columns, document, showEdit, 
-    currentUser, options, editFormOptions, rowClass, Components, modalProps } = props;
+    currentUser, options, editFormOptions, rowClass, Components } = props;
   const canEdit = collection && collection.options && collection.options.mutations && collection.options.mutations.edit && collection.options.mutations.edit.check(currentUser, document);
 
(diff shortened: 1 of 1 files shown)


SachaG:
Thanks but I'm not sure I like the `_modalProps` variable name, we never use that naming style anywhere else in the codebase…

carloslfu:
Yes, you are right, what do you think is a good name for it?

SachaG:
You could have `defaultModalProps` and `customModalProps`? Also this is a little hard to read:

```
const _modalProps = { title: <code>{document._id}</code>, ...(_isFunction(modalProps) ? modalProps(document) : (modalProps || {})) };
```

I would write it as:

```
const { modalProps = {} } = props
```

And then:

```
const defaultModalProps = { title: <code>{document._id}</code> }
const customModalProps = { ...defaultModalProps, ...(_isFunction(modalProps) ? modalProps(document) : modalProps) };
```

Or something similar. What do you think?

carloslfu:
It makes sense, thanks @SachaG. I'm gonna fix it.

carloslfu:
It's done, what do you think?

SachaG:
Looks good! Btw, is there a reason why you called the function `_isFunction` rather than just `isFunction`?

carloslfu:
Is following the same convention as [_sortBy](https://github.com/VulcanJS/Vulcan/blob/devel/packages/vulcan-core/lib/modules/components/Datatable.jsx#L9) that comes from Lodash.

SachaG:
Thanks for making all those changes! Nice work :)

PR #2272, opened 2019-04-10
FormInnerComponent unify Bootstrap and Material UI API

eric-burel:
I've noticed differences between FormComponentInner in Bootstrap and Material UI. This lead to issues in Storybook stories and code redundancy.

I created a `vulcan:ui-react` package to share reusable UI code, I am not sure if it's the best pattern.
I try to progressively reduce the codebase size so UI becomes more manageable.

This PR needs a bit of testing, mostly to check that it does not break apps using Bootstrap UI.

Isn't the whitelisting of inputProperties a bit weird though? What if I want to add specific accessibility parameters (I am not knowledgeable on this so I don't have much examples) ?


SachaG:
I don't know if we need one more package? Couldn't we add those helpers to `vulcan:lib` or `vulcan:core` instead?

Also the reason for the whitelisting is that if you allow any random `foo` property, that can sometimes be applied directly as attribute on an `<input>` element and React complains. You make a good point though, I'm not sure what the workaround is…

eric-burel:
Ok I'll move this back to vulcan:lib then, I think it's the most appropriate place indeed. I'll try to find a list of all valid props, I think they should be listed somewhere on official docs

eric-burel:
I've merged, it works as expected with Bootstrap in the Starter, I've put the ui utils in vulcan:lib

PR #2274, opened 2019-04-12
Fix a falsy-vs-undefined bug in getSetting

jimrandomh:
`getSetting` was checking its `settingDefault` parameter for truthiness, rather than for presence. This meant that if you pass a settingDefault of false, and the setting isn't defined and doesn't have a registered default, then instead of getting false you'll get undefined. This caused some subtle breakage in LessWrong when the result of getSetting was then used in a mongodb selector, where undefined and false have very different meanings.


SachaG:
Thanks!

PR #2275, opened 2019-04-16
ui-material: updated Datatable.jsx - header fix

Julien-Sytadelle:
intl.formatMessage returns undefined and not the defaultMessage if id is not found then resulting in empty headers.


eric-burel:
@SachaG  This behaviour of the `formatMessage` does not sound normal to me. The tests I wrote previously on this precise feature are also broken.
I think the issue is not Material UI but a breaking change in the intl API

eric-burel:
I'll merge though because those changes are also legitimate when columnName is undefined

eric-burel:
Thanks @Julien-Sytadelle !

PR #2281, opened 2019-04-17
add consideration to default values on accounts extra fields

yairtal:
This will respect defaultValue on extraFields in AccountsLoginForm, even with no onChange event triggered, and will allow hidden fields when needed

Should also fix https://github.com/VulcanJS/Vulcan/issues/2258


SachaG:
Thanks! But please make the PR to `devel` so I can merge it properly :)

yairtal:
see here: https://github.com/VulcanJS/Vulcan/pull/2282

PR #2282, opened 2019-04-18
add consideration to default values on accounts extra fields

### LINKED ISSUE #2258: Respect "email" if set when creating users
If the email is provided on User creation, it seems that it is still overriden. We should respect the email field if it is provided by the user.
(Note: this does not concern the `emails`, array which is tied to Meteor but the `email` field)
https://github.com/VulcanJS/Vulcan/blob/devel/packages/vulcan-users/lib/modules/schema.js#L153

@juliensl: also an interesting first issue

yairtal:
This will respect defaultValue on extraFields in AccountsLoginForm, even with no onChange event triggered, and will allow hidden fields when needed

Should also fix #2258


SachaG:
Thanks!

PR #2284, opened 2019-04-19
Add storybook, few fixes on FormError + Intl messages

eric-burel:
I got rid of the FormMessage regression regarding the defaultMessage prop. I also made it possible to return global form error without tying them to a specific field name.

And more importantly...I've added Storybook to the core. Bootstrap is used as the default UI lib. It works fine, except that you have to rerun `populateComponent` in each file after you imported your component, in order to have a correctly defined `Components`. 
We would have to find a way to run `populateComponent` after story are imported but before they are run to improve this, a bit like `Meteor.startup` does.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,12 @@
+{
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "shippedProposals": true,
+        "useBuiltIns": "usage"
+      }
+    ],
+    "@babel/preset-react"
+  ]
+}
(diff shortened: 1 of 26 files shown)


enzolutions:
@eric-burel we will have to change anything in the Material UI implementation?, how we could test the Storybook implementation?

eric-burel:
Hi @enzolutions, the draft of the install doc is here: https://github.com/VulcanJS/vulcan-docs/blob/master/source/storybook.md

To test Material UI, you can uncomment the corresponding Decorator in `.storybook/config.js` and comment the Bootstrap decorator. It will work as long as the correct npm packages are present in your app (so not in Vulcan or Vulcan Starter as-is but it should work in your own app).

No change is needed, it works out-of-the-box. Just keep it mind that it's experimental and the Meteor mocks may need to be improved in order for certain components to work correctly.

PR #2285, opened 2019-04-22
Fix awesome-vulcan requirement

justinr1234:


eric-burel:
Hi,
The request should already be provided by the `sink` object as `sink.request`, see usage example here: https://github.com/ErikDakoda/vulcan-material-ui/blob/830b47376573beb873166faa783d1b24d34b010c/server/wrapWithMuiTheme.jsx#L33

What was your use case? Do you need this additional field for legacy code?

justinr1234:
@eric-burel what branch is this on? The main branch is using `req`

eric-burel:
It's in the currently open PR actually not in a branch, I don't have access to Erik Dakoda's repo

justinr1234:
@eric-burel got it, that’s where my confusion was! Any chance we can ping him via email or otherwise to get it merged?

eric-burel:
Not sure sorry he seems quite busy :/ Also this PR was voluntarily left unmerged I think because it is only compatible with the Apollo 2 version of Vulcan so it implies maintaining 2 versions until enough people have migrated their apps

PR #2286, opened 2019-04-23
[WIP] fix MUI inputProperties handling

eric-burel:
Hi,
As suggested by @justinr1234 we may have regression in Material UI forms in the current implementation. `inputProperties` are taken into account, but they overshadow props specific to Material UI inputs.

I tried to provide a solution here, @justinr1234 what's your opinon on this?

I've had some troubles with `inputProps` prop of Material UI, because it seems to create unexpected behaviour. Eg if `inputProps` is defined, then the top level `onChange` seems to be ignored.

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -4,11 +4,11 @@
 import { addDecorator, configure } from '@storybook/react';
 
 // init UI using a Decorator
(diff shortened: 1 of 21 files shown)


justinr1234:
@eric-burel looks good

eric-burel:
Thanks I'll merge as soon as I can test it a bit more

PR #2289, opened 2019-04-29
Update FormNestedObject.jsx

juliensl:
hasErrors is the number of errors.
It should be a number, shouldn't it ?


eric-burel:
Thanks for the PR. Can you make it an "error or boolean" instead (PropTypes.oneOf I think)? That would guarantee backward compatibility. 

I think in the future we should explicitely have a "nbErrors" prop instead because it's confusing.

juliensl:
It's done.
Could you verify and add it please ? (I changed FormNestedArrayLayout too)

PR #2295, opened 2019-05-06
inputProperties function parameters bug

ErikDakoda:
When the schema defines inputProperties.disabled as a function, it's already evaluated to a boolean by this point. However props.inputProperties still has it as a function. Having ...props.inputProperties at the end of the spread overwrites the boolean with the function, resulting in an error.

### DIFF
--- a/packages/vulcan-lib/lib/modules/ui_utils.js
+++ b/packages/vulcan-lib/lib/modules/ui_utils.js
@@ -9,6 +9,7 @@ export const getHtmlInputProps = props => {
   // these properties are whitelisted so that they can be safely passed to the actual form input
   // and avoid https://facebook.github.io/react/warnings/unknown-prop.html warnings
   const inputProperties = {
+    ...props.inputProperties,
     name,
     path,
     options,
@@ -17,7 +18,6 @@ export const getHtmlInputProps = props => {
     onBlur,
     value,
     disabled,
-    ...props.inputProperties,
   };
 
   return {


eric-burel:
Sorry if there are issues I've had hard time wrapping my head around this part of the code. Isn't there an underlying issue? `inputProperties.hidden` should be either a boolean too, or should not even be defined?

ErikDakoda:
I apologize, I actually meant inputProperties.disabled, which can be a boolean or a function. Here is an abbreviated field schema from my code which includes a function for both `disabled` and `help`:

```
  platformAppShop: {
    type: String,
    label: 'Shop name',
    . . .
    inputProperties: {
      disabled: function ({ document }) {
        if (!document) return false;
        return !!document.hasPlatformAppToken;
      },
      help: function ({ document }) {
        if (!document) return false;
        return !!document.hasPlatformAppToken ? 
          'To change the Shop name, first click “Uninstall app” below' :
          'Enter your Shopify shop’s name or the web address of its admin';
      },
    },
    . . .
  },
```

eric-burel:
Hi, Sacha added you as a contributor, feel free to merge PRs that you feel relevant especially on Material UI. 
Right now I have issues with props passing, similar to the one you fix here. For example, props tends to be blacklisted, eg by cleanProps, which is hard to maintain because we progressively add more features to form. 
Instead I think we could adopt the reverse pattern like we tried to do in getHtmlInputProps(), ie to limit strictly props passed to end-input components to the actually possible props (valid html props for html inputs or valid Material UI Props for the relevant components).

ErikDakoda:
Thanks for the info. Moving to that reverse pattern does make sense. I'll keep that in mind.

PR #2297, opened 2019-05-06
MaterialUiFormGroupLayouts

ErikDakoda:
Material UI replacements for FormGroupLayout and FormGroupHeader (3 different styles). Also includes minor bug fixes in MuiSuggest.

vulcan-material-ui groups support 3 additional features: group.startComponent, group.endComponent, and an anchor tag at the top of each group to enable scrolling to it. I would be happy to move these features out of here and up to FormGroup.jsx so that they become standard Vulcan features.

This pull request depends on https://github.com/VulcanJS/Vulcan/pull/2296


eric-burel:
Sounds awesome to me. Could we call them "beforeComponent" and "afterComponent" though, like we do for inputs ? I've added their support for arrays too a few days ago, so you can pass them both at the top/bottom of an array of nested objects + at the top/bottom of the form for each nested object.

eric-burel:
Also I don't know if you tried it yet but Storybook is installed in the core, so you can easily visualize your components. It works with Material UI too.

ErikDakoda:
@eric-burel I renamed those fields and moved them to FormGroup.jsx.

I am very excited to try out your Storybook setup! It must have been tricky to get it working with Meteor. Kudos! Is there a quick getting started guide so I don't fumble around?

eric-burel:
`npm run storybook` in your Vulcan install ;) The only "tricky" point is that you sometimes may need to improve the existing Meteor mocks and you can't put stories anywhere. Components object also needs to be initialized after imports (there are examples in the code).

The full doc to install it in your own app is here: https://github.com/VulcanJS/vulcan-docs/blob/master/source/storybook.md.

PR #2298, opened 2019-05-07
material-ui : i18n files were not imported

Julien-Sytadelle:


eric-burel:
I am not sure we should load them systematically, except maybe for english in order to get a fallback. I think the correct pattern would be to require those files in your own app, so that they are only loaded if necessary. Not 100% sure how it's handled in my own apps though, I may have misunderstood the issue you are trying to solve.

ErikDakoda:
I have been thinking about how do deal with this too. I guess en_US should be loaded by default as the fallback, though I usually use `api.addFiles('en_US.js', ['client', 'server']);` in package.js.

Don't know if we have a standard way to load other languages dynamically.

Julien-Sytadelle:
Hi, search input from Datatable uses it. At least we should have the English version, but why not both?
I am just trying to solve a warning about missing traduction from search input. 

Le 7 mai 2019 à 18:40, à 18:40, Eric Burel <notifications@github.com> a écrit:

ErikDakoda:
Yes, I have the same bug with the search input. I say we might as well include both EN and FR, since they are quite small. But they should be added to package.js for consistency:
```
  api.addFiles([
    'accounts.css', 
    'forms.css',
    'en_US.js',
    'fr_FR.js',
  ], ['client', 'server']);
```

Julien-Sytadelle:
OK, I'll do that tomorrow. 

Le 7 mai 2019 à 19:20, à 19:20, Erik Dakoda <notifications@github.com> a écrit:

PR #2299, opened 2019-05-07
Fixed MuiCheckBoxGroup

EloyID:
With the commit before, all check box were selected from the beginning and couldn't be unselected. This is fixed.

### DIFF
--- a/packages/vulcan-ui-material/lib/components/forms/base-controls/MuiCheckboxGroup.jsx
+++ b/packages/vulcan-ui-material/lib/components/forms/base-controls/MuiCheckboxGroup.jsx
@@ -77,8 +77,9 @@ const MuiCheckboxGroup = createReactClass({
   
   changeCheckbox: function () {
     const value = [];
-    this.props.options.forEach(function (option, key) {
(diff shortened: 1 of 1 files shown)


packages/vulcan-ui-material/lib/components/forms/base-controls/MuiCheckboxGroup.jsx
```diff
@@ -94,21 +95,22 @@ const MuiCheckboxGroup = createReactClass({
   },
   
   renderElement: function () {
-    const controls = this.props.options.map((checkbox, key) => {
-      let value = checkbox.value;
-      let checked = (this.props.value.indexOf(value) !== -1);
-      let disabled = checkbox.disabled || this.props.disabled;
+    const { name, options, disabled: _disabled, value:_values } = this.props.inputProperties;
```
eric-burel: Why are those values imported from this.props.inputProperties instead of this.props? It seems that they were coming from inputProperties before

eric-burel:
Fixed in #2300 , thanks!

PR #2301, opened 2019-05-07
add visibleItemIndex function/prop in FormNestedArray

juliensl:
Hey !

Currently we have an *itemIndex* which is use to identify each items modify them.
If we want to use *itemIndex* in our beforeComponent to have an Array text like:
" TEAM
     Member 1
     Member 2
     Member 3
     Member 4
     Member 5
     Member 6
"
We will have some bugs when we will delete some items.
Like if I delete: 2 and 4, and add 2 more, I'll have:

" TEAM
     Member 1
     Member 3
     Member 5
     Member 6
     Member 7
     Member 8
"

So, I'm adding a *visibleItemIndex* which should solve this problem.

Now, it's working when I delete the lasts members and add new, but It still have some problems when I delete some inner, like I can have:

" TEAM
     Member 1
     Member 3
     Member 5
     Member 4
     Member 5
     Member 6
"

I think it's just one little think but I didn't find, maybe someone can improve this commit and merge it.

### DIFF
--- a/packages/vulcan-forms/lib/components/FormNestedArray.jsx
+++ b/packages/vulcan-forms/lib/components/FormNestedArray.jsx
@@ -5,9 +5,7 @@ import _omit from 'lodash/omit';
 
 // Replaceable layout, default implementation
 const FormNestedArrayLayout = ({ hasErrors, label, content, children }) => (
-  <div
-    className={`form-group row form-nested ${hasErrors ? 'input-error' : ''}`}
-  >
+  <div className={`form-group row form-nested ${hasErrors ? 'input-error' : ''}`}>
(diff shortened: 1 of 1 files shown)


eric-burel:
I am still not sure why you can't pass it down to the FormNestedLayout. I'll try that

eric-burel:
I've merged for the moment because it seem to work ok. You may try to run `npm run storybook` in the Vulcan folder : you will see a "FormNestedArrayLayout" story on storybook that works.

If it behaves badly when interacting with the form it may be due to another issue because the code seems fine to me and behave as expected, then we may open another issue/PR. I don't know how `instancianteComponent` works for instance.

PR #2305, opened 2019-05-09
FormNestedArray Footer modification via props fixing

EloyID:
With this modification we can pass a custom footer for the array forms.


eric-burel:
Perfect thanks for the fix

PR #2306, opened 2019-05-09
FormComponentInner calling an inexisting element (FormNested) 

EloyID:
When passing a custom Input for an array in a schema it crashes because it calls an inexisting element FormNested. I just erased this if et it will just use the FormInput


eric-burel:
Thanks, we will need to validate that on arrays of values though before merging, to see if it leads to any breaking change, because this part of the code looks suspicious to me (though your PR is probably correct).

eric-burel:
I'll PR now because this seems legitimate, this component is literally non existant, it is now probabily unfied witht usual FormComponent.

PR #2313, opened 2019-05-23
ResetStoreCallbackNotBeingRun

ErikDakoda:
- In some instances, for example post sign-up, the `then` clause is never called. Apollo Client now has the `onResetStore` function to solve this.

- `resetStoreAndThen` and `postLogInAndThen` essentially both just call `resetStore` and then the next hook. Using them both one after the other seems superfluous

- In `SignUp` I moved `self.setState({ waiting: false });` and `self.clearDefaultFieldValues();` higher to reduce the chance of React errors if the user is redirected in the `onPostSignUpHook` hook

### DIFF
--- a/packages/vulcan-accounts/imports/ui/components/LoginFormInner.jsx
+++ b/packages/vulcan-accounts/imports/ui/components/LoginFormInner.jsx
@@ -33,21 +33,27 @@ export class AccountsLoginFormInner extends TrackerComponent {
 
     const resetStoreAndThen = hook => {
       return () => {
-        props.client.resetStore().then(() => {
+        const resetStoreCallback = () => {
           hook();
-        });
+          removeResetStoreCallback(resetStoreCallback);
+        };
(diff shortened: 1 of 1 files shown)


eric-burel:
Awesome! I've had a bug when using the login form in a popup, when setting up oauth and I could'nt fix it. It was definitely due to resetStore never resolving. I will try your fix as soon as possible, tomorrow I think.
Related: https://github.com/apollographql/apollo-client/issues/2919#issuecomment-480310858

eric-burel:
My understanding is that with a `then` is not called sometimes because the component can be unmounted before the promise resolves, due to redirections. So with your solution since the callback is registered at the Apollo client level, now it runs correctly. Not sure though

eric-burel:
I've tested it, it works fine. Thanks to this we will now be able to do oauth using Vulcan apps.

ErikDakoda:
Awesome! It's always a good feeling to squash a pesky bug LOL! I'll go ahead and merge.

PR #2320, opened 2019-05-29
Head.jsx: added script rendering server side

Julien-Sytadelle:
It is useful if you use helmet for json-ld structured data


Julien-Sytadelle:
I close it as it duplicates the scripts.

PR #2322, opened 2019-05-29
FormWrapper is correctly reinitialized on documentId change

eric-burel:
Hi,

Currently there might be issues when changing a SmartForm from a "new" mode to an "edit" mode, and vice-versa. Even using 2 different SmartForm won't work. This is because in the `FormWrapper` component we rely on the constructor to define the component, which may sometimes be incorrectly shared across instances that are different.

Example of faulty code:
```
if (!document) return (
        <Components.SmartForm
          collection={projectChoicesCollection}
        />
    )
    return (
      <Components.SmartForm
            collection={projectChoicesCollection}
            documentId={document._id}
          />
```

React won't recreate a SmartForm instance in this case even if "document" is updated, probably for perf reasons, because it detects that you use the same component with different props. Thus, the `FormComponent` is not updated in the `FormWrapper` because the constructor does not run a second time.

I provide a quickfix (see the top of the files, other changes are only here to be able to use `nextProps` instead of `this.props`), however I'd like a review because this is not very clean. 
When adding a `documentId`, the Form will "blink" because the component is destroyed, and also show a warning `Warning: Can't perform a React state update on an unmounted component.`.

I think we can do better, but I don't know yet how. Maybe by using state instead of directly setting the component as a class variable React would handle the switch in a smoother way.

### DIFF
--- a/packages/vulcan-forms/lib/components/FormWrapper.jsx
+++ b/packages/vulcan-forms/lib/components/FormWrapper.jsx
@@ -57,7 +57,7 @@ class FormWrapper extends PureComponent {
   constructor(props) {
     super(props);
   }
-  componentWillMount(){
+  UNSAFE_componentWillMount(){
     // instantiate the wrapped component outside of render
     // see https://reactjs.org/docs/higher-order-components.html#dont-use-hocs-inside-the-render-method
     this.FormComponent = this.getComponent(this.props);
--- a/packages/vulcan-forms/lib/components/FormWrapper.jsx
+++ b/packages/vulcan-forms/lib/components/FormWrapper.jsx
@@ -53,44 +53,56 @@ import { callbackProps } from './propTypes';
 
 const intlSuffix = '_intl';
 
+const RESET_PROPS = ["documentId"]
+
 class FormWrapper extends PureComponent {
   constructor(props) {
     super(props);
-    // instantiate the wrapped component in constructor, not in render
+  }
+  componentWillMount(){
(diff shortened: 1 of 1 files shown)


eric-burel:
More globally the issue is that using HOCs dynamically is kind of an anti-pattern, because HOCs are mostly meant to be used statically, outside of React component. That's also what makes `withMulti` hard to grasp as discussed previously, and we have a similar issue in DataTable.

I think we would greatly benefit from a small refactor of the with* hocs. We could write additional React components like `<SingleQuery />`, `<MultiQuery />`, based on Apollo's `<Query />` instead of the `graphql` HOC, and use them to rewrite `withSingle` etc. 
This way we can have 2 pattern, the usual "static" HOC (ideal for providing data to a whole page for example), and a more dynamic usage with actual React component, that would be suited here for the FormWrapper and all complex components that manipulates data.

ErikDakoda:
This article about React hooks in Apollo client for GraphQL caught my eye.

https://graphqlmastery.com/blog/react-hooks-in-apollo-client-for-graphql-queries-and-mutations

We do need to think about what the best pattern will be for supporting the new Apollo Client 2.1-2.5 features like Optimistic UI, Local state, Updating the cache, etc. Or have you been able to use these features with our existing HOCs?

eric-burel:
Hooks are now merged, so we will solve this by cleaning up components.

I did not use Apollo advanced feature because we first needed a solid and up to date basis but when we get rid of those long lasting issues we can focus on perf, caching etc.

PR #2323, opened 2019-05-31
Support GraphQL schema generation for nested objects/arrays

eric-burel:
This PR aims at solving the issues [with dates in nested objects](https://github.com/VulcanJS/Vulcan/issues/2211), due to the GraphQL schema generating blackbox JSON type for those objects.

For the moment I just splitted the code to make the GraphQL schema generation readable/testable, it's still under progress. That's a feature I will need soon so I hope it will progress quickly.

It seems that there are 2 steps:
- generating the correct types
- but also the correct fragments

Not sure yet if it will lead to breaking changes or not, maybe people will need to update their fragments with nested objects (not a big deal at all though).
Codewise it's non trivial because it means generating multiple types per field (one field + N nested objects = N+1 types in the schema)

Todo list for the main type:
- [ ] take "blackbox" option into account => did not work, it seems that `SimpleSchema` scraps out the `blackbox` field, I don't know why. Anyway using `blackbox` with a custom SimpleSchema does not make much sense so I guess it's ok.
- [x] in the main schema, get the correct typeName for the nested field: `JSON` for blackbox and `MainTypeNestedType` (eg `FoobarAddress` for an `address` field) for actual nested Object
- [x] handle arrays. We need to remove the `$` from the nested field name.
- [x] define what happens in the schema. To handle nested objects, we need to give them a new GraphQL type.
- [x] Actually generate the nested schema GraphQL type for nested objects
- [X] generate the schema for arrays of nested objects too
- [x] generate valid default fragment
- [X] check what needs to be improved in the `graphql_template` file. => Nothin', just added a template for enums.
- [x] while we're at it, support "enum" types generation (see #2332) based on `allowedValues` for String fields. Implementation would be similar. Note: we should ignore allowedValues for more complex objects though (objects, arrays), graphql enum are limited to string (in JS at least) as far as I understand => works ok for strings
-  [X] Do not generate enums when allowedValues contains forbidden chars, instead simply print a warning. This is necesary because enums must be normalized strings (no special chars, no -, only letters, numbers and _) according to the GraphQL spec. Anyway this is not a good practice to have complex strings as allowedValues.
- [ ] define a behaviour for permissions. Currently you'll need to define canRead etc. for all subfields too in order for the schema to be generated (I don't know if we checked them in the previous implementation or if we relied on the parent permissions)
- [x] generate fragment for the nested types
- [x] generate correct input types for nested objects and use them in templates. We can't simply reuse the type, we have to create an input type for each nested field too.
- [x] generate input types for possible operations for nested objects. Currently we generate only one input type `FooNestedFieldInput` for the field `nestedField` of the collection `Foo`. This prevents us from handling `canCreate` correctly for example, instead we should have `CreateFooNestedFieldInput `, `UpdateFooNestedFieldInput`...
- [ ] check casing of enums during normalization
- [ ] generate the correct fragment in `FormWrapper`. Look for `FormFragment` code (eg FooEditFormFragment). Work is ongoing. I first write an independant  `formFragments` function that generates the fragments, then I'll refactor `FormWrapper` to use this function.

Relevant files:
- `vulcan:lib lib/modules/graphql/schemaFields`: generate relevant fields. Here we match the schema types to actual GraphQL types. We probably will need to reuse this function to generate the nested object type too, and maybe improve it a bit.
- `collectionToGraphQL`: here we generate the whole schema. We will need to modify it to take nesting into account.

Existential questions:
- [ ] What we are building here are `composition` in UML language, meaning that the subschema is strongly tied to the main schema, it can't exists without it. This means that the subschema type is namespaced with the main schema (`Event` with `Adress` will give an `EventAdress` subtype).
- [ ] What about `aggregations` ? For example, if we have a reusable `address` schema, that is not tied to a specific collection, the concept of "adress" still exists as its own: this is an aggregation (note: if it had its own collection it would be an `association`, one-to-one, one-to-many and so on, that would be handled using a resolver).
My current solution will generate compositions only, which means multiple `adress`-like types, one for each schema: `EventAdress`, `PersonAdress`, instead of generating only one `Adress` schema. This is because there is no global generation step that takes all schemas in consideration, so we can't detect such redundancies. Plus that depends on the final use case. This is more showerthoughts than a real problem but handling relation is challenging.

Expected behaviour would be:
```
const foobarSchema = {
   bar : { type: Number },
   nested : new SimpleSchema({
      zed: { type: String }
   })
}
```
===>>>
```
type Foobar {
    bar: Number,
    nested: FoobarNested
}
type FoobarNested {
   zed: String
}
```


ErikDakoda:
This is great - I have been meaning to implement this for a while.

eric-burel:
Done, so now:
- nested objects will spawn a new type in the graphQL schema
- same for objects within nested objects
- same for enum types either in the main schema or a nested schema
- default fragment will work for nested objects

**This is a breaking change but only if you use custom fragments**, you will now need to explitely list the fields you need for nested object. I did not generate an additionnal fragment for those nested fields yet.

The function that generates types for schema fields now detects Objects, Arrays, Enums, and generate an additional types for those object. Then the function that build the whole graphQL schema add those additionnal types in the schema.
Basically the new features is supporting outputing not only one type per field of the schema but multiple types.

This works with one level of recursion for the moment. We don't support resolveAs for the nested fields yet but I think it would not be hard to implement.
I did not manage to support blackbox option because SimpleSchema has a crappy way to handle them: it stores a `_blackboxKeys` array at the root of the schema. So at the field level you lose this information. It's perfectly doable to fix this though, it's just a bit more code.

I've added relevant unit tests too so we are now safer :)

Tagging @SachaG for review.

eric-burel:
We noticed a bug, enum types don't tolerate special characters like French cedilla (ç). I'll try to fix it.

Apollinaire:
I started testing on my side too - this looks very promising and I have long awaited that feature!
Here are my propositions to answer a few of the pending questions from above: 

#### Reusable schemas / aggregation

I think this is the most problematic thing at the current state of this PR. It bloats the graphql schema, and is not in sync with the best practices of graphql. It's more TreeQL than GraphQL actualy.
Consider the following schema:
```js
import TypeSchema from './TypeSchema';

const schema = {
  _id: {
    type: String,
    optional: true,
    canRead: ['guests'],
  },
  fieldOne: {
    type: TypeSchema,
    canRead: ['guests'],
    canCreate: ['admin'],
  },
  fieldTwo: {
    type: TypeSchema,
    canRead: ['guests'],
    canCreate: ['admin'],
    canUpdate: ['admin']
  },
  fieldThree: {
    type: Array,
    canRead: ['guests'],
  },
  'fieldThree.$': {
    type: TypeSchema,
  },
};

export default schema;
```

Currently, this defines the following schema: 
image

Ideally, this would be the schema generated:

62821232-11c06c80-bb72-11e9-84f0-fdc2a9c76694-modif

To achieve this in regular GraphQL apps, you define all your types (=nodes) and their fields (=edges, or refs to other types). Also, you attach a resolver to the type and to every field.

In order to do the same thing within Vulcan, here is my suggestion: 
- create a `createGraphQLType` function that takes a sweetSchema and resolvers (for queries and mutations) and builds the type with its resolvers. Inside `createCollection` we can use this function too.
- To reference a registered type inside a schema, we could do something like :
```
const schema = {
  nestedField: {
    type: getType('TypeName'),
    optional: true,
    canRead: ['guests']
}
```
What remains unknown is how to make this work with SimpleSchema, but I'm confident that it can work.
IMO this is the only way for us to do real GraphQL.

#### Permissions

Looking at the proposal above, I think it would be logical to add the permissions inside every subfield.

I will create a branch from this commit and start working on the implementation

Apollinaire:
About the enums: I think it's a bad idea to generate an enum for the `allowedValues` field, because it mixes graphQL and validation. `allowedValues` is a field interpreted by SimpleSchema : https://www.npmjs.com/package/simpl-schema#allowedvalues

eric-burel:
Great suggestions! 
Indeed we have to fight with Simple Schema currently, which complicates everything. I agree that we should be able to have a specific field to handle enums instead of allowedValues, all the more that it creates issues with non-normalized allowed values (accents, uppercase etc.), but I did not manage to create such field, it was somehow dumped after calling `new SimpleSchema`.

Basically the JSON schema should only be transformed into a SimpleSchema when necessary, for validation, and the JSON should be kept as is otherwise. Then we'll be able to define our own syntax that works better with GraphQL. But that means changes in a few places.

Apollinaire:
I believe we can work with SimpleSchema. I like the fact that it normalizes the schema, so we can rely on it to have well formated schemas all the time. But indeed it's extra work to get the original fields.
Also, it would be quite a big regression to remove the fact that people can pass a SimpleSchema in their collection.
For the extra fields, you have to declare them into SimpleSchema here: https://github.com/VulcanJS/Vulcan/blob/devel/packages/vulcan-lib/lib/modules/config.js

eric-burel:
Just to reference another issue related to nesting: https://github.com/VulcanJS/Vulcan/issues/2178

PR #2343, opened 2019-06-25
Fix SSR issues with react-apollo and style wrappers

eric-burel:
Fixes https://github.com/VulcanJS/Vulcan/issues/2338
Fix based on:https://github.com/mui-org/material-ui/issues/10522

It seems that React tree traversal triggers side effects. For example, when using Material UI, it will generate className.

During SSR, 2 steps needs a traversal of the tree:
- graphQL data generation with `react-apollo`
- and the actual render with ReactDOM/server.

In the current implementation, they were done in one call to `renderToStringWithData` from `react-apollo` **which I advise to never use** (won a line of code, lost hours of debugging... one function = one side effect and one goal, not 2).

I split those steps. 
Data fetching is now done on the unwrapped `App` component, before `router.server.wrapper` is run. It will find `Query` components and fill Apollo's store accordingly.
 
Rendering is done on the whole app, including wrappers registered with `router.server.wrapper`, so the Material UI theme wrapper, Styled Component wrapper, Redux store provider and so one.

Only edge case I can imagine is having queries that depends on a wrapper component. Maybe with Redux? If this happens the solution will be to create a new callback, differentiating wrapper that needs top be included while fetching data and style wrappers.

### DIFF
--- a/packages/vulcan-lib/lib/server/apollo-ssr/renderPage.js
+++ b/packages/vulcan-lib/lib/server/apollo-ssr/renderPage.js
@@ -6,7 +6,7 @@
  */
 import React from 'react';
 import ReactDOM from 'react-dom/server';
-import { renderToStringWithData } from 'react-apollo';
+import { getDataFromTree } from 'react-apollo';
 
 import { runCallbacks } from '../../modules/callbacks';
 import { createClient } from './apolloClient';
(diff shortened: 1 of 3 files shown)


justinr1234:
@eric-burel if redux is broken, you have a fix?

eric-burel:
It would be doable to use a different callback for wrapper like Redux, that would be run before `router.server.wrapper` and included in the `getDataFromTree` call. But I wonder if it's really worth it or not, I don't have example to experiment this.

eric-burel:
I found a non breaking solution, afterthought the best is to create 2 callbacks instead of one, will update the PR later

eric-burel:
I've created an additional `router.server.styleWrapper` callback to handle style related libraries, and kept the `router.server.wrapper` for other tools like Redux. So that should not create regressions.

eric-burel:
Ok now I am actually done. Will be testing for a few days and then I will merge. There are now 2 new callbacks : `router.server.dataWrapper` and `router.server.renderWrapper` (sorry for the names but I can't think about something clearer).

The first one applies wrapper ONLY during the Apollo `getDataFromTree` call. The second ONLY during the `renderToString`. This allow to have correct behaviour during each React tree traversal, that previously generated unexpected side effects that broke SSR.

`router.server.wrapper` is still available and should be used for other libs that don't trigger side effects during rendering, such as adding a Redux provider server side etc.

PR #2346, opened 2019-06-28
MuiSuggestShowAllOptions

ErikDakoda:
Fixed bug: MuiSuggest would not show all options even when showAllOptions property was set


eric-burel:
Hi, I think you can merge those minor fixes on Material UI if they work for you. I'll be able to test them later when updating the Live for Good app, I'll be more active this month.

ErikDakoda:
Hi @eric-burel, thank you for chiming in. I will go ahead and merge my pull requests. Do you mind if I go ahead and close some really old and stale pull requests so we have a better idea of what's really outstanding?

eric-burel:
Yes there are a few of them that are very old indeed. @SachaG might confirme now he is done modelling for State of CSS t-shirts :)

PR #2348, opened 2019-07-01
EJSONStackOverflow

ErikDakoda:
Fixed bug: EJSON.stringify() will throw a stack overflow error when attempting to stringify an object with circular references


eric-burel:
Hi, shouldn't a warning be thrown? When does this happen?

PR #2349, opened 2019-07-02
Unit Testing Update

ErikDakoda:
* Got client unit tests running
 * Upgraded to Meteor 1.8.1
 * Added a custom version of meteortesting:mocha package - the only way to get it to use meteortesting:mocha-core@6.1.2, meteortesting:browser-tests@1.2.0, and lmieulet:meteor-coverage@3.2.0
 * Updated some client unit tests to get them to pass
 * Made experimental change to mutators.test.js - client tests would not run when the Foo2s collection is initialized inside before()

PR #2350, opened 2019-07-06
Redux tests for setupRedux(initialState)

### LINKED ISSUE #2351: vulcan:redux does not handle initialState of createStore
When you initialize redux with vulcan:redux, the initialState is always empty.
PR #2350 add initialState in setupRedux and tests covering the feature

lc3t35:
this should close #2351

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -16,3 +16,5 @@ accounts-password@1.5.1
 vulcan:debug
 meteortesting:mocha
 apollo
+vulcan:lib
+vulcan:ui-bootstrap
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -88,7 +88,7 @@ vulcan:email@1.13.0
 vulcan:i18n@1.13.0
 vulcan:i18n-en-us@1.13.0
 vulcan:lib@1.13.0
-vulcan:routing@1.13.0
+vulcan:ui-bootstrap@1.13.0
 vulcan:users@1.13.0
 webapp@1.7.0
 webapp-hashing@1.0.9
--- a/package-lock.json
+++ b/package-lock.json
@@ -1142,6 +1142,64 @@
         }
       }
     },
+    "@emotion/babel-utils": {
+      "version": "0.6.10",
+      "resolved": "https://registry.npmjs.org/@emotion/babel-utils/-/babel-utils-0.6.10.tgz",
+      "integrity": "sha512-/fnkM/LTEp3jKe++T0KyTszVGWNKPNOUJfjNKLO17BzQ6QPxgbg3whayom1Qr2oLFH3V92tDymU+dT5q676uow==",
+      "dev": true,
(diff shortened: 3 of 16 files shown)


GraemeFulton:
I've tested this in my own app, works great - [see here for guide](https://www.notion.so/Redux-for-Vulcan-b21807a78b574bee8bb6e93360165c2d).

lc3t35:
@GraemeFulton great documentation, could you add a few lines about the initialState parameter in setupRedux ?

eric-burel:
Awesome, glad you got it working. Does it provide a sufficient fix for your apps in the short run?

Regarding tests Erik Dakoda pushed a PR to improve testing and allow running client side tests. I started reviewing #2349 but I need a bit more time because some tests were not triggered as expected. 

Last time I checked vulcan:lib tests were passing but Meteor testing works in mysterious ways.

Can the Notion page be converted into markdown so we can add it to the official doc?

GraemeFulton:
> @GraemeFulton great documentation, could you add a few lines about the initialState parameter in setupRedux ?

I've added the info about initialState parameter today, thanks again for this fix!

eric-burel:
Hi, tests are passing fine for me, @GraemeFulton @lc3t35 is it fine to merge for you? Does it solve your issues with Redux?

GraemeFulton:
hey it does solve my issue @eric-burel !

yairtal:
I solves my issues as well, is there anything I can do to help get this merged?

kevinashworth:
I use Redux. This change is very helpful to me, it's one that I keep pulling into my local Vulcan repo again and again. The 2 conflicting files can just be deleted. (I delete them after after every merge.) I'd like to see this PR merged into Vulcan.

eric-burel:
Hi,

Do you think it would doable to avoid the explicit `setupRedux` call? Basically what bothers me is that you can't add values to the `initialState` from another package.

Instead, I'd rather have `setupRedux` be called automatically, maybe in a `Meteor.startup` function, and the initialState built from a callback, so that any package can add new parts to the initialState. 
It feels more "vulcanish" to me, in the sense that it would respect the package oriented architecture where any package can freely enhance the app.

That's why I did not merge it in the first place. However I am not sure what I expected would work, maybe it's an over-optimization. Would need @SachaG opinion on this.

eric-burel:
Smth like: 
```
let store = null
export const getStore = () => store
const setupRedux = () => {
  const initialState = runCallbacks({name: 'redux.initialState', iterator: initialState}) // this part allow a package to enhance the initialState implicitely 
  store = initStore();
  addCallback('router.server.wrapper', function ReduxStoreProvider(app) {
    return <Provider store={store}>{app}</Provider>;
  });

Meteor.startup(() => setupRedux())
```

lc3t35:
@eric-burel you mean `store = initStore(initialState);` ;)

PR #2352, opened 2019-07-09
Bugfix/mui ssr

eric-burel:
See https://github.com/VulcanJS/Vulcan/pull/2343 (reopned to update missing commits)

### DIFF
--- a/packages/vulcan-lib/lib/server/apollo-ssr/renderPage.js
+++ b/packages/vulcan-lib/lib/server/apollo-ssr/renderPage.js
@@ -6,7 +6,7 @@
  */
 import React from 'react';
 import ReactDOM from 'react-dom/server';
-import { renderToStringWithData } from 'react-apollo';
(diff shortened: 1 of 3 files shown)


ErikDakoda:
Hi @eric-burel, I am having SSR issues with the current devel branch, which only really become obvious when you build and deploy the code and the jss class names are minimized. I wonder if this PR will fix it. Is this close to being ready to merge? Or do you know how I can apply the changes in this PR into my own dev code?

eric-burel:
Hi @ErikDakoda ,
We are currently battle testing this feature at Live for Good, it seems to work as expected and reduced the numer of SSR issues (some issues still happens with data oriented component such as the SmartForm, but it's due to the way they are coded not to this issue). 

I am not eager to merge it quickly as those kind of changes sometimes create unexpected troubles that we only detect after a few weeks, but it should work already. You can merge it in your Vulcan install to test it, there is 99% chances that it will be merged in devel soon.

PR #2354, opened 2019-07-23
StorybookAndMaterialUi

ErikDakoda:
* Added `STORYBOOK_UI` environment variable to select which Vulcan UI package to use when starting Storybook (instead of having to uncomment code)
 * Added `storybook-material` script in package.js to launch Storybook using the Material UI package
 * Modified `build-storybook` script to build Storybook stories for both Bootstrap and Material UI and save them to `docs/storybook-bootstrap` and `docs/storybook-material` respectively so that they can be published using GitHub Pages
 * Updated vulcan:ui-material, including the new `TooltipButton` component which replaces both `TooltipIntl` and `TooltipIconButton`


eric-burel:
Thanks a lot, I am reviewing the PR.
Currently I stumble on this issue with Babel/core-js: https://github.com/coinbase/rest-hooks/issues/25

eric-burel:
Awesome PR! 
I have merged, my remarks are:
- I removed the "docs" folder. I don't know if it's best to keep it or not but I am not fond of having autogenerated stuffs to be stored in the github. That would be preferable to push it to a github pages or something from time to time instead.
- Being finnicky, we could try to avoid loading stories containing "material" when bootstrap is selected, by modifying the regex used in `config.js`. We could have special "bootstrap.stories.js" and "material.stories.js" files for this usage
- I locked version to 5.0.x. Further versions of Storybook use `core-js@3.X` and that's a HUGE mess with versions. I did not know core-js existed only a few weeks ago but it has been my nemesis lately, it breaks up every updates of every libs... We will have to avoid updating anything related to Babel for a while until it stabilizes

Otherwise the STORYBOOK_UI env variable with Bootstrap as a default is awesome

ErikDakoda:
@eric-burel I am happy to collaborate with you - this stuff is not straight forward and two heads are better than one - though you did most of the heavy lifting. Thank you!

> I removed the "docs" folder. I don't know if it's best to keep it or not but I am not fond of having autogenerated stuffs to be stored in the github. That would be preferable to push it to a github pages or something from time to time instead.

That makes sense. What we should be doing is adding a build step to CircleCI to build storybook and deploy it to GitHub Pages, or Netlify, so that we can link to life, functioning stories from the Vulcan Docs.

> Being finnicky, we could try to avoid loading stories containing "material" when bootstrap is selected, by modifying the regex used in config.js. We could have special "bootstrap.stories.js" and "material.stories.js" files for this usage

You are right, but I would add a third regex pattern: "vulcan.stories.js", which would be build for both bootstrap and material. If fact the majority of the components should be common between the two and should function the same betweetween the two UIs - although they will look different. So for Bootstrap we would build anything that matches `*.vulcan.stories.js` and `*.bootstrap.stories.js`, and for Material UI we would build `*.vulcan.stories.js` and `*.material.stories.js`.

Again, thanks for your efforts in getting Storybook working with Meteor and Vulcan - very useful, especially once we can publish the stories somewhere online to enhance our Docs.

eric-burel:
Thanks! Hopefully Sacha initiated this otherwise I was going to rewrite an offbrand Storybook directly in Vulcan instead on heading toward webpack lol

PR #2357, opened 2019-08-05
Bugfix/button props

juliensl:
Hi everyone, 

Since a time, I've React warnings to change size of buttons from "small" to "sm".

I searched them on Vulcan and changed them, but there are 2 fun facts:

1. On the material-ui button API doc , it's written that the props should be "small, medium & large", for the version 3 and 4.
https://v3.material-ui.com/api/button/
https://material-ui.com/api/button/
2.  When I change the size in the file AccountsButton.jsx, I've a warning to change it from "sm" to "small" x)

So, I changed them in the 3 other files but not this one, and I've no more warnings.
Maybe you've more explanations.

Have a good day!

Julien.

### DIFF
--- a/packages/vulcan-forms/lib/components/FormNestedArrayLayout.jsx
+++ b/packages/vulcan-forms/lib/components/FormNestedArrayLayout.jsx
@@ -28,7 +28,7 @@ const FormNestedArrayLayout = (props) => {
         {
           addItem &&
 
-          <FormComponents.Button className="form-nested-button" size="small" variant="success" onClick={addItem}>
+          <FormComponents.Button className="form-nested-button" size="sm" variant="success" onClick={addItem}>
             <FormComponents.IconAdd height={12} width={12} />
           </FormComponents.Button>
         }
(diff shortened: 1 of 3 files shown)


ErikDakoda:
Hi @juliensl ! I had a look at this pull request, and I'm puzzled.

The changes to `FormNestedArrayLayout.jsx` and `FormNestedItem.jsx` in vulcan:forms look correct, because in the spec for the Vulcan `Button` component, valid `size` props are `sm` and `lg`, which matches the react-bootstrap spec: https://react-bootstrap.github.io/components/buttons/

However the changes to `Button.jsx` in vulcan:ui-material look wrong, because the spec for the `size` prop for buttons in Material UI is `enum: 'small' | 'medium' | 'large'`. The `Button.jsx` does the conversion from `sm` to `small`.

In fact I am not sure why you are getting warnings about this - I have not seen them in my Vulcan Material UI project. Which version of `@material-ui/core` do you have installed?

juliensl:
Hi @ErikDakoda 

You're right, I don't have any warning with "small | medium | large" in Button.jsx, my bad.
I cancelled this change and commit it, only `FormNestedArrayLayout.jsx` and `FormNestedItem.jsx` are modified now.

PR #2358, opened 2019-08-12
add from props

juliensl:
Hi there!

I think it could be a good idea to have the possibility to add a props "from" in the Vulcan email.

What do you think ?


ErikDakoda:
@juliensl This is not a bad idea, however, I would make the new `from` parameter of the `VulcanEmail.send ` function the **last** parameter, to preserve backwards compatibility.

juliensl:
Hi !

It's true, I changed the position of the "from".
I added it in the *budildAndSendHTML* too.

PR #2361, opened 2019-08-20
Handle Nested types the graphql way

Apollinaire:
Hi all,
This is a draft of the way we should handle nested types IMO. 
Here is a recap of how it works, as said in #2323 : 

Every time we create a collection with a schema, it registers the graphql schema through a function called `addTypeAndResolver` (name to change maybe). This function creates the type definition, input types and resolvers for the **flat** type. Nested types have to be registered directly through this function, and can be referenced thanks to another function `getType` (name to change because too generic). Also, each field resolver checks the `canRead` to make sure the field can be read by the user. One thing to get though, is that the `canRead` is tied to the type, so if you reuse the type at two places, then it will have the same schema.
You can see an example in `vulcan:users/modules/schema`, on the emails field.

This way, we can really handle nested types and deeply nested fields inside one document.
I have reworked the `Users` collection which had some kind of nesting in it (field emails, but not viewable).

This is just a draft. What remains to be done is : 

- [ ] find better names for the new functions
- [ ] rework / clean / optimize the code in vulcan:lib/lib/modules/graphql
- [ ] rework default fragment generation
- [ ] rework validation on mutation resolvers (will probably need to memo the schema for each type)
- [ ] Forms
- [ ] Datatable
- [ ] fix unit tests
- [ ] test it a lot

I will begin developing my side project on this branch, so I can test it.

Any feedback is welcome!

PS : This will be a breaking change, but it is necesary. I suggest we take advantage of this and remove all the backwards compatibility related to previous releases, for instance all related to the openCRUD change. So we might want to add warnings on every soon-to-be-deprecated feature to the next minor release.

### DIFF
--- a/packages/vulcan-lib/lib/modules/graphql/graphql.js
+++ b/packages/vulcan-lib/lib/modules/graphql/graphql.js
@@ -14,7 +14,7 @@ import { disableFragmentWarnings } from 'graphql-tag';
 
 import collectionToGraphQL from './collection';
 import { generateResolversFromSchema } from './resolvers';
-import { mainTypeTemplate } from '../graphql_templates';
+import { mainTypeTemplate, createDataInputTemplate, updateDataInputTemplate } from '../graphql_templates';
 import getSchemaFields from './schemaFields'
 
 disableFragmentWarnings();
@@ -127,11 +127,13 @@ export const GraphQLSchema = {
       throw Error('Error: trying to add type without typeName')
     }
 
-    const { fields, nestedFieldsList, schemaResolvers = [] } = getSchemaFields(schema._schema, typeName);
+    const { fields, nestedFieldsList, schemaResolvers = [],  } = getSchemaFields(schema._schema, typeName);
     mainType = fields.mainType;
     // generate a graphql type def from the simpleSchema 
     console.log('--------------------')
     console.log(typeName)
+    console.log(fields.create)
+    console.log(fields.update)
     // console.log(mainType)
     // console.log(fields)
     // console.log(nestedFieldsList)
@@ -140,10 +142,18 @@ export const GraphQLSchema = {
 
     // console.log(JSON.stringify(fields, null, 2))
     const mainGraphQLSchema = mainTypeTemplate({typeName, fields: mainType, description, interfaces });
-
     // console.log(mainGraphQLSchema)
     // add the type and its resolver
     this.addSchema(mainGraphQLSchema);
+    
+    // createTypeDataInput
+    if ((fields.create || []).length) {
+      this.addSchema(createDataInputTemplate({typeName, fields: fields.create}));
+    }
+    // updateTypeDataInput
+    if ((fields.update || []).length) {
+      this.addSchema(updateDataInputTemplate({typeName, fields: fields.update}));
+    }
     const resolvers = generateResolversFromSchema(schema)
     if(resolvers){
       this.addResolvers({[typeName]: resolvers});
@@ -181,9 +191,9 @@ export const GraphQLSchema = {
     } = getCollectionInfos(collection);
 
 
-    const { nestedFieldsList, fields, resolvers: schemaResolvers = [] } = getSchemaFields(schema._schema, typeName);
+    // const { nestedFieldsList, fields, resolvers: schemaResolvers = [] } = getSchemaFields(schema._schema, typeName);
 
-    addTypeAndResolvers({ typeName, schema, description, });
+    addTypeAndResolvers({ typeName, schema, description, interfaces });
 
     const {
       graphQLSchema,
--- a/packages/vulcan-lib/lib/modules/graphql/graphql.js
+++ b/packages/vulcan-lib/lib/modules/graphql/graphql.js
@@ -29,9 +29,7 @@ disableFragmentWarnings();
  */
 const getCollectionInfos = collection => {
   const collectionName = collection.options.collectionName;
-  const typeName = collection.typeName
-    ? collection.typeName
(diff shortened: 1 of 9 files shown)


packages/vulcan-lib/lib/modules/graphql/graphql.js
```diff
@@ -13,9 +13,37 @@ import Vulcan from '../config.js'; // used for global export
...
 
+/**
+ * Extract relevant collection information and set default values
+ * @param {*} collection
+ */
+const getCollectionInfos = collection => {
+  const collectionName = collection.options.collectionName;
+  const typeName = collection.typeName
+    ? collection.typeName
+    : Utils.camelToSpaces(_initial(collectionName).join('')); // default to posts -> Post
```
eric-burel: Initial is not defined here

eric-burel:
Just a quick answer because that feature is quite rich and complex.

Can't we have both behaviour? Eg if I specify "typeName", I can push an existing and reusable schema, and if I don't, I generate a oneshot nested schema for this collection using the preexisting code (maybe with a warning if it's not a best practice?) ? So aggregation or composition in the UML language. 
Because sometimes you define a nested schema on the fly, for a very specific case, that you don't intend to reuse that much, it's just to keep the schema clean. 
Also on a different level but that would mean less breaking changes, only the fragments would need an update, not the schemas, and the users can refactor progressively to unbloat the GraphQL schema.
I am not sure yet, because having too many patterns is also confusing. 

Anyway I think this feature will need long-run experimentation, and your proposal sounds good.

I'd simply rename  `getType` something more explicit like `getExistingSchema` or smth (I don't have a better name yet).

Regarding enums, what I missed is this `Schema.extendOptions` function. Do you think it would make sense to add an "enums" field for this use case?

packages/vulcan-lib/lib/modules/graphql/graphql.js
Apollinaire: Thanks for the review. Actually, collection.typename is always defined in createCollection (and createCollection throws an error if you don't pass in a typeName), so I'll just remove this piece of code. On the other hand, collectionName is not necesary.

Apollinaire:
I think we can handle both behaviour, but that will mean it's going to be harder to maintain. But indeed this is a good idea!
About the enums: I don't know if it's a good idea; I don't know how used they are, if we must cover them directly in the schema or if we can just let people do it themselves with `addGraphQLSchema`, as it's already the case.

eric-burel:
If we refactor it correctly it shouldn't be much additional work to support multiple ways of creating nested schemas because the underlying functions will be the very same, the only difference is the way we trigger them. 
Passing directly the `new SimpleSchema` in a field would be equivalent to a call to `addTypeAndResolvers` then `getType` on this field/schema.


What bothers me with `getType` is that the user needs to call it explicitly. Instead I'd rather keep it internal, we can call it inside `createGraphQLAPI` when the field have a `typeName` for example.

This how I imagine usage: 
Eg 
```
const foobarSchema = {
   // an e
   referenceNestedField: {
      nestedType: 'Whatever',
      ....
   },
   onTheFlyNestedField: {
       type: new SimpleSchema(...)
  },
  normalField: {
      type: String
  },
  enumField: {
     type: String,
     typeName: 'FooEnum'
   }
}

// createGraphQLAPI == what you call addTypeAndResolvers, it's just for name consistency with createCollection

addTypAndResolvers({ typeName: 'Whatever', schema: whateverSchema })
addToGraphQLSchema('enum FooEnum { FOO BAR }')
createCollection({graphQLTypeName: 'Foobar', schema: foobarSchema })
```

So here `Whatever` is detected to be a type associated with a canonical schema that will be automatically merged into the main schema.

But maybe that's too much magic? I am looking for a way to satisfy beginners that will look for a syntax similar to Simple Schema as well as advanced use cases with schema reuse.

Apollinaire:
If you don't like using `getType` (I'm not a big fan of it either) , you can do this and it would work exactly the same: 
```js
const foobarSchema = {
   referenceNestedField: {
      type: Object,
      blackbox: true,
      typeName: 'Whatever',
      ....
   },
   onTheFlyNestedField: {
       type: new SimpleSchema(...)
  },
  normalField: {
      type: String,
      typeName: 'String' // useless but works the same
  },
  enumField: {
     type: String,
     typeName: 'FooEnum'
   }
}

addTypeAndResolvers({ typeName: 'Whatever', schema: whateverSchema })
addToGraphQLSchema('enum FooEnum { FOO BAR }')
createCollection({graphQLTypeName: 'Foobar', schema: foobarSchema })
```
the `getType` was only here to automatically make sure that we don't get a validation error by SimpleSchema, by adding `type: Object, blackbox: true` (for mutation validation, when we see `typeName` on a field, we validate it through the schema registered with the typeName if there is one)

This whole PR is about supporting a `typeName` field on schemas to customize the GraphQL schema on a field basis. We could also add a `resolver` field to overwrite the resolver that injects data into the Types' resolvers, or fieldbased callbacks like `fieldResolvers.MyType.myField` that can validate, affect the field.

eric-burel:
This is merged into `feature/nested-schemas`, I reenabled the one-time generation of a schema for the moment because this reduces breaking changes to a minimum. I am pretty sure the code can be factored with `addTypeAndResolvers` so as to make it simpler, this is a great addition.

I agree that `getType` might be more confusing in this case, I did not get it at first (I thought it was picking into a list of registered type like getCollection etc.), I think it's ok to have user to add  `blackbox: true` manually if he actually doesn't want to add a schema.

I've added more tests for both use cases, I have only one broken test (not investigated). This version supports `typeName` for primitive types also, so allowing to handle Enums correctly and more broadly connection Vulcan to  any existing GraphQL schema.

PR #2363, opened 2019-08-29
[WIP] TypeScript

eric-burel:
Since TypeScript will soon be included into Meteor:

- [ ] Add TypeScript to the webpack loaders
- [ ] Move to Meteor 1.8.2 when it's officially released (and stable) to access TypeScript
- [ ] Write the types
- [ ] Use the types whenever some code is refactored

I have written only one small app in TypeScript. The only thing I know for sure is that statically typing JS code correctly is quite hard, especially for the frontend, so any contribution, tip, previous experience, feedback, best practices, will be very welcome.

### DIFF
--- a/packages/vulcan-lib/simple-schema.d.ts
+++ b/packages/vulcan-lib/simple-schema.d.ts
@@ -1,0 +1,158 @@
+// @see https://stackoverflow.com/questions/37406989/how-to-use-mongo-schema-validation-with-typescript-in-meteor
+declare module "simpl-schema" {
+
+    export class ValidationContext {
+        constructor(ss: any);
+        addValidationErrors(errors: any): void;
+        clean(...args: any[]): any;
+        getErrorForKey(key: any, ...args: any[]): any;
+        isValid(): any;
+        keyErrorMessage(key: any, ...args: any[]): any;
+        keyIsInvalid(key: any, ...args: any[]): any;
+        reset(): void;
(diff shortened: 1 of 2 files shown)


jonathanstiansen:
Hey Eric, not a pro either but my guess will be that the best way to use smart forms and such will be to use Generics -> then we can say something like `FormFieldsOf<T>`, and when `T` is a schema of type `User` then it will automatically know the form will have `id` `name`, etc. 

Keep me in the know, I don't have time to do the work right now but I'd be open to pairing on it in the future if that's something you're interested in.

eric-burel:
Thanks! I will probably merge this when we update to the latest version of Meteor. Then we will be able to actually enable TypeScript and start using it progressively in the core.

jonathanstiansen:
@eric-burel Nice - what's the roadmap on latest meteor being integrated? (Just for when people look at this they know not to ask ;-) )

eric-burel:
Hi, we will release a "patch" for version 1.14 soon, and then we will be able to update Meteor in the release afterward. We just need to check what it breaks beforehand, I don't want to mix it with other bugfixes. Sacha has opened an issue for discussion #2512

SachaG:
Oh nice I had forgotten about this PR. This could be a good starting point too.

eric-burel:
Closing as https://github.com/VulcanJS/vulcan-npm is now the reference, and... implemented in TS

PR #2293, opened 2019-05-05
vulcan-i18n: New function formatMessage(id, values, locale)

ErikDakoda:
Get a localized i18n string on the server side

PR #2369, opened 2019-09-04
Reenable data injection during SSR

### LINKED ISSUE #2218: Vulcan-using sites break when loaded inside archive.org
When a Vulcan-using website is rendered inside an archive.org snapshot (or a downloaded copy of the page, for that matter), the server-side rendered version displays briefly, and then is replaced with an Error404 component after Javascript initializes.

eric-burel:
TODO: 
- [X] Reenable Data injection during SSR + unit tests  (ongoing) 
- [ ] Check data injection from another part of the code outside renderPage
- [x] Handle mismatch between URL used during SSR and client side URL (see #2293) 
- [x] Test in the starter

Closes #2218


eric-burel:
@jimrandomh I'd be happy to have your feedback on this.
In the PR https://github.com/LessWrong2/Vulcan/pull/61, you compare the `pathname` to check whether the SSR URL and the client side URL are matching.
I did not understand this choice, so I went to use the request `hostname` vs the location.hostname instead. So loading from Archive.org will trigger a mismatch for example. Do you think this is ok?

I am affraid that it could break something eg if the request gets redirected to another host so I try to be as cautious as possible.

I went for the nuclear (and also simpler) option of disallowing ReactDOM rehydratation if there is a mismatch instead of modifying the React router history.

Btw InjectData should now be working back with the same API as before, sorry I disabled that during the Apollo 2 update by lack of understanding of this feature, I thought this was a patch or smth.

jimrandomh:
I think using `hostname` is probably better. We made a version that compared `pathname`, and reported mismatches without doing anything about them; our instrumentation ended up reporting a lot of spurious mismatches (between pairs of pages that link to each other), and we haven't figured out where those mismatches came from. Possibly something in the timing of clicking the SSR version of links, before scripts are fully initialized?

The main obstacle to using `hostname` is that, depending on DNS configs etc, the server doesn't necessarily know its own name. It's easy to accidentally end up configured in such a way that the site works as `www.example.com` but not as `example.com`, or vise versa.

eric-burel:
Thanks for your insights! Anyway there is a warning triggered when there is a mismatch so if there is a case we didn't think of it will still be easy to debug.

If it goes wrong, the alternative is whitelisting a few use cases, like loading a page saved as HTML or displaying the page in Archive.org.

PR #2374, opened 2019-09-09
PrefilledProps for form nested arrays

ErikDakoda:
In SmartForms, add support for prefilled props for nested array elements

### DIFF
--- a/packages/vulcan-forms/lib/components/Form.jsx
+++ b/packages/vulcan-forms/lib/components/Form.jsx
@@ -122,7 +122,7 @@ const getInitialStateFromProps = nextProps => {
   Utils.removeProperty(initialDocument, '__typename');
 
   return {
-    disabled: false,
+    disabled: nextProps.disabled,
     errors: [],
     deletedValues: [],
     currentValues: {},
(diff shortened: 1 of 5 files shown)


eric-burel:
Hi, I am trying to unit test this but I don't get the expected behaviour.

If I write `<Form prefilledProps={{addresses: { street: "foobar" }} />` for example, the problem is that `{street: "foobar"}` will be the prefilled value for the whole array too, so rendering fails.


Maybe I misunderstood the expected syntax? Should it be `addresses.$` instead?

eric-burel:
You can checkout the "ErikDakoda-FormNestedArrayPrefilledProps" branch on Vulcan, I've tested adding ".$" to the prop name to fill the array children instead of the array itself.

ErikDakoda:
Hi @eric-burel! That looks like a good change. Thank you! Could you create a pull request for your version?

eric-burel:
Yep it's merged, both having prefilled props for the whole array and for new items using ".$" will work. So you can even differentiate the first few values and subsequent values. Thanks!

PR #2375, opened 2019-09-10
Remove necessity to register HOCs

eric-burel:
Currently, all HOCs relying on fragment (`withSingle` etc.) cannot be called directly because GraphQL fragments are generated only after the app is started.
So this can't work:
```
export const FooBarWithMulti = withMulti({...})(FooBar)
```
Since `FooBarsDefaultFragment` is not yet defined at this point.

This led us to use `registerComponent` systematically in Vulcan app to defer the HOC creation. However `registerComponent` is a "power" pattern and is initially meant to share components and allow their replacement in a modular architecture, not to wrap them with HOC.

So, the goal of this PR is to allow calling `withMulti`, `withSingle` etc. directly.
First step is adding unit tests and refactoring the existing code to simplify it. I've managed to write a first working test. It took almost a day because React-Apollo warnings are just awful, however future tests will be similar so it's worth it. Incoming React Apollo version should take care of this but are still not mature at this point.

This might lead us to create hooks version of the HOCs, because hooks are inherently more dynamic than HOCs, since they are called within the component at runtime instead of "around" it at build time.

Related doc:
-  `graphql` HOC with dynamic fragments : https://github.com/apollographql/react-apollo/issues/330
- same with `hooks` : https://github.com/trojanowski/react-apollo-hooks/issues/83


eric-burel:
I've got a first working draft with withSingle.
This feature should not bring in regression, we support the same API.

React Apollo must be updated to 3.0.0 to provide this feature, and preferably replaced by the companion packages `@apollo/*`.

However I can't unit test mutations yet due to this bug: https://github.com/apollographql/react-apollo/issues/3478 :/

PR #2378, opened 2019-09-22
Ensure  props that can be applied to the SmartForm component  are propagated when said props are specified on the Card component.

Benefiter:
### Purpose:
    • Enable the Components.Card component to additionally support the props supported by the Components.SmartForm component
### Result:
    • Is is now possible to hide schema fields displayed on an editable Card component on the form displayed when the Edit button is clicked
### Process:
    • It is possible to elect to not show schema fields that ordinarily would be shown by default on a SmartForm using the hideFields prop. This same feature was not available on the Card component and more specifically to be able to hide schema fields on the Card edit form. The edit form uses the SmartForm so the solution was minimal and required simply passing the applicable SmartForm props specified on the Card component down the child component hierarchy to the SmartForm called to enable Card editing.

### DIFF
--- a/packages/vulcan-core/lib/modules/components/Card.jsx
+++ b/packages/vulcan-core/lib/modules/components/Card.jsx
@@ -191,7 +191,7 @@ const CardEdit = (props, context) => (
 
 CardEdit.contextTypes = { intl: intlShape };
 
-const CardEditForm = ({ collection, document, closeModal, ...editFormOptions }) => (
+const CardEditForm = ({ collection, document, closeModal, ...editFormProps }) => (
   <Components.SmartForm
     collection={collection}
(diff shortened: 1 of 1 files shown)


Benefiter:
> Great contrib! Could you please rename `editFormOptions` to `editFormProps` so it's clearer they are React props similar to the SmartForm props (I am being finnicky sorry)?
> Edit: except of course if you need to have this name for other reason like it's passed down by another Vulcan component or smth

This is great feedback and makes perfect sense. I'll make the change and push :-)

eric-burel:
Merged, thanks!

Benefiter:
> Merged, thanks!

Thanks so much Eric! You and the VulcanJS team are awesome!!

PR #2384, opened 2019-09-27
Material-UI - README Installation - Set Version of react-jss

diadama:
When you add react-jss without a version it will install the newly released version 10 (released 5 days ago)
You will get an error "Module not found: Can't resolve 'react-jss/lib/JssProvider"
There is already in issue in material-ui 
https://github.com/mui-org/material-ui/issues/13146

It solved the issue for me with this comment
https://github.com/mui-org/material-ui/issues/13146#issuecomment-534532789

I also tried to install it with @9.0.0 but npm didnt find any matching version. With the last version of 8.1.0 it works great


eric-burel:
Thanks! We indeed try to lock the versions of most libs until we can test newer updates safely.

PR #2392, opened 2019-10-10
Edit FormNestedArray to fix #2391

### LINKED ISSUE #2391: SmartForm delete two or more items from an array
To recreate this bug:

Clone Vulcan-Starter and run the `example-forms` package.
Log in, create a new `customer`, and add multiple `addresses` to it. (3 or more addresses to make it easiest to see problem in action.)
Edit the newly-created customer and delete two or more addresses. For instance, deleting the last address works as expected, but if you delete one more, it doesn't get removed from the page. This is the bug. (Note that if you submit the form, the address that was not removed from the page will be deleted from the database.)

kevinashworth:
- deletedValues is on the context, not props
- add forceUpdate to removeItem (is there a better way to accomplish a re-render?)


kevinashworth:
Revisiting this PR. It does not fix #2391.

PR #2402, opened 2019-10-15
split Card.jsx in differents files

juliensl:
Hi everybody ! 

I splited the *Card* code in different files, I think it's more readable like this.

What do you think ? :)


ErikDakoda:
This is probably a good idea. 

Also, I wonder if all this code should be moved to vulcan:ui-bootstrap to avoid duplication for those that use vulcan:ui-material.

juliensl:
Thank's Erik.

Actually I do not know exactly how is the better way to do this, I'm afraid to destroy Vulcan :laughing: 

So, If you agree, I let you do this, is it ok for you ?

ErikDakoda:
Okay, no worries.

juliensl:
Thank you very much Erik :smiley:

eric-burel:
Not sure if moving too Bootstrap is the right solution. It's the same issue with the Datatable and the Form, it makes sens to reduce the bundle size, but you keep having duplicate code in the end. 

What I tried to do instead is to split big components in a lot of smaller components whenever there is the slightest piece of logic, so the Material UI version only replace the visual components, with almost zero logic or at least zero non trivial logic. 
Eg if you compute a "showEdit" based on currentUser, you then pass this value to a replaceable components, so yo don't do the computation again in Material UI or Bootstrap.

It's difficult to figure out the right structure though.

eric-burel:
This piece of code from Card is a good example: 

```
  const canEdit =
    showEdit &&
    currentUser &&
    collection &&
    collection.options.mutations.update.check(currentUser, document);

  return (
    <div
      className={classNames(className, 'datacard', collection && `datacard-${collection._name}`)}>
      {title && <div className="datacard-title">{title}</div>}
      <table className="table table-bordered" style={{ maxWidth: '100%' }}>
        <tbody>
          {canEdit ? (
            <CardEdit collection={collection} document={document} {...editFormProps} />
          ) : null}
```
Instead it should be smth like: 
```
  const canEdit =
    showEdit &&
    currentUser &&
    collection &&
    collection.options.mutations.update.check(currentUser, document);

  return (
    <Components.CardInner canEdit={canEdit} {...props /> 
  )
```
So the UI can be replaced more easily. I don't know if this is realistic but I'd be glad to reduce the code duplication in UI packages somehow.

ErikDakoda:
Yes, that's right separating the logic and the actual rendering into different components is usually the best solution to minimize duplication.

PR #2403, opened 2019-10-15
Bugfix/version 1.13.4

kevinashworth:
I had to modify the `master` branch that sits at 1.13.4. A lot of packages in 1.13.4 are marked 1.13.3, and I don't think the `@=1.13.4` grammar is standard, so I changed that, too.

I don't know if PRs on the `master` branch are cool? But I hope this helps.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "Vulcan",
-  "version": "1.13.3",
+  "version": "1.13.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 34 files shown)


eric-burel:
I will check for the 1.13.3 but the `@=1.13.4` is needed to run older versions of Meteor. Without this you can't run older versions of Vulcan because it will load newer packages :/ Eg if you import explicitely import `vulcan:lib@1.13.2`, it will itself import `vulcan:users@1.13.4` because it's the latest version. But then it may fail because those versions are not necessary compatible.
Only solution I found was to lock versions, even if it makes update more tedious.

kevinashworth:
I understand on the `@=1.13.4`

eric-burel:
Closing as this will be sorted out progressively during next updates

PR #2413, opened 2019-10-20
Fixed links to changelog and migration documentation to

linegel:
For some reason, links led just to repo, instead of files relevant for branch.


eric-burel:
The reason is probably me being exhausted lol

PR #2418, opened 2019-10-22
Bugfix/get viewable fields

eric-burel:
@SachaG I've ended up creating a `Users.getReadableFields` that does return an array, and `Users.getReadableProjection` for the mongo projection.  I couldn't add an optional args because we already have the optional `document` so it would be messy.
I've deprecated  `getViewableFields` (it used the legacy "viewable" wording anyway) and rolled it back so it returns a Mongo projection.

I've also updated and tested the functions depending on `getViewableFields`.


SachaG:
Seems good to me, thanks! :)

PR #2419, opened 2019-10-22
Feature/mui 4

eric-burel:
Update to material ui 4, but not yet tested, still a few things to udpate.

- [X] Test on Storybook
- [X] Basic update following the doc
- [x] Check SSR, JSSProvider does not exist anymore, use StylesProvider instead
- [ ] Check the Composition Guide for dialogs, modals https://material-ui.com/guides/composition/#caveat-with-refs
- [ ] Test on non-trivial app, debug, clean warning
- [ ] Update versions in the doc

Migration guide: https://material-ui.com/guides/migration-v3/

@marieqg @juliensl @ErikDakoda


eric-burel:
It seems mostly functional, I will test it a bit on Live for good before merge.

The SSR API changed slightly. It uses the [`ServerStylesheets`](https://material-ui.com/styles/api/#serverstylesheets) object which setup the `StylesProvider` for you, which itselfs replaces the `JSSProvider` with a similar API.
It needs a bit of testing, I am not 100% sure there is no regression. 

Otherwise it's ok, there are no big changes apart those listed in the migration doc. I've tried to update components as much as possible in Vulcan, the same needs to be done in downstream apps.

ErikDakoda:
Sounds great, @eric-burel! Fast work!

PR #2425, opened 2019-10-29
Mailchimp

preminfy:
Issue:
Condition 'error.code == 214' is always false at this point because the false branch of the condition 'error.code == 214' at line 54 has been taken. 

Solution:
packages/vulcan-newsletter/lib/server/integrations/mailchimp.js
CONSTANT_CONDITION
        const message = error.message;
        if (error.code == 214) {
          name = 'has_unsubscribed';
        } else if (error.code != 214) {  //Changed to not equal to
          name = 'already_subscribed';


eric-burel:
Thanks !

PR #2427, opened 2019-10-29
Condition 'value === ''' is always false at this point

preminfy:
Issue:
Condition 'value === ''' is always false at this point because the false branch of the condition '!value' at line 294 has been taken. 

Solution: //Removed - || value === ''
packages/vulcan-core/lib/modules/components/Datatable/DatatableFilter.jsx 
CONSTANT_CONDITION
      inputProperties={{
        onChange: event => {
          const value = event.target.value;
          if (!value) { //Bug fix - Condition 'value === ''' is always false 
            const newFilters = Object.assign({}, filters);


preminfy:
Hi Eric,

Sure, yes it was a mistake that I closed them.

Thanks

PR #2429, opened 2019-10-31
Bugfix/close modal

EloyID:
In the version before the closeModal receives the data argument passed to the successCallback, so it can cause an error if closeModal treats this as an event object.


eric-burel:
Thanks this is the right fix
Don't you need to update Material UI too with the same fix?

EloyID:
For the editbutton its done, the newbutton of material ui didn't have
anything related to closeModal

Le jeu. 31 oct. 2019 à 10:34, Eric Burel <notifications@github.com> a
écrit :

PR #2434, opened 2019-11-07
prevent cacheUpdate from returning variables of a query similar name 

EloyID:
Example if we have in cache
myCustomQuery({limit:8}) //The one I want
myCustomQuery //I know I don't want this because causes an error in the line JSON.parse 
myCustomQueryFoo({limit:8}) //I don't want it

The test tries this three cases and asks to return just one, the first. If in the future we would want to manage the second case too.


eric-burel:
Thanks for this PR!

Can you maybe add a test to document the fact that `myCustomQuery` won't be detected, and that it is not a mistake from our side?

PR #2435, opened 2019-11-08
split getVariablesListFromCache test in two

EloyID:
Split of the test to two cases : 
- Queries starting with the same string (myCustomQuery({... and myCustomQuery2({...)
- Ignore zero variable queries 'myCustomQuery'


eric-burel:
Thanks! Almost there, however your test won't catch all mistakes. You should check the `variables` value in addition to its length (and use different variables for the queries). Here, your code could match the wrong query, and the test would still pass, as long as you match only one query.

I am being a bit finicky here, thanks for those units test, it helps a lot during updates to check for regressions.

PR #2437, opened 2019-11-09
Check if results is non-emtpy, not just present

larsivi:
If the collection exists, but is empty, and no columns are specified, the code would fail as results[0] would be undefined.


eric-burel:
Thanks!

PR #2442, opened 2019-11-13
Feature/add redux startup

yairtal:
As @eric-burel suggested, added a call back on startup.
This should get redux to work out of the box.

Please let me know if there is anything else I can do?


eric-burel:
Sounds allright, except for some console.log you left. Ping me if I forget to merge within a few weeks (I am bit busy I can't run the code right now)

yairtal:
Thank you for the feedback, removed console.log as if it was never there.
I'll send a reminder in a couple of weeks.

eric-burel:
Thanks a lot!

PR #2450, opened 2019-11-21
add preventDefault to ui-material ModalTrigger

yairtal:
The lack of a preventDefault is blocking the `a tag` or the Link from working.


eric-burel:
Thanks that's a good catch

PR #2451, opened 2019-11-22
Field level permissions validation take nested objects into account (read, create, update)

eric-burel:
Behaviour before PR:
```
const fooSchema = {
  nested: {
    type: new SimpleSchema({
      foo: {
         type: String,
         canCreate:["admins"] // this is ignored => members can actually create foo
    })
   canCreate: ["members"]   // this is validated, only members can create the "nested" object
}
```
The obvious issue is that nested permissions are ignored. This is not a bad beahviour per se, but it could confuse users.

After the PR:
```
const fooSchema = {
  nested: {
    type: new SimpleSchema({
      foo: {
         optional: true,
         type: String,
         canCreate:["admins"] // this is also taken into account
    })
   canCreate: ["members"]   // you still need this
}
```
So "members" could create an empty "nested", but not "nested.foo".
This is better. 

However, if you remove the `canCreate: ["members"']` at the root of the object, the whole object can't be created. You need both a permission at the root, and a permission per nested field.

So my question is: should we drop the check of the root of the nested object, and expect user to only provide a permission per nested field? Or is the current behaviour ok?
It feels a bit redundant. Same with array of objects, right now you may need to add a `canCreate` to `array`, `array.$`, and `array.$.yourNestedField`...

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -8,6 +8,7 @@ Please open an issue or a pull request if you feel this doc is incomplete.
 
 - See migration article from [Vulcan Blog](https://blog.vulcanjs.org/)
 - `serverTimezoneOffset` object is no longer injected in the head during SSR. Use `import { InjectData} from 'meteor/vulcan:lib; ...; await InjectData.getData("utcOffset");` instead. The value is the reverse from `getTimezoneOffset`, see [Moment doc](https://momentjscom.readthedocs.io/en/latest/moment/03-manipulating/09-utc-offset/)
+- `validateModifier` takes `data` as the second param (`validateModifier(modifier, data, document)` instead of `validateModifier(modifier, document)`)
 
 ### Material UI
 - Update to v4 `meteor npm i --save-exact @material-ui/core@4.5.1`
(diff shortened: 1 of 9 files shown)


doronrk:
The “After the PR” behavior you describe in the sounds very reasonable to me. When it comes to security, I’d always prefer some redundancy over ambiguity. Also, one could argue that specifying `array, array.$, and array.$.yourNestedField` is not actually redundant.

eric-burel:
Hi @SachaG @Apollinaire, I am going to merge this PR and test read permissions too, does the new API sound ok to you?

eric-burel:
Btw should'nt the "validation" code from `vulcan:lib` be moved in "permissions" from `vulcan:users`? Because that's where read permissions are tested with field restrictions

eric-burel:
I've added the same behaviour for `canRead`, I am now cleaning up the code.

Actually, maybe this PR could go into `1.14.1` instead of `1.14`, to make migration easier. 

The new nested schemas feature do not actually change how permissions worked in versions before 1.13 (describe as "Before PR" in my first post), where users were expected to set the permission for the whole object and couldn't set them for each nested field. So keeping the same behaviour at least for a "1.14.0" version sounds ok, even if it had limitations.

To sum it up I think we can push the `devel` branch in production safely (regarding field permissions at least) even without this PR, and merge it later.

eric-burel:
Even better: I made it backward compatible.

So if you don't set the permissions of a nested field that's fine, as the parent's permissions will be checked anyway. If you do, the field permissions will be used, so you should be able to have stronger permissions on some fields.

I haven't tested lighter permissions but I think it doesn't work (like if `nested` is restricted to members and `nested.foo` to guests, `nested.foo` won't be readable/createable/updateable).

PR #2453, opened 2019-11-26
Set non-null to false for data input type on update/upsert

doronrk:
Before this patch, the new single variable `input` API was causing the following error when used with `useUpdate2`. 

```
[GraphQL error]: Message: Variable "$data" of required type "UpdateGroupDataInput!" was not provided., Location: line 1, col 85, Path: undefined
```

This patch fixes the issue for update & upsert.


SachaG:
Thanks!

PR #2459, opened 2019-12-02
check options.input.limit and loadMoreInc accepts input

folgarr:
Fixes two issues I'm running into using `withMulti2({input: ...})`:

1. The limit that is provided as part of the input (`options.input.limit`) was not being used, instead the limit was always `defaultInput.limit`.

This is because `options.input.limit` is not one of the values that the function `initialPaginationInput` looks for, so it ends up going with the default for paginationInput.

This will now check for a provided `options.input.limit`.

2. `loadMoreInc` was returning duplicate values.

This seems to be because `fetchMore` changes the input variable of the original query to use terms: `variables: { input: { terms: newTerms } }`.

Instead of overriding the input like this, I've changed loadMoreInc to accept an input (which user can set as desired, including offset) or default to an input that contains the pagination input and a new offset.


eric-burel:
Totally right, thanks a lot for this fast PR

PR #2460, opened 2019-12-03
fix default_mutations check if user owns and belongs to authorised group

EloyID:
Before if a user owned a document and belonged to an authorized group but it wasn't allowed to owners to modify, the user would not be able to edit.


eric-burel:
Thanks, we may need to check that again after the permissions update

eric-burel:
Just read the code and this issue should'nt happen in the new version. Now, we use a special `owners` group, so the permission system is unified and can't produce this bug anymore.

PR #2463, opened 2019-12-03
Add Vulcan.generateGraphQLQueries to debug.js

doronrk:
**Context:** 
This function is aimed at enabling generation of typescript definitions for the default queries provided by Vulcan. Tools like [apollo codegen:generate](https://github.com/apollographql/apollo-tooling) generate typescript definitions when provided with a schema and queries/fragments. These definitions can be used to strongly type `useQuery`, `useMulti2`, and `useSingle2`. We're leveraging this currently at TradeWing and it really improves the developer experience. However, apollo codegen:generate needs a file that specifies the queries in order to generate the definitions.

**Changes:** 
* Adds an optional fileName argument to Vulcan.getGraphQLSchema - this helps with the auxiliary typescript generation scripts.

* Adds the Vulcan.generateGraphQLQueries function - this writes all of the default queries & default fragments for each collection to the path specified by the argument. 

**Feedback Welcome:**

There's almost certainly room for improvement here. I'm very open to feedback, particularly the substantive part in lines 81-105. But even this naive 1st pass has already provided some really solid benefits to us.

### DIFF
--- a/packages/vulcan-lib/lib/server/debug.js
+++ b/packages/vulcan-lib/lib/server/debug.js
@@ -60,6 +60,13 @@ if (Meteor.isServer) {
     });
   };
 
+  /*
+  This function is aimed at enabling generation of typescript definitions
+  for the default queries provided by Vulcan.
+
+  Tools like apollo codegen:generate generate typescript definitions when
+  provided with a schema and queries/fragments.
+  */
   Vulcan.generateGraphQLQueries = fileName => {
     let fd;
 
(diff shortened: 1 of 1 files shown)


eric-burel:
Not tested but that sounds ok to me, we are eager to merge PR related to debug tool since they can only bring improvements. 
Leveraging such external tooling is extremely interesting, as it's a continuation of Vulcan's schema-centered architecture. I dream of generating Mongoose schema based on the simple schema definition for example. It's great that we can do similar things for the GraphQL part already.

Can you plase just add a comment in the code to explain this function goal, or maybe add a few line to the doc, explaining your use case, in this page for example https://docs.vulcanjs.org/debug.html ? It could just be the first sentences of your PR, I don't want to bring you additional work.

It's an awesome contrib, thanks a lot.

doronrk:
Happy to make these suggested changes! Will do soon.

doronrk:
Cool. Let me know if the comment is sufficient. 

Will add mutations soon too.

eric-burel:
Perfect! Thank you

PR #2466, opened 2019-12-04
Omit username from schema validation after deleting it from options

doronrk:
When username is non-optional, the current code throws a validation error. 

If this function is going to delete username from options, then username should be omitted from the schema validation.


eric-burel:
Yeah cleaning up and testing this function is on the roadmap, this part of the code have a few discrepancies.

PR #2467, opened 2019-12-05
Feature/backoffice 1.14

Julien-Sytadelle:
Simply merged devel 1.14


eric-burel:
Thanks! We are hopefully reaching a working version of this backoffice

PR #2471, opened 2019-12-10
use apollo client in multiqueryupdater to broadcast to watched queries

folgarr:
Change `multiQueryUpdater` to use `client.writeQuery` which "leads to watched queries being re-broadcast" according to apollo team here: https://github.com/apollographql/apollo-client/issues/4398

Tested this on our app and it seems to fix the issue we were seeing with relevant queries not updating as expected.


eric-burel:
Omg Apollo API is so misleading, it really feels like they are writing the Meteor of GraphQL, half awesome, half terrible...
Thanks for this improvement!

eric-burel:
It may break the way we test it though (I was putting a spy on the cache object, so now the client could be passed to create a closure), but it should work in apps

Issue #2257, opened 2019-03-29
Create a "password" control


eric-burel:
When creating a user outside of Meteor Accounts, we could need to handle passwords in the SmartForm. We can set the input type to "password", but this does not allow to create a nicer default input for all passwords. I think passwords deserves to be handled like date, numbers and so on.

That would mean adding a "password" control in `FormComponent.jsx`.

eric-burel:
@juliensl could be an interesting issue for you
@SachaG would you be ok with this feature? Not an emergency at all of course

SachaG:
Yes that seems like a good idea.

eric-burel:
Closed by #2472

PR #2472, opened 2019-12-10
Add a password component

loustak:
Add a password component for both bootstrap and materialize.

Based on the Default (text) input, I just added a `type='password'`.


eric-burel:
Nice! For your next PRs you can write "Closes #2257" where 2257 is the related issue number at the end of your message, then github can close the issue automatically when the PR is merged.
Thanks!

PR #2474, opened 2019-12-11
Add GraphQLSchema.context to context in runGraphQL

doronrk:
This was causing a TypeError in vulcan-files because the file collection was not being added to the context that is passed to the resolvers in that package when querying with `runGraphQL`

### DIFF
--- a/packages/vulcan-lib/lib/server/query.js
+++ b/packages/vulcan-lib/lib/server/query.js
@@ -37,8 +37,10 @@ export const runGraphQL = async (query, variables = {}, context) => {
     queryContext[collection.options.collectionName] = collection;
   });
 
+  const fullQueryContext = merge({}, queryContext, GraphQLSchema.context);
+
   // see http://graphql.org/graphql-js/graphql/#graphql
-  const result = await graphql(executableSchema, query, {}, queryContext, variables);
+  const result = await graphql(executableSchema, query, {}, fullQueryContext, variables);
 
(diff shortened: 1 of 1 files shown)


eric-burel:
Just out of curiosity why merging with the graphQL schema context solves this? As far as I understand the file collection does not belong so it's missing in the `Collections` object, that was your issue?

doronrk:
The vulcan-files package adds the file collection to the context with 

`GraphQLSchema.addToContext({ [collectionName]: FSCollection });`

and all of its resolvers expect to be able to find the collection at context[collectionName]. But the collection is missing from the context object when run with `runGraphQL` because only the apollo-server context initialization did the merge, but not `runGraphQL`.

That was the specific issue we were facing, but in general, it seems that `GraphQLSchema.addToContext` should be respected regardless of whether the entry point to the query is from the client via the apollo server or a server-side query via `runGraphQL`. 

Does that make sense? Happy to clarify further.

eric-burel:
Yeah I think I undestand, I didn't even know about this `addtoContext` (I barely actually use Vulcan GraphQL related features outside of automated generation)

eric-burel:
The collection should be registered in the Collection object somehow I think, it's overkill to extend the contextfor this, but you are right that some other part of the code could want to update the context for other reasons through this method. 

I think my idea was to use a callback instead to enhance the context, this may need some clarification in future releases.

PR #2476, opened 2019-12-11
create2 delete2 multiQueryUpdater minor changes

folgarr:
1. export the `multiQueryUpdater` from delete2.js so others can provide it with their own args to `mutationOptions.update`.
2. remove the `resolverName` argument to the create2.js `multiQueryUpdater` because it can be created inside the function (users shouldn't have to construct this resolverName string from their app)

This is what we're using now in our project.

### DIFF
--- a/packages/vulcan-core/lib/modules/containers/create2.js
+++ b/packages/vulcan-core/lib/modules/containers/create2.js
@@ -50,8 +50,8 @@ export const multiQueryUpdater = ({
   fragment,
   fragmentName,
   collection,
-  resolverName,
 }) => async (cache, { data }) => {
+  const resolverName = `create${typeName}`;
   const multiResolverName = collection.options.multiResolverName;
   // update multi queries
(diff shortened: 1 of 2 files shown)


eric-burel:
Copy pasting from Slack so we keep track of this:

> You need to reuse this function for upserting so you can't hardcode the resolver name, we must find another approach, maybe a closure would do what you want?
> basically smth like (computeResolverName) => ({typeName,...}) =>
> In create2 you export a version where computeResolverName returns createFoobar and in upsert upsertFoobar. This function needs to be abstracted away anyway because the main logic should be reusable for all mutations, only the resolver name and updzte logic actually change

folgarr:
Going to hold-off on this as we now realize we don't need to be passing multiQueryUpdater to useCreate2 and useDelete2 because Vulcan seems to be calling it with what we want if we just pass our own fragment via options.

eric-burel:
Ok no problem! Thanks

PR #2478, opened 2019-12-12
Pass full doc and document to validate to validateDocumentPermissions

doronrk:
The full document must be passed to validateDocumentPermissions even on a partial update because permission logic may rely on other fields in the document. Even the core 'owners' behavior relies on this, and will fail if 'owners' isn't included in the document passed to `Users.canUpdateField`


eric-burel:
Makes sense thanks!

Issue #2415, opened 2019-10-21
Update order and filters in lists after a document is updated


eric-burel:
Apollo now updates the cache directly when updating a value, so values in lists are immediately updated after a mutation.

However, we still need to filter the values again and rerun the sort, because the update can have modified them.

The code already exists since it was used in the previous implementation with watched:mutations, we simply need to write a test and reenable it through a `multiQueryUpdater` (see the create version for inspiration).

yannick-mayeur:
When I make an update to an item in a list for example the movie list in the Starter, first the item stays at the same place and then after a few seconds it moves to the correct position without me doing anything. Is the expected behavior that the item instantly moves to the correct position?

When I add a new item the behavior more or less the same. First the item spawns at the top of the list and then after a few seconds it gets moved to the correct position. Does this that the create is buggy too?

eric-burel:
Beware of versions, the Datatable uses the old resolvers so they are not good for testing this. the `multi.js` has indeed a small bug (`collection.getParameters` overrides the order wrongly I think) but it's not a top priority because this code is deprecated. We use `multi2.js` currently, the `multi.js` will be deleted in a future release.

You may need to write a unit test instead or use directly "multi2" resolver in a sample package.

eric-burel:
@yannick-mayeur See the first schema in this article to understand what happens with the Datatable: https://blog.vulcanjs.org/vulcan-js-1-14-filtering-new-apis-fc8efc0027dd

We plan to update it of course but did not have enough time during the latest release.

yannick-mayeur:
>  Beware of versions, the Datatable uses the old resolvers so they are not good for testing this. the multi.js has indeed a small bug (collection.getParameters overrides the order wrongly I think) but it's not a top priority because this code is deprecated. We use multi2.js currently, the multi.js will be deleted in a future release.

`multi2.js` seems to be used by the Datatable in the `devel` branch. So if I use a 2 repo install it should use `multi2.js` right?

eric-burel:
Not sure, but you may easily test that with a console.log in the multiQuery updater function

eric-burel:
There are unit tests for the create version too that you can run, I remember unit testing the sort specifically

yannick-mayeur:
I think that the tests related to the create version are those in: `vulcan-core/test/containers2/mutations.test.js` but they fail with the following error message:

```
I20191217-11:35:14.788(1)?   2) vulcan:core/container/mutations2
[...]

PR #2484, opened 2019-12-17
Move mutations.test to the client side

yannick-mayeur:
This has to be done on the client side because on the server side we do not have access to a full client with a cache.
Linked to #2415.


eric-burel:
You probably should keep most of the test on both client and server actually, only tests related to `multiQueryUpdater` are to be moved only client side.

yannick-mayeur:
I have updated this to only move the correct tests. The client side tests are still failing though.

eric-burel:
Can you show the error message for client side test? You may have something either in the server console or the browser console

eric-burel:
I think we can merge this already if you remove the "Draft" state (I can't change it myself), and you can  open another PR for your fixes and improvements later

yannick-mayeur:
The error I have is the same I already wrote about in the issue: https://github.com/VulcanJS/Vulcan/issues/2415#issuecomment-566570429

eric-burel:
Fixed and merged, thanks a lot!

PR #2485, opened 2019-12-17
Remove hardcoded head tag in template-web.browser

yannick-mayeur:
This is done to avoid two generated opening head tags. The dynamicHead
passed as a paramter to the headTemplate already contains the opening
tag.

Related issue: VulcanJS/Vulcan#2443

### DIFF
--- a/packages/vulcan-lib/lib/server/apollo-ssr/components/Head.jsx
+++ b/packages/vulcan-lib/lib/server/apollo-ssr/components/Head.jsx
@@ -6,11 +6,11 @@ const Head = () => {
   //@see https://github.com/nfl/react-helmet/releases/tag/5.0.0
   const helmet = Helmet.renderStatic();
   return (
-    <head>
+    <React.Fragment>
       {helmet.title.toComponent()}
       {helmet.meta.toComponent()}
       {helmet.link.toComponent()}
-    </head>
+    </React.Fragment>
   );
 };
 export default Head;


eric-burel:
Can you try this with SSR disabled too? I am afraid that in this case you may have 0 head tag...

yannick-mayeur:
I disabled SSR on the VulcanStarter repository and ended up having 0 head tag:

```
<!DOCTYPE html>
<html>

  <link rel="stylesheet" type="text/css" class="__meteor-css__" href="/merged-stylesheets.css?hash=54f41571f9fdcaeb8cec2e672edddf3f7076ee64">

</head>
```

yannick-mayeur:
We will have to see if we can remove the head tag from the `dynamicHead` instead in case SSR is enabled.

yannick-mayeur:
Is the `dynamicHead` something that is generated by the Apollo framework or is it done in Vulcan.

If it is something done by Apollo I think the only solution would be to dynamically test if the `dynamicHead` contains an opening head tag or not and then decide if we should add one.

eric-burel:
`dynamicHead` is provided by Meteor so that you can add head tags.

@yannick-mayeur I think the correct solution is modifying this file: packages/vulcan-lib/lib/server/apollo-ssr/components/Head.jsx

Basically that's what happens:
- `template-web.browser.js` will provide the basic HTML structure of your page, including a `head` tag.
- if you have SSR, we had some content to the `dynamicHead`, which is rendered by this template. If you don't, it is still rendered by this template but it's empty
- the page is served

So the mistake is to add another `<head>` during SSR. I was not familiar with the `template-web.browser.js` file at the time we worked on Apollo 2 so I mistakenly added it.
You can use `React.Fragment` instead, I think it will do the trick. And you can cancel your changes to the template.

eric-burel:
This `Head` is used here by the way: `packages/vulcan-lib/lib/server/apollo-ssr/renderPage.js`

yannick-mayeur:
I have updated the PR. It now works with and without SSR.

eric-burel:
Thanks a lot!

PR #2490, opened 2019-12-20
Jt/extend collection fix

josephto:
customFilters is a nested array, and _.merge doesn't deal with that properly. Use _.mergeWith instead


eric-burel:
Hi, thanks for this PR, next time can you please just cleanup the commits a bit? You can run "git rebase -i HEAD~X" and stash them to have only one clean commit

PR #2493, opened 2019-12-30
datatable fix for material ui

### LINKED ISSUE #2475: Material-UI Datatable doesn't populate columns
When using the datatable with Material-ui, the columns do not show up, however, rows show up and when the edit button is clicked, the proper data is populated there.

Neobii:
The `columns` in the data were not populating correctly so I grabbed https://github.com/VulcanJS/Vulcan/blob/devel/packages/vulcan-core/lib/modules/components/Datatable/DatatableContents.jsx#L9 from the core DatatableContents component.

Also I added getCollection so `collectionName` can be specified on the collections.

Fixes  #2475

### DIFF
--- a/packages/vulcan-ui-material/lib/components/core/Datatable.jsx
+++ b/packages/vulcan-ui-material/lib/components/core/Datatable.jsx
@@ -4,9 +4,10 @@ import {
   Components,
   replaceComponent,
   withCurrentUser,
-  withMulti,
   Utils,
+  withMulti
 } from 'meteor/vulcan:core';
+import compose from 'recompose/compose';
 import { intlShape } from 'meteor/vulcan:i18n';
 import withStyles from '@material-ui/core/styles/withStyles';
 import Table from '@material-ui/core/Table';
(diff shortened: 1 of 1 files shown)


packages/vulcan-ui-material/lib/components/core/Datatable.jsx
```diff
@@ -132,14 +134,16 @@ class Datatable extends PureComponent {
...
+
+      const collection = this.props.collection || getCollection(this.props.collectionName);
 
       const listOptions = {
         collection: collection,
         ...options,
       };
 
-      const DatatableWithMulti = withMulti(listOptions)(Components.DatatableContents);
+      const DatatableWithMulti = compose(withMulti(listOptions))(Components.DatatableContents);
```
eric-burel: Why the compose call hre?

Neobii: I saw it here: https://github.com/VulcanJS/Vulcan/blob/devel/packages/vulcan-core/lib/modules/components/Datatable/Datatable.jsx#L219 so I decided to keep similar code style to the Datatable in core.

eric-burel:
Merged, thanks!

Issue #15, opened 2012-09-30
Log in with Gravatar possible?


b-d-m-p:
I thought you said that we could log in with Gravatar at the meeting. I thought it would be a nice feature and it would be cool to see people's faces. Not a serious issue though.

PR #2494, opened 2020-01-03
Add extendedUpdateFunc in create2 (#15)

doronrk:
Currently, `useCreate2` will place the created document at `options.propertyName || 'document'` in the returned result. This is nice because the client of the hook doesn't have to know about the structure of the executionResult/resolverName to access the created document.

This PR adds basically the same functionality to the update function so that clients of the hook can access the created document at at `options.propertyName || 'document'`

### DIFF
--- a/packages/vulcan-core/lib/modules/containers/create2.js
+++ b/packages/vulcan-core/lib/modules/containers/create2.js
@@ -105,6 +105,18 @@ const buildResult = (options, resolverName, executionResult) => {
   return props;
 };
 
+const extendUpdateFunc = (originalUpdate, options, resolverName) => {
+  const propertyName = options.propertyName || 'document';
+  return (cache, executionResult) => {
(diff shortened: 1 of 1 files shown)


eric-burel:
You intend to use that in a custom update function for example, that would bypass the default `multiQueryUpdater` ?
 Could you please maybe wrap this into a small function call (smth like `enhanceUpdateFunc(options)`), just for a matter of readability, since this part of the code is quite complex to understand?

doronrk:
> You intend to use that in a custom update function for example, that would bypass the default multiQueryUpdater ?

Yep exactly. 

> Could you please maybe wrap this into a small function call 

Sure thing!

doronrk:
@eric-burel  I'm closing this because it will just be easier to land PR 2501 first.

doronrk:
merged a commit into PR 2501 with this functionality.

Issue #14, opened 2012-09-30
Mobile version has extra space at bottom of page


b-d-m-p:
The mobile version is great. It just has some extra space at the bottom. Also, the chat doesn't really work in it, but still, great job.

PR #2495, opened 2020-01-03
For create mutation, move onCreate to before validation to avoid vali…

doronrk:
…dation errors. (#14)

Currently Vulcan does not support a required field that is created via `onCreate`. A really common example is `dateCreated`, but in general it should be possible to have fields that are required but generated by the server in `onCreate`.

One could argue that passing a dummy value would solve the problem. While this might be reasonable for `dateCreated`, it is not a general solution, since there may very well be fields whose purpose/existence is completely hidden from the client.

### DIFF
--- a/packages/vulcan-lib/lib/server/mutators.js
+++ b/packages/vulcan-lib/lib/server/mutators.js
@@ -98,6 +98,30 @@ export const createMutator = async ({
 
   /*
 
+  onCreate
+
+  note: cannot use forEach with async/await.
+  See https://stackoverflow.com/a/37576787/649299
+
+  note: clone arguments in case callbacks modify them
+
+  */
+  for (let fieldName of Object.keys(schema)) {
+    let autoValue;
+    if (schema[fieldName].onCreate) {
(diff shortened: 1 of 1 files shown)


eric-burel:
Why would you need the field to be required then? So that we guarantee that it has a value in the database?
I think the validation step is more targeted at user input, than actual database content. There seems to be some ambiguity in what we mean by "required", I am afraid that for example in the future we could decide to check required field client side first, then onCreate would not be able to run.

doronrk:
Yes, the purpose is for making a field that is required in the database, but that should not be supplied by the user. For example, `dateCreated`, `authorId` (of a post/comment).

Since Vulcan asks the programmer to combine the database schema with the API specification, and because the semantics of `optional` are defined by simpl-schema, it seems reasonable to apply `optional` to the database schema, and perhaps introduce a new field like `inputRequired` (naming is hard :) ) for the functionality you described above. Otherwise, if we use `optional` to refer to the user input, I worry that the programmer has no way of making a field required in the database schema, but optional/not-allowed in user-input. 

Definitely curious to hear your thoughts, since I know you understand the future plans for Vulcan much better than I do. Thanks!

Issue #16, opened 2012-10-01
Issue when editing a post


ekianjo:
When editing a post the page does not come back at the top, therefore it appears you arrive on a blank page and you need to scroll up to see your actual comment to be edited. 

Minor bug, but nice to fix if possible.

PR #2496, opened 2020-01-06
fix bugs in extendCollection (#16)

doronrk:
The current implementation is buggy when Array.isArray(a) !== Array.isArray(b)

This can be the case for permissions `['members']` vs.  `() => ....` for example


eric-burel:
Makes sense thanks

PR #2497, opened 2020-01-06
Use multiQueryUpdater in update2 to properly sort/filter

yannick-mayeur:
This needs to be properly tested before getting merged. Based on Unit
Tests it does not break anyting.

When building the demo application for the new datatable with @loustak,
we will make sure to try this commit out.

### DIFF
--- a/packages/vulcan-core/lib/modules/containers/update2.js
+++ b/packages/vulcan-core/lib/modules/containers/update2.js
@@ -37,6 +37,8 @@ import {
 } from 'meteor/vulcan:lib';
 import { computeQueryVariables } from './variables';
 
+import { multiQueryUpdater } from './create';
+
 export const buildUpdateQuery = ({ typeName, fragmentName, fragment }) => (
   gql`
     ${updateClientTemplate({ typeName, fragmentName })}
@@ -58,6 +60,7 @@ export const useUpdate2 = (options) => {
   const [updateFunc, ...rest] = useMutation(query, {
     // see https://www.apollographql.com/docs/react/features/error-handling/#error-policies
     errorPolicy: 'all',
+    update: multiQueryUpdater({ typeName, fragment, fragmentName, collection, resolverName: `update{typeName}` }),
     ...mutationOptions
   });
 
--- a/packages/vulcan-core/lib/modules/containers/update2.js
+++ b/packages/vulcan-core/lib/modules/containers/update2.js
@@ -60,7 +60,7 @@ export const useUpdate2 = (options) => {
   const [updateFunc, ...rest] = useMutation(query, {
     // see https://www.apollographql.com/docs/react/features/error-handling/#error-policies
     errorPolicy: 'all',
-    update: multiQueryUpdater({ typeName, fragment, fragmentName, collection, resolverName: `update{typeName}` }),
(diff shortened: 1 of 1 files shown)


yannick-mayeur:
Maybe this turns out to be useless thanks to https://github.com/VulcanJS/Vulcan/pull/2494

packages/vulcan-core/lib/modules/containers/update2.js
```diff
@@ -58,6 +60,7 @@ export const useUpdate2 = (options) => {
   const [updateFunc, ...rest] = useMutation(query, {
     // see https://www.apollographql.com/docs/react/features/error-handling/#error-policies
     errorPolicy: 'all',
+    update: multiQueryUpdater({ typeName, fragment, fragmentName, collection, resolverName: `update{typeName}` }),
```
eric-burel: it's `update${typeName}` you missed a $

eric-burel:
As far as I understand #2494 is targeted at custom update functions, so it is compatible with what you've done
I've just added a comment because the resolver name was wrong

eric-burel:
It sounds fine to me, let me know when you think it's ok to merge. Thanks!

yannick-mayeur:
I finally got around to testing this.
In the `getting-started` package the `MovieApp2` component uses the `Datatable` component, which I think uses `multi2`.
But when updating it first stays at position it was and then moves to the correct one. The same thing happens when creating a new object though (it first appears at the top of the list and then after a few seconds moves to the correct position).

eric-burel:
Ok can you mark this PR as valid (not a draft, I can't do it myself)? I'll merge it because at least update is consistent with create.

eric-burel:
@yannick-mayeur Datatable uses the new multi but the form itself do not use the new `create`/`update` so I think it makes sense. The `multiQueryUpdater` is tied to the mutation create/update, not to the multi query.

eric-burel:
Yeah I confirm I was wrong in https://github.com/VulcanJS/Vulcan-Starter/issues/140, the issue is not the DataTable using multi2, it's the Smart form using create/update.
So basically we probably need to update the smart form instead. But the behaviour is consistent, I think the multiQueryUpdater is ok in itself. Can you please also confirm #2494  is ok to be merged?

yannick-mayeur:
https://github.com/VulcanJS/Vulcan/pull/2494 is closed

eric-burel:
Yes I mean the newer one https://github.com/VulcanJS/Vulcan/pull/2501

yannick-mayeur:
With #2501, when creating the item does not appear for few seconds but when it appears it directly at the correct position.

eric-burel:
That's not good it should appear right away. When it appears only after a few seconds, it's because of a polling request (you can see it in the Network tab). The multiQueryUpdater should make it appear immediately (ideally at the correct position but that's the another issue).

eric-burel:
@yannick-mayeur I've fixed the order issue. The solution is to force a default order in the default resolver. Otherwise you can't tell in which order the data will be returned by the server, so it's impossible to match the order client side and server side. With a default order, if no `sort` is provided, it will now use `{createdAt: -1}` so the order stays consistent.

Please mark your PR as not a draft so I can merge it. Thanks a lot for investigating this.

yannick-mayeur:
Great!

I marked the PR ready for review.

PR #2498, opened 2020-01-07
fix field permission validation bug

folgarr:
Fixes a bug I ran into when `isNested === false` and `mode === create` and `fieldSchema.canUpdate === undefined`. Surprisingly, the `if` on this line was evaluating to true leading to the early return.

Adding the parentheses fixes the operator precedence issue on this line and runs as intended. Verified that it fixes the issue we were seeing.


eric-burel:
Thanks! Yeah I try to unit test those feature but it's easy to miss some cases :/

PR #2499, opened 2020-01-07
Semantized classname

### LINKED ISSUE #2423: Semantized classNames
As stated by @juliensl , autogenerated HTML code lacks some semantic classNames, in order to write stable selector.

This is especially necessary for end-to-end testing, eg with Cypress (that we could add to the Starter as the default testing framework btw).

To solve this we would need to add className based on the schema field names and collection/type name on Datatable, Smartform and Card. Those classes only needs to be added to key parent elements (a form group, and input wrapper etc.).

loustak:
Classnames of wrapper component such as datatable and forms are now semantized. (In particular buttons and links to sign in and register)
It's possible to access each part of the component with a unique class or id name.

This PR is linked with this one: https://github.com/VulcanJS/Vulcan-Starter/pull/156

Closes #2423

### DIFF
--- a/packages/vulcan-accounts/imports/ui/components/Button.jsx
+++ b/packages/vulcan-accounts/imports/ui/components/Button.jsx
@@ -10,15 +10,24 @@ export class AccountsButton extends PureComponent {
       // href = null,
       type,
       disabled = false,
+      id,
       className,
       onClick
     } = this.props;
 
     return type === 'link' ? 
-      <a href="#" className={className} onClick={onClick} style={{marginRight: '10px'}}>{label}</a> :
+      <a
+        href="#"
+        id={id}
+        className={className}
+        onClick={onClick}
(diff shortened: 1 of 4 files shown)


eric-burel:
Awesome! Does it works with cards to? We use them less often but they can still be useful in some case

loustak:
> Awesome! Does it works with cards to? We use them less often but they can still be useful in some case

I don't think, I only tried components present in the example-forms package of vulcan starter. Should I take a look at that?

eric-burel:
Yes it should not be a big deal but basically SmartForm, Datatable and Card are always working together when we talk about collections (table to list multiple items, cards to list one item, form to create/update an item). You don't need to write a full test for that, just check that you have a relevant className in the Card component.

loustak:
Ok I'll take a look at Card.

PR #2501, opened 2020-01-08
Added single query updaters for create2 and delete2

josephto:


eric-burel:
Awesome work! We will test that alongside #2497

eric-burel:
Hi, according to @yannick-mayeur tests it does not seem to correctly add data in lists on item creation, any idea why?

PR #2502, opened 2020-01-11
fix material ui seletmultiple

yairtal:
official selectmulti example:
https://codesandbox.io/s/hzojf

the special handling of a change is relevant only while `native` and `multiple` are selected


eric-burel:
Ok makes sense, I'll trust you on this change

PR #2503, opened 2020-01-13
Enhancement/mui radio

marieqg:
Hello, I added 3 new views for the MuiRadioGroup to display : 
- 10 columns of MuiRadio (if the max length of the label is < 3) 
- 8 columns of MuiRadio (if the max length of the label is <7 ) 
- 5 columns of MuiRadio (if the max length of the label is <12 ) 
and made some minors changes to the responsivity of already existing views. 

Also, I fixed some margin issue with the label. The label had a 12px margin, and the radio none. So, the label was higher than the radio.

You can find here screenshot of a SmartForm with radiogroup on different devices: 
https://ibb.co/WKzyHTM
https://ibb.co/ZgzJyC9
https://ibb.co/bv4SJCV
https://ibb.co/Sxsy52t
https://ibb.co/4fNFPtN
https://ibb.co/vDqKz9r
https://ibb.co/6rqDK1q
https://ibb.co/p0B3fGm

Thanks in advance!

### DIFF
--- a/packages/vulcan-ui-material/lib/components/forms/base-controls/MuiRadioGroup.jsx
+++ b/packages/vulcan-ui-material/lib/components/forms/base-controls/MuiRadioGroup.jsx
@@ -22,6 +22,7 @@ const styles = theme => ({
   },
   twoColumn: {
     display: 'block',
+
     [theme.breakpoints.down('md')]: {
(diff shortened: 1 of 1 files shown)


eric-burel:
Thanks!

PR #2504, opened 2020-01-14
Prevent calling forEachDocumentField on fields without a schema

eric-burel:
Avoid failing when a field is present in a JSON document but is not declared in the schema.

Theoretically this is not a valid pattern, however as `forEachDocumentField` is a generic helper we can reasonnably expect it not to fail on invalid documents and let the user decide what to do with missing fieldSchema instead 

@kevinashworth for feedback


kevinashworth:
This one change works wonders! And, yes, at least in my case it has to do with one or more fields in the JSON document that are not in the schema. I'm going to cleanse/validate my database, but in the meantime, this PR is exactly what I need.)

eric-burel:
Thanks for your feedback and for helping finding the issue!

PR #2505, opened 2020-01-15
Allow itemProperties to be passed on to FormComponent

kevinashworth:
Move `itemProperties` above `field` inside `FormComponents.FormComponent`

It seems that destructing `{...field}` effectively removes itemProperties from getting sent to FormComponent because `{...field}` includes an itemProperties

So for me, this change gets `itemProperties: { layout: 'inputOnly' }` working again in `schema.js` use


eric-burel:
Makes sense, thanks

PR #2509, opened 2020-01-22
Properly provide collection callbacks in vulcan-users on_create_user …

doronrk:
…runCallback invocations

Mimics the way `runCallbacks` is invoked in mutations


eric-burel:
@SachaG with the new callback system how do you add a callback to a collection from any part of the code? Smth like `Users.options.callbacks.create.push(myCallback)`?

To me the call to `runCallbacks('user.create.validate.before')` is a deprecated pattern in itself but I am not sure how we are supposed to update existing code regarding the new API.

PR #2511, opened 2020-01-28
Remove root element from SSR

yannick-mayeur:
This has to be done because is also added by the Start.jsx file.

### DIFF
--- a/packages/vulcan-lib/lib/server/apollo-ssr/renderPage.js
+++ b/packages/vulcan-lib/lib/server/apollo-ssr/renderPage.js
@@ -81,12 +81,7 @@ const makePageRenderer = ({ computeContext }) => {
       }
     }
 
-    // TODO: there should be a cleaner way to set this wrapper
-    // id must always match the client side start.jsx file
-    const wrappedHtmlContent = `<div id="react-app">${htmlContent}</div>`;
-    sink.appendToBody(wrappedHtmlContent);
-    // TODO: this sounds cleaner but where do we add the <div id="react-app"> ?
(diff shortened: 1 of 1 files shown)


eric-burel:
Thanks, does it work both with SSR and without SSR? You don't have any SSR errors?
Also, are the tests from "packages/vulcan-lib/test/server/apollo-ssr.test.js" still ok? The one that checks `expect(sink.result.body).toMatch('<div id="react-app">');` might probably be deleted after your PR.

yannick-mayeur:
How can I deactivate SSR to test locally if it still works?

eric-burel:
@yannick-mayeur you did smth similar to debug the double head tag, you can disable SSR through a setting in `settings.json` (I can't remember exactly but you'll find it easily, smth like apolloServer.disableSsr)

yannick-mayeur:
Oh yeah I remember doing something like that. I will try to find out what I did exactly.

yannick-mayeur:
Okay this solution does not seem to be the right one. When testing with the `getting-started` package from the Starter, I have no `react-app` when it is not applied by the SSR. The better solution is probably https://github.com/VulcanJS/Vulcan/pull/2204.

eric-burel:
Yeah it's not exactly the same behaviour, can you please give #2204 a shot and tell me if it's mergeable?

eric-burel:
Closing in the meantime

yannick-mayeur:
I am not able to reproduce the bug of having two root elements in the application. But it does not crash the application nor completely remove `react-app`.

PR #2513, opened 2020-01-30
Feature/backoffice

yannick-mayeur:


packages/vulcan-backoffice/lib/components/BackofficeLayout.jsx
```diff
@@ -0,0 +1,89 @@
...
+import {
+  getAuthorizedMenuItems,
+  menuItemProps,
+  registerComponent,
+  withCurrentUser,
+  Components,
+} from 'meteor/vulcan:core';
+import { Link } from 'react-router-dom';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
```
eric-burel: No you can't do that. In core packages you don't have access to a styling solution, you should expect Bootstrap/Material UI to style the component. You can use the `registerComponent` pattern so that your component can be overriden by those packages in order to add style.
You shouldn't have any dependency to styled-components in the code.

packages/vulcan-backoffice/lib/components/BackofficeLayout.jsx
```diff
@@ -0,0 +1,89 @@
...
+};
+
+MenuItem.propTypes = {
+  ...menuItemProps,
+  // parent can pass another onClick callback
+  // eg to close the menu
+  afterClick: PropTypes.func,
+};
+
+const PageLayout = styled.div`
```
eric-burel: Instead you should have smth like `const PageLayout = (props) => <div {...props}>{children}</div>; registerComponent({name: "VulcanBackofficePageLayout", component: PageLayout});`. Then only in Bootstrap/MaterialUI you can replace the component and add style to it.

eric-burel: This sound a bit unnatural in React but that's how we allow a package oriented architecture in Vulcan with easy customization of components.

packages/vulcan-backoffice/lib/components/BackofficeVerticalMenuLayout.jsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import { registerComponent } from 'meteor/vulcan:lib';
+import styled from 'styled-components';
+
+const StyledWrapper = styled.div`
```
eric-burel: Same, no style here, only replaceable components

packages/vulcan-backoffice/package.js
```diff
@@ -0,0 +1,18 @@
+Package.describe({
+  name: 'vulcan:backoffice',
+  summary: 'Vulcan automated backoffice generator',
+  version: '1.13.3',
```
eric-burel: version is wrong now it should be 1.14.1 I think (same  as core after a merge from devel)

packages/vulcan-ui-material/lib/components/core/Datatable.jsx
```diff
@@ -44,7 +44,8 @@ const baseStyles = theme => ({
     alignItems: 'center',
   },
   scroller: {
-    overflow: 'auto',
+		overflowX: 'auto',
+		overflowY: 'hidden'
```
eric-burel: Not sure this work, there is a known issue with overflows which I can't remember (like if you get overflowX you can't hide  overflowY but maybe I am wrong)

eric-burel: An issue in browsers themselves so we can't do anything about it

packages/vulcan-backoffice/lib/components/BackofficeLayout.jsx
eric-burel: Also in material UI we don't use Styled components either but Material UI own styling solution (syntax is exactly the same so you can probably copy paste most of you code and just change the import)

eric-burel:
I think you are trying to merge in the wrong branch too, you should probably merge in the existing `feature/backoffice` branch instead.

eric-burel:
The weird positionning is explained by the fact you tried to use styled-components in the core. When doing that you don't have SSR, since you need the additional vulcan:styled-components package to do so. You can read the code of this package, it's very short and you'll understand what it does.
See my comments to fix that by moving style into material ui/bootstrap packages instead.

packages/vulcan-backoffice/lib/components/BackofficeLayout.jsx
loustak: Ok for the first point, we are going to change the code. But what are the adventages of using material-ui own styling? It makes inconsistencies in the code, no? It's easier to have the same (or almost) code for both bootstrap and material-ui.

loustak:
Ok, we will move styling to bootstrap and material packages, but will this solve the broken SSR? We will still need to enable the vulcan:styled-components package to make it work?

packages/vulcan-backoffice/lib/components/BackofficeLayout.jsx
eric-burel: For React component that may be true but for the CSS because they simply don't use the same approach. Material UI uses CSS-in-JS, while Bootstrap doesn't. We can't force people to setup styled-components or CSS-in-JS in their own app, Vulcan's users usually uses only one of those framework so they don't care about consistency, it just affects us as contributors. You are right that it duplicates our code a bit but for the end users it feels way more intuitive. 

I worry more about the discrepancies we can have within the components logic, that are actually difficult to solve, than about the CSS.

eric-burel: Using material UI solution is consistent within the `vulcan:ui-material` package so it's ok

eric-burel:
Yes it will solve it because `vulcan:ui-material` have a specific setup for SSR. You won't need `vulcan:styled-components`, Material UI uses its own solution already (it's very similar, you can even use the same syntax nowadays). If you are interested you can see what we do in the `server` part of Material UI for SSR.

eric-burel:
For Bootstrap you won't have styling during SSR because it's a separate file but that's ok, you can't do much about it, that's how it works.

loustak:
> For Bootstrap you won't have styling during SSR because it's a separate file but that's ok, you can't do much about it, that's how it works.

I'm not sure to understand, how I'm supposed to apply CSS? Using an external stylesheet? Inline style? Styled components?

eric-burel:
I think we use an external stylesheet in bootstrap (sorry I belong to the Material UI category of users ^^). You could maybe use inline `style` prop for the moment, don't lose to much time on this this is just a first iteration of the backoffice, we can improve that kind of stuff later.

PR #2514, opened 2020-01-30
fix call to sortBy in Datatable.jsx

yairtal:
With out this dot the app will crash with an error

### DIFF
--- a/packages/vulcan-ui-material/lib/components/core/Datatable.jsx
+++ b/packages/vulcan-ui-material/lib/components/core/Datatable.jsx
@@ -264,6 +264,7 @@ const getColumns = (columns, results, data) => {
     const convertedColums = columns.map(column =>
       typeof column === 'object' ? column : { name: column }
     );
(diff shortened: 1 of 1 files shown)


eric-burel:
Hi, can you give a shot at my commit? Using `_.sortBy` with Underscore should be deprecated in favour to explicitely import functions from Lodash. Otherwise you end up with a zillion ko imports of Underscore + a dependency to magic global variables, it's a source of bug in Meteor apps

yairtal:
app crashes with the following error:

```
ReferenceError: sortedColumns is not defined
at getColumns (packages/vulcan:ui-material/lib/components/core/Datatable.jsx:267:5)
at DatatableContents (packages/vulcan:ui-material/lib/components/core/Datatable.jsx:328:23)
```

yairtal:
fixed it, works fine

eric-burel:
My bad I deleted it (tried to update the file from Github UI...).
Thanks!

PR #2519, opened 2020-02-06
Enhancement/forms accessibility

juliensl:
Hi ! 

I add some aria-labels for the accessibility ;)


eric-burel:
Blind persian users of Vulcan are thanking you :tada: :iran:

PR #2521, opened 2020-02-10
Add useSignout hook to the accounts package

Apollinaire:
Also added `@apollo/react-hooks` as a dependency since it's also used elsewhere in the codebase.
The hook simply takes a callback and does the same as the already existing logout function from `LoginFormInner` from the same package

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     }
   },
   "dependencies": {
+    "@apollo/react-hooks": "^3.1.3",
     "@babel/runtime": "7.1.2",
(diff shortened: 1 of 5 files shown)


eric-burel:
Thanks, I will include this in #2530 so we have a unified syntax

eric-burel:
Thanks, I am renaming it `useMeteorLogout`. Now we can handle logout with GraphQL only so I want to differentiate it from the usual `Meteor.logout` call (which basically does the same as the gql call, except that it cuts DDP connections too).

PR #2522, opened 2020-02-13
Check if user is admin before access to back office

yannick-mayeur:


eric-burel:
You should give keep the Layout accessible, because guest user may want to auth as admin on the backoffice home. You could use the HOC on collection pages instead, so the backoffice index is available but not other routes.

You may also want to hide the collection list for non admins (I thought the menu system came with "groups"). Let me know if you don't have time to handle this it's ok.

eric-burel:
I've done the change and merged in devel

PR #2529, opened 2020-02-18
Update to meteor 1.19

eric-burel:
Update did work after updating `fourseven:scss` to `4.12.0`
However the app do not run yet, `npm install` fails.
A package is trying to install `bcrypt@3.0.2` while we need > 3.0.6 for node 12 but I did not find who is responsible (most probably a meteor package) => seems solved by reinstalling

But I get `Cannot find moduleVulcan/.meteor/local/build/programs/server/node_modules/@babel/runtime/helpers/objectSpread2'` when running the app.

Most issues will be coming from Meteor using Node 12 so some packages need updates.

There is also a `feature/meteor-1.19` branch in the starter


eric-burel:
Ok I've got a working version. I propose to wait that some users can test this after upgrading to 1.14.1 before merging, because we may have version conflicts in real apps with many npm packages.

eric-burel:
Latest devel includes 1.10 directly

PR #2530, opened 2020-02-24
Account/auth features available outside DDP using GraphQL

eric-burel:
- [X] Login draft OK
- [X] Logout draft OK
- [x] Signup draft => createUser is meant for the admin to create the new user. We need a specific endpoint for signup. OK
- [X] verification and forgotten password draft => to be tested though
- [ ] Enrollment => not yet coded
- [x] test verification and email workflow
- [ ] Meteor attemptLogin validation (logic to avoid brute forcing passwords by limiting rate for example) => note that Meteor simple rest doesn't do that either so we may merge without this feature
- [x] CORS
- [X] Handle same origin in cors (needs to handle "origin === undefined")
- [x] Demo with an external app, eg with Next or Gatsby
- [X] Document GraphQL endpoints => see packages/vulcan-users/test/server/mutation.test.js for sample queries
- [ ] Create a new account ui that uses only GraphQL (and keep the old one for Meteor apps!)
- [x] Correctly split between queries and mutations (most are mutations)
- [ ] Create reusable hooks (will probably come from the NPM new version of Vulcan) => useSignOut already done by Apollinaire https://github.com/VulcanJS/Vulcan/pull/2521
- [X] fix body parser with "application/graphql" => see relevant PR
- [ ] Finish unit tests

Note: this authentication mode through graphql is NOT compatible with DDP. If you actually use Meteor methods and publication system, instead use the traditionnal auth workflow with Accounts. Vulcan's login form will probably keep using Meteor methods.
This features targets other types of non-meteor clients like  React Native, Next and so on.

This is an experimental feature, **not yet secured**. Use at your own risk and take time to read the code. Inputs welcome :)

To test: test should be doable using directly GraphQL Playground. To pass a token: provide HTTP header

Relevant discussion:
https://github.com/VulcanJS/Vulcan/issues/1896
See example usage (not yet done but still a good demo of how to consume this API):
https://github.com/VulcanJS/vulcan-next-starter
Reusing Meteor server methods:
https://github.com/meteor/meteor/issues/10937


eric-burel:
Almost done! I only need to test more advanced endpoints using GraphQL Playground.

PR #2538, opened 2020-02-28
Add skipPermissionCheck field to context and repsect it in to validateDocument and validat…

doronrk:
…eModifer 

This behavior defaults to false.


SachaG:
Should this be on the context though? That feels weird since the context is not specific to a mutation or a permission check. I'd probably pass a fourth `options` argument and put the property on that instead, and then base that on an argument of the mutator function (maybe have `validateData` and `validatePermissions` while keeping the current `validate` for backwards compatibility, which would do both?)

PR #2543, opened 2020-03-10
Update to current valid strikethrough tags

kevinashworth:
The `strike` tag is no longer a valid HTML tag. See https://www.w3schools.com/tags/tag_strike.asp

Two very similar tags are its replacement, `s` and `del` -- and `del` has a corresponding `ins` tag. See https://www.w3schools.com/tags/tag_del.asp and https://www.w3schools.com/tags/tag_s.asp

So take out the old `strike` tag and add in `s`, `del` and `ins`.

### DIFF
--- a/packages/vulcan-lib/lib/server/utils.js
+++ b/packages/vulcan-lib/lib/server/utils.js
@@ -6,7 +6,7 @@ Utils.sanitize = function(s) {
   return sanitizeHtml(s, {
     allowedTags: [
       'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'blockquote', 'p', 'a', 'ul',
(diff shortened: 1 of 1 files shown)


eric-burel:
Thanks! That's the kind of stuff that should probably handled by some 3rd party package nowadays. Do you know why we need an explicit tag list?

eric-burel:
For example there is no SVG. I don't know where this function is used btw.

kevinashworth:
I came to edit this function because apparently there is no official Markdown / Remark / etc support for strikethrough. 

But when I added an HTML “s” tag in a text area to compensate — and it should, as HTML is valid Markdown — it didn’t come through and I found that this sanitize function had removed it.

PR #2544, opened 2020-03-12
Bugfix/modaltrigger

EloyID:
Changes
a) The way the DatatableFilter passed the filter symbol react object didn't let add onClick, now it can
b) Material UI ModalTrigger accepts trigger as prop ass bootstrap's does

### DIFF
--- a/packages/vulcan-core/lib/modules/components/Datatable/DatatableFilter.jsx
+++ b/packages/vulcan-core/lib/modules/components/Datatable/DatatableFilter.jsx
@@ -62,7 +62,7 @@ const DatatableFilter = props => {
           />
         }
         size="small"
-        trigger={<Filter count={getCount(columnFilters)} />}>
+        trigger={Filter({ count: getCount(columnFilters) })}>
         {query ? (
           <Components.DatatableFilterContentsWithData {...props} />
         ) : (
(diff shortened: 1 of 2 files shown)


packages/vulcan-core/lib/modules/components/Datatable/DatatableFilter.jsx
```diff
@@ -62,7 +62,7 @@ const DatatableFilter = props => {
           />
         }
         size="small"
-        trigger={<Filter count={getCount(columnFilters)} />}>
+        trigger={Filter({ count: getCount(columnFilters) })}>
```
eric-burel: I don't get this change sorry. Why would you need to call the React component directly as a function? That could provoke performance issuses

eric-burel: More precisely what's the behaviour/error you encounter when directly using a React component?

eric-burel:
Can I close this PR and merge only #2545 ?

packages/vulcan-core/lib/modules/components/Datatable/DatatableFilter.jsx
EloyID: Il was the only way for the trigger to accept the onClick to open the modal.

eric-burel:
Merged through #2545

PR #2545, opened 2020-03-12
Bugfix/modaltrigger2

EloyID:
Change modaltrigger pattern to assure onClick acceptation and accept trigger prop


eric-burel:
Excellent PR, merged manually and added relevant stories. Thank you !

PR #2552, opened 2020-03-24
bugfix for the input api so it can accept more than one input for a samme field

EloyID:
Before, if your input was
{_gte : 1, _lte : 5} 
your output wqs
{$gte:1}
now it is
{$gte:1, $lte:5}

### DIFF
--- a/packages/vulcan-lib/lib/modules/mongoParams.js
+++ b/packages/vulcan-lib/lib/modules/mongoParams.js
@@ -84,21 +84,25 @@ export const filterFunction = async (collection, input = {}, context) => {
     { title_intl.value: { $in: ["foo", "bar"] } }
   
     */
-  const convertExpression = fieldExpression => {
+   const convertExpression = fieldExpression => {
     const [fieldName] = Object.keys(fieldExpression);
-    const [operator] = Object.keys(fieldExpression[fieldName]);
-    const value = fieldExpression[fieldName][operator];
-    if (isEmptyOrUndefined(value)) {
-      throw new Error(`Detected empty filter value for field ${fieldName} with operator ${operator}`);
-    }
-    const mongoOperator = conversionTable[operator];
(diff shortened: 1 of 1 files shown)


eric-burel:
Can you precise the test case? You mean in an `_and` feature? The full input would be great so I can easily add a unit test comparing input/expected result and validate the PR.

EloyID:
A) Maybe it should be an issue to discuss about. Now to do what I explained we should do
{_and:[{_gte : 1},{ _lte : 5}]}
B) And currently for your input {_gte : 1, _lte : 5} your output is {$gte:1}.
In B) we are missing information (_lte) and thats a pity. For users using normally mongo, having _and by default is very useful, i.e., 
{_gte : 1, _lte : 5} =>{myField: {$gte:1, $lte:5}} === {$and:[{myField:{$gte:1}},{myField: {$lte:5}}]}

eric-burel:
So your test case is just `{_gte : 1, _lte : 5}` which did not work?

EloyID:
Je n'ai pas testé si `{_and:[{_gte : 1},{ _lte : 5}]}` elle marche
My test case is just `{_and:[{_gte : 1},{ _lte : 5}]}`
Before 
A) fieldExpression was i.e. `{_gte : 1, _lte : 5}`
B) line 90 : `const [operator] = Object.keys(fieldExpression[fieldName]);`
`Object.keys(fieldExpression[fieldName])` vaut `['_gte','_lte']`
So with B operator will be just `_gte` and we lose half of the input.

eric-burel:
Great PR, thanks! I've added a test file, so far almost empty but that's a basis.

I think this feature was not implementend because "_and" operator is already sufficient to write this filter. However I agree it's not intuitive for Mongo user, so we may try to have selectors as similar as mongo selector as can be.

PR #2554, opened 2020-03-24
move muicheckboxgroup proptypes name and options to inputProperties a…

EloyID:
…nd name is not required anymore


eric-burel:
Hmm it sounds like a breaking change, you may want to keep them as facultative fields? Are you sure it doesn't break existing code relying on this?

EloyID:
If you look at the current file, name and options are extracted from props.inputProperties and the object this.props is never passed to a child object. In my opinion it made nonsense before
Regarding name not required anymore, I am using it without giving a name and it works correctly.

eric-burel:
Okay, you know better about MUI components anyway I trust you on this

PR #2555, opened 2020-03-24
Enhancement/datatable

EloyID:
Improvements : 
- Delete 
- Select 
- Export 
- Add custom aboveComponent in the datatable
- Change submit filter from a to button because javascript:void(0) causes errors in react


eric-burel:
@SachaG you may be interested by this PR, it open sources improvements to the datatable from Live for Good

EloyID:
If you have any suggestion to improve the system, I'll be glad to include them. I would like this commits to join Vulcan, so the framework can evolve with them.

eric-burel:
To be continued in #2558

PR #2561, opened 2020-04-14
Don't run `options` function until `loading` is false, and `data` is populated

kevinashworth:
This structure is necessary for my app to not crash when it loads a SmartForm. 
It is also in keeping with the code in Apollo example. See https://www.apollographql.com/docs/react/api/react-hooks/#example

### DIFF
--- a/packages/vulcan-forms/lib/components/FormComponentLoader.jsx
+++ b/packages/vulcan-forms/lib/components/FormComponentLoader.jsx
@@ -26,7 +26,7 @@ const FormComponentLoader = props => {
     }
   }
 
-  if (loading) return <Components.Loading />;
+  if (loading) return <div className="form-component-loader"><Components.Loading /></div>;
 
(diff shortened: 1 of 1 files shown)


eric-burel:
Thanks! It makes sense, so I merge already, but why does it fail in the first place? Shouldn't the options function be robust to missing data? Can you maybe share errors or code sample?

kevinashworth:
It has to do with dynamically loaded `options`, and the delay in fetching them from the database.

kevinashworth:
Sacha had a different solution in mind. (It may not need to be one or the other. Both could work?) Be sure to see some discussion at https://vulcanjs.slack.com/archives/C2LJQHTLP/p1586661514092700 and https://vulcanjs.slack.com/archives/C02L990AF/p1586869590077600

eric-burel:
Ok let me know, anyway your PR makes it more efficient so we can keep it this way (no need to compute options if you know you don't have data yet)

PR #2568, opened 2020-05-07
Forms submit target verification

Julien-Sytadelle:
When I open multiple SmartForms, submit one submits all forms.

For example, I open a first form to edit/create an object, then open another form to create another object to fill the first form.

As the SmartForm id prop is passed to the form node, I added a check for id equality between target and form in submitForm only if an id is passed to the smartForm.


Julien-Sytadelle:
Needs more work...

eric-burel:
And a `stopPropagation` is not sufficient?

Julien-Sytadelle:
Indeed, I learned something today. Thanks.

PR #2574, opened 2020-05-16
bugfix - SmartForm cannot update nested input fields

kevinashworth:
Reverse commit 594bf423bc86aa83061b53866f6fcefa6415be03

SmartForm changes from a week ago cause an error — cannot create or update nested inputs, like those in the Addresses subschema of VulcanStarter example-forms.

Commit 594bf423bc86aa83061b53866f6fcefa6415be03 introduced the error. When I reverse just this one commit, I am once again able to update nested fields.


SachaG:
Thanks!

PR #2583, opened 2020-06-07
Jun2020BugFixes1

ErikDakoda:
* `meteor_login_token` was not set with the correct sameSite attribute
 * `instantiateComponent()` did not handle `React.forwardRef` components correctly
 * `restrictDocument()` would fail for nested fields without a fieldSchema


eric-burel:
Thanks!

PR #2587, opened 2020-06-14
change cookie secure to false on localhost

yairtal:
fix for: on localhost, on chrome, after refreshing the browser, user will get logged out
see: https://github.com/reactivestack/cookies/issues/233


eric-burel:
Awesome I wouldn't have guessed that.

eric-burel:
Learnt something today about cookies, thanks ;)

PR #2589, opened 2020-06-20
add new update check api to ui-material Card.jsx

yairtal:
using ui-material Card component with options will crash the app, changed to the new API


packages/vulcan-ui-material/lib/components/core/Card.jsx
```diff
@@ -169,8 +170,8 @@ const styles = theme => ({
 const Card = ({ className, collection, document, currentUser, fields, classes }, { intl }) => {
   
   const fieldNames = fields ? fields : _.without(_.keys(document), '__typename');
-  const canUpdate = currentUser && collection.options.mutations.update.check(currentUser, document, { Users });
-  
+  const canUpdate = currentUser && get(collection, 'options.mutations.update.check');
```
eric-burel, line 173: Hi, this seems to just disable the check, since it's not called anymore?

PR #2590, opened 2020-06-24
[WIP] Api schema

eric-burel:
@SachaG I've opened this PR for apiSchema branch so we can check diffs more easil

New features:
- datatable selection
- access control on routes with AccessControl component
- database debug
- improvements on vulcan:email
- api schema

### Broken stuffs 
- `FormGroupNone` and `FormGroupLine` stories are broken
- Material UI probably not up to date, I need to check if it's broken
- @SachaG Default fragment seems to often remove "_id", is that normal? It's like its not included into "filterable" or smth like that.
- @SachaG `Expected substring: "type Foo { intlField(locale: String): String @intl intlField_intl(locale: String): [IntlValue] @intl }"
 Received string:    "# Type for Foos type Foo { intlField(locale: String): String @intl intlField_intl: [IntlValue] }` => now there is a `locale` argument to `intlField_intl`, is that expected? That's a breaking change when you don't provide this parameters or is it optional?

Otherwise other unit tests are ok. We could add some for the new features of apiSchema.

### Need unit test
- The "filterable" field in packages/vulcan-lib/lib/server/graphql/collection.js (that's a quite critical feature)


packages/vulcan-core/lib/modules/components/App.jsx
```diff
@@ -46,7 +46,9 @@ const RouteWithLayout = ({ layoutComponent, layoutName, component, currentRoute,
         const children = (
           <ErrorCatcher>
             <Components.RouterHook currentRoute={currentRoute} />
-            {React.createElement(component, childComponentProps)}
+            <Components.AccessControl currentRoute={currentRoute}>
```
eric-burel, line 49: Sounds nice!

packages/vulcan-core/lib/modules/decorators/autocomplete.js
```diff
@@ -2,9 +2,9 @@ import { getCollectionByTypeName, isEmptyOrUndefined, fieldDynamicQueryTemplate,
 import get from 'lodash/get';
```
eric-burel, line 3: Just a details but we could put decorators into "vulcan:lib", as the other stuffs related to schema are living there

PR #2592, opened 2020-06-29
Feature/showother ui material

yairtal:
Bug fix:
1. `showOther` for radio groups won't work with out disregarding the `allowedValues`
2. fix a bug with radio group that rightfully expects a string but sometimes gets an array (when back to update form)

Features:
1. showOther support for checkbox group at ui-material
2. showOther support for radio group at ui-material

### DIFF
--- a/packages/vulcan-core/lib/modules/decorators/radiogroup.js
+++ b/packages/vulcan-core/lib/modules/decorators/radiogroup.js
@@ -1,4 +1,8 @@
+import get from 'lodash/get';
+
 export const makeRadiogroup = (field = {}) => {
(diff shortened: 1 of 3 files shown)


SachaG:
I don't use MUI myself but if others say this is ready to go I can merge it?

yairtal:
It's basically the same logic as ui-bootstrap, @eric-burel or @ErikDakoda might want to review it

eric-burel:
Sounds right, however I'd like to take time to factor the code a bit because most of the logic does not seem MUI specific.

yairtal:
Sure, you're welcome to PM me if there is anything I can do to help

PR #2594, opened 2020-06-29
add itemProperties openNested option for nested schema

yairtal:
Add the option to have a nested schema on mount, this way the user doesn't have to click the + button in order to add the first item, he/she can just start typing in the data.
```
itemProperties: {
   openNested: true
}
````

### DIFF
--- a/packages/vulcan-forms/lib/components/FormNestedArray.jsx
+++ b/packages/vulcan-forms/lib/components/FormNestedArray.jsx
@@ -28,7 +28,7 @@ class FormNestedArray extends PureComponent {
   }
 
   addItem = () => {
-    const { prefilledProps, path } = this.props;
+    const {prefilledProps, path } = this.props;
     const value = this.getCurrentValue();
     this.props.updateCurrentValues(
       { [`${path}.${value.length}`]: _get(prefilledProps, `${path}.$`) || {} },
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by eric-burel** [comment]
Hi, does passing `prefilledProps` directly to the SmartForm works? I'd rather go for the solution that require less changes in the code if this use case is already covered by props

**Comment by yairtal** [comment]
Didn't work for me, the item stays closed. Please let me know if I missed anything.

**Comment by yairtal** [comment]
Also in my use case I'd prefer it to stay empty, simply open, so the user doesn't have to click the +

**Comment by eric-burel** [comment]
@yairtal can you please send me a snippet with prefilledProps that did not work? 
In the form example I can have this:
image


With those props (used in a Datatable): 
```
![image](https://user-images.githubusercontent.com/7983005/86384352-e5807380-bc8e-11ea-88d1-f9fd9fbeaf39.png)

```

eric-burel:
Closing in the meantime so I don't forget but of course I'll reopen if you have any additional info

eric-burel:
Maybe it messes up placeholders though? But your solution would have the same limitation as you programmatically do the same thing as having prefilledProps with an empty object.

yairtal:
This didn't work in my case since my case is an Edit/Update case, I have a `documentId` prop in my `SmartForm`.

eric-burel:
Ok I get the use case that makes sense.

eric-burel:
I am not a huge fan of adding more and more options to the form but I know this one is pretty common. I hope to make it easier to build custom forms based on the Vulcan components instead of systematically having to tweak the whole form behaviour to add such features, it should be easier in the first palce.

PR #2598, opened 2020-07-06
add useMulti2 to Datatable.jsx

yairtal:
useMulti2 on Datatable at ui-material

allows new filter API


yairtal:
Errors and bugs found, ignore this PR

PR #2601, opened 2020-07-28
Bugfix/add correct selector to delete

kevinashworth:
Deleting documents currently doesn't happen. I get an error in the console: "Error: GraphQL error: Selector cannot be empty". 

(It can easily be replicated by running the `example-forums` package, adding a comment to a post, then trying to delete that comment.)

I am guessing that this error stems from [this commit](https://github.com/VulcanJS/Vulcan/commit/314e1d99800cb366c403c59fa5c168a3f175ff16), where the delete mutator didn't get the correct selector. So here is a quick edit to `Form.jsx` to resolve this error.

NOTE: I also removed `contextName` for deletions in `mutations.js`, because I couldn't see how contextName is used when you delete a document. I could be wrong about this, but pretty sure about the need for the other edit to Form.jsx

### DIFF
--- a/packages/vulcan-forms/lib/components/Form.jsx
+++ b/packages/vulcan-forms/lib/components/Form.jsx
@@ -999,7 +999,11 @@ class SmartForm extends Component {
     const deleteDocumentConfirm = this.context.intl.formatMessage({ id: 'forms.delete_confirm' }, { title: documentTitle });
 
     if (window.confirm(deleteDocumentConfirm)) {
-      this.props[`delete${this.props.typeName}`]({ documentId })
+      this.props[`delete${this.props.typeName}`]({
+          input: {
+            id: documentId
+          }
+        })
         .then(mutationResult => {
           // the mutation result looks like {data:{collectionRemove: null}} if succeeded
           if (this.props.removeSuccessCallback) this.props.removeSuccessCallback({ documentId, documentTitle });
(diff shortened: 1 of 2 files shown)


kevinashworth:
I'm seeing warnings as well for refetch. "Cannot read property 'refetch' of undefined" 

Does https://github.com/VulcanJS/Vulcan/blob/f43a2b509b9b656a733362dce662214c371ebf3e/packages/vulcan-forms/lib/components/Form.jsx#L1006 still make sense?

eric-burel:
The cannot read 'refetch' of undefined message does not make sense, is "removeSuccessCallback" triggering the error too?
@SachaG

eric-burel:
Thanks for the PR, I am merging without testing but I think it's right, remove was just forgotten in the process as we do not use it that often

kevinashworth:
> The cannot read 'refetch' of undefined message does not make sense, is "removeSuccessCallback" triggering the error too?

I do not seeing any errors for `removeSuccessCallback`. I supply that function in my SmartForm and it works without error. But I use `history.push` in `removeSuccessCallback` to go to another route. Is that why `this.props` on Line 1006 is undefined? (On the other hand, I don't understand the point of `refresh`ing a form after deleting its contents.)

eric-burel:
Yeah very weird I cant get what happens. I think the code just fails before the this.props.removeSuccessCallback is called any way.

PR #2608, opened 2020-08-15
FormGroupTestBackground

ErikDakoda:
When the document being edited has a `test` field set to true, the background of form groups is changed using theme.utils.testPaper


eric-burel:
Sorry, I didn't get where this `test` field is coming from? Is that a convention/feature I am not aware of?

ErikDakoda:
Hi @eric-burel, you are right, the `test` field is specific to my code so I am moving this change to my own codebase. Thanks!

PR #2613, opened 2020-08-16
GenericMutationWrappersResponse

ErikDakoda:
The generic create, update, and delete mutation (HoC/Hooks) wrappers would return their result in `response.data.<resolverName>`. This commit restored the name of those fields.


eric-burel:
Thanks, I'd advise to use the new create2 containers however, v1 stuff is deprecated.

PR #2617, opened 2020-08-16
ValidateDocumentPermissionsBugFix

ErikDakoda:
`validateDocumentPermissions` would throw an error when the document field was `isNested` but had no `fieldSchema`

### DIFF
--- a/packages/vulcan-lib/lib/modules/validation.js
+++ b/packages/vulcan-lib/lib/modules/validation.js
@@ -28,7 +28,7 @@ const validateDocumentPermissions = (fullDocument, documentToValidate, schema, c
   const { Users, currentUser } = context;
   forEachDocumentField(documentToValidate, schema,
     ({ fieldName, fieldSchema, currentPath, isNested }) => {
-      if (isNested && (mode === 'create' ? !fieldSchema.canCreate : !fieldSchema.canUpdate)) return; // ignore nested without permission
+      if (isNested && (mode === 'create' ? !fieldSchema?.canCreate : !fieldSchema?.canUpdate)) return; // ignore nested without permission
       if (!fieldSchema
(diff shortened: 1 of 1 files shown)


packages/vulcan-lib/lib/modules/validation.js
```diff
@@ -28,7 +28,7 @@ const validateDocumentPermissions = (fullDocument, documentToValidate, schema, c
   const { Users, currentUser } = context;
   forEachDocumentField(documentToValidate, schema,
     ({ fieldName, fieldSchema, currentPath, isNested }) => {
-      if (isNested && (mode === 'create' ? !fieldSchema.canCreate : !fieldSchema.canUpdate)) return; // ignore nested without permission
+      if (isNested && (mode === 'create' ? !fieldSchema?.canCreate : !fieldSchema?.canUpdate)) return; // ignore nested without permission
```
eric-burel, line 31: We should probably do "isNested && fieldSchema && (...)".
`?.` is a very error-prone, abstract, syntax that we might want to use sparsely, eg on a trivial condition.

eric-burel: Also does it happen on a valid schema? I guess if `isNested` is true we should definitely have a schema for the field?

ErikDakoda: @eric-burel This was an issue when I first upgraded to Vulcan 1.15. I have since refactored my schemas, and I can not reproduce this issue anymore. I wonder if we should still make this change to avoid problems for those who may upgrade in the future - or I can close this PR. What do you think?

eric-burel: Yeah I think we can close, if it's still an issue I think the fix would be in "forEachDocumentField" helper that should not call the callback on invalid data eg with isNested but no schema.

ErikDakoda: Agreed. Thanks!

PR #2621, opened 2020-08-18
CurrentUserRefetch

ErikDakoda:
* `currentUser` HoC/Hook now returns an extra parameter: `currentUserRefetch`


GraemeFulton:
This is very useful! I needed to refetch the user after they were added to the paying user group, and this helped lots!

PR #2622, opened 2020-08-19
Validate document permissions bug fix

ErikDakoda:
* `validateDocumentPermissions` would throw an error when the document field was `isNested` but had no `fieldSchema`
 * This is the case for a nested schema that has a graphql-only field generated using `resolveAs`
 * This will be better handled once people start using apiSchema Fields


ErikDakoda:
Hi @eric-burel! I couldn't reproduce this, but of course, as soon as I pushed my code to production it was obvious. Let me know if you need further explanation.

eric-burel:
Ok makes more sense! I'll merge, in the near future we might review schema helpers such as this one as we will need them more and more.

PR #2624, opened 2020-08-20
update ui-material datatable to withMulti2

yairtal:
After upgrading to the `withMulti2` HoC I add the Datatable's input to `listOptions`, including the search query

**Please note that `orderBy` and `terms` were removed, so this might be a breaking change for some users**

My apologies in advance for upgrading the HoC instead of refactoring it to hooks

### DIFF
--- a/packages/vulcan-ui-material/lib/components/core/Datatable.jsx
+++ b/packages/vulcan-ui-material/lib/components/core/Datatable.jsx
@@ -5,7 +5,7 @@ import {
   replaceComponent,
   withCurrentUser,
   Utils,
-  withMulti,
+  withMulti2,
   getCollection,
   instantiateComponent,
 } from 'meteor/vulcan:core';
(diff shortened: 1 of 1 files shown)


eric-burel:
Is there anything blocking this PR? If this works ok in your app I'll merge.
Is sorting still ok since `orderBy` is removed as you said?

yairtal:
did try sorting yet, but again - this is a breaking change!

I am using it on my production app, but would love if you'll test it before merging

eric-burel:
I don't really have any environment to test so that's why I prefer to "blind merge" if you run it in production already. The breaking change is ok here, old versions of "multi" etc. are deprecated anyway. I'll wait for some bugfixes to be merged though, but then we can merge safely.

Maybe @ErikDakoda is this change fine for your applications?

@yairtal you might want to also update the "MIGRATING" documentation too then with this breaking change

ErikDakoda:
Hello, @yairtal and @eric-burel! My recommendation is that you commit the updated Datatable component as a new Datatable2 component and let each dev migrate to it at their own pace.

I just checked my code and I have 101 `addView()` statements in my codebase that need to be migrated to `inputs`. I feel that we need to give people a longer migration window.

Neobii:
Should be `useMulti2`.

PR #2628, opened 2020-09-02
Add Basic PWA support

harryadel:
As discussed in https://github.com/VulcanJS/Vulcan/issues/2239, this PR is the first step towards implementing PWA support in Vulcan. Please tell me what do you think and what can be improved. Thanks! 
cc @eric-burel


packages/vulcan-lib/lib/modules/fragments.js
```diff
@@ -89,23 +89,23 @@ export const getDefaultFragment = collection => {
   return fragmentText ? gql`${fragmentText}` : null;
```
eric-burel, line 89: Can you please rollback changes to this file? This is just a formatting thing but it bloats your PR

eric-burel:
Sounds pretty good ! I am not a huge mobile user and I don't know much about PWAs, I'll link your pr on the Slack

packages/vulcan-lib/lib/modules/fragments.js
harryadel: The reason I ventured into this file because `string-similarity` module was not found. And to test those changes I had to revert all the way back to 1.14.0!
Maybe my way of cloning the repo wasn't proper, I don't know. :man_shrugging:

SachaG: you can just do `yarn add string-similarity` to install that module.

harryadel: `string-similarity` not found module wasn't the only issue though :sweat_smile: and it's better not to bloat this PR with unrelated changes/fixes especially that I'm unsure what caused such issues

eric-burel:
@harryadel  What's the state for this PR? I think you solved the reviews comment so I can probably merge? Did you test it recently on one of your app?

packages/vulcan-lib/lib/client/serviceWorker.js
```diff
@@ -0,0 +1,8 @@
+export const createServiceWorker = () => {
```
Apollinaire, line 1: I would prefer calling it `registerServiceWorker`

public/manifest.json
```diff
@@ -0,0 +1,21 @@
+{
+  "short_name": "vulcan",
+  "name": "Vulcan",
+  "start_url": "/",
```
Apollinaire, line 4: Since we are in the Vulcan repo, this along with the images will probably not be included in the applications. I think it's ok to let it here as an example though.

packages/vulcan-lib/lib/client/serviceWorker.js
Apollinaire: Also, this needs to be opt-in via a setting. I don't want a service worker shoved in my app without explicitely consenting to it;

Apollinaire (changes requested):
Some issues with this PR: 
- I refuse to impose a service worker to everyone, so registering it should be opt-in via a setting, i.e. it does not register anything unless a boolean is set to true.
- The Head tags need to be inside a setting too.
- Finally, every file that has been added to /public will not be deployed to Vulcan apps, because for that it needs to be inside the app's /public folder, not the core modules.

So this may work as a proof of concept, but it is far from being ready to merge.

PR #2637, opened 2020-09-19
Use Symbol.for for vulcan-accounts const STATES

kevinashworth:
Instead of `Symbol()`, `Symbol.for()` will allow global access to Symbols like `STATES.SIGN_IN` in the other files in a two-repo Vulcan install. This allows customization (i.e. `replaceComponent`) of the `AccountsStateSwitcher`, for example.

More info on `Symbol.for()` from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol: 
> the Symbol() function will not create a global symbol that is available in your whole codebase. To create symbols available across files ... use the method Symbol.for()...

On the other hand, if there is no use of the `STATES` const outside of this `helpers.js` file, I believe there is no harm in using the global Symbol registry anyway.

### DIFF
--- a/packages/vulcan-accounts/imports/helpers.js
+++ b/packages/vulcan-accounts/imports/helpers.js
@@ -6,12 +6,12 @@ try {
 }
 export const loginButtonsSession = Accounts._loginButtonsSession;
 export const STATES = {
-  SIGN_IN: Symbol('SIGN_IN'),
-  SIGN_UP: Symbol('SIGN_UP'),
-  PROFILE: Symbol('PROFILE'),
-  PASSWORD_CHANGE: Symbol('PASSWORD_CHANGE'),
-  PASSWORD_RESET: Symbol('PASSWORD_RESET'),
-  ENROLL_ACCOUNT: Symbol('ENROLL_ACCOUNT')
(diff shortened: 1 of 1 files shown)


eric-burel:
Interesting, I didn't know that. There is no harm either into using `Symbol.for`, thank you

eric-burel:
Doesn't it break the way you consume them though? With `Symbol.keyFor`? Did you test a Vulcan app after this change?

kevinashworth:
> Doesn't it break the way you consume them though? With Symbol.keyFor? Did you test a Vulcan app after this change?

In my two-repo install, I've modified `helpers.js` as shown in this PR, then I use `Symbol.for('SIGN_IN')` in my own codebase. I have no need for `Symbol.keyFor()`. To clarify, `Symbol.for()` creates symbols if they don't exist or returns the appropriate symbol if it does exist, and that's all the functionality I need to modify the Sign In process. As far as I can tell, `Symbol.for()` instead of `Symbol()` doesn't "break the way you consume them"; it allows them to be consumed in a two-repo install.

eric-burel:
Thanks for the details!

PR #2641, opened 2020-10-11
Update README.md

coffeebeani:


StorytellerCZ:
???
Looks like Hacktoberfest spam.

eric-burel:
Yeah we expect some, but I'd like to keep the hacktoberfest label as much as possible, we'll see if it becomes problematic.

eric-burel:
Btw the policy against such spam is "spam" label (which will block further submission to Hacktoberfest) + closed PR + signaling at GitHub

PR #2648, opened 2020-10-23
Minor Data Layer Fixes Oct 2020

ErikDakoda:
* Updated JSDoc comments for `Mongo.Collection.prototype.addField()` and fixed a minor bug.
 * Updated `validateModifier()` to include `documentId` in the context of custom field validators. Validators can use it to make sure that a value is unique in the collection. Without the `documentId`, the validator can't tell if an existing document with a unique value is being updated or if a new document with a conflicting value is attempted to be inserted.
 * Updated default resolvers to include the collection name when throwing a `missing_document` error.

### DIFF
--- a/packages/vulcan-lib/lib/modules/collections.js
+++ b/packages/vulcan-lib/lib/modules/collections.js
@@ -71,7 +71,7 @@ Mongo.Collection.prototype.attachSchema = function (schemaOrFields) {
 
 /**
  * @summary Add an additional field (or an array of fields) to a schema.
- * @param {Object|Object[]} field
+ * @param {Object|Object[]} fieldOrFieldArray
  */
 Mongo.Collection.prototype.addField = function (fieldOrFieldArray) {
   const collection = this;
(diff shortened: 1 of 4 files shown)


ErikDakoda:
Hi @SachaG! I noticed that you have reverted my change to `collection.addField()` in this [commit](https://github.com/VulcanJS/Vulcan/commit/be86cc7cda460f42a6bb770d316261368e7cda00).

I wonder if you meant to do that. I am not sure if you have seen my explanation for that change (please see above). Could we discuss this further? As I mentioned it looks to me that the usage of `collection.options.schema` is an outlier, as everywhere else in the code `collection.simpleSchema()._schema` is used, for example in removeField().

EloyID:
Hi I have the same issue due to this modification https://github.com/VulcanJS/Vulcan/commit/be86cc7cda460f42a6bb770d316261368e7cda00#diff-6bef4e846751805d4e7f0e4bc392cc8125125fdaae89046b5e0fc06233eb8e38
I think it's the same as ErikDakoda has mentionned. Is there any reason to do this modification?

PR #2651, opened 2020-10-24
Multi HoC/Hook Returned Props

ErikDakoda:
`withMulti`, `useMulti`, `withMulti2`, `useMulti2` should return all the props from Apollo Client's `useQuery`, including `startPolling`, `stopPolling`, etc.


eric-burel:
Thanks for all your PR! 

Note that active development are ongoing there for hooks: https://github.com/VulcanJS/vulcan-npm/tree/devel/packages/react-hooks
Technically, you can already use this package, though you might need to build it locally and all, and to manually convert the Vulcan collection into a "Vulcan Model" (the new architecture). I'll provide an helper to do so in a near future.

PR #2652, opened 2020-10-27
update permissions for debug schemas

Apollinaire:
Package: vulcan:debug
The previous permissions were `['admins']` which is problematic when you're not logged in. So now I've set it to `['guests', 'members']` (I was not sure that if you're logged, you're still part of the guests group). I also factorized it a little, putting all inside a single variable, located in a `permissions` file.


eric-burel:
Nice! `guests` will be renamed `anyone` in the future to avoid confusion, so yes guests encompass members too currently.

PR #2655, opened 2020-11-03
Material UI Minor Updates Nov 2020

ErikDakoda:
* MuiInput: Improved the debounce functionality and fixed some edge cases
 * EndAdornment: Fixed bug: When the form control had an empty value the clear button would be hidden, but it was not disabled
 * MuiCheckbox: Added support for start and end adornment in the checkbox base control and added additional classes that can be overridden for cusomization
 * MuiSwitch: Renamed `MuiSwitch` component to `SwitchBase` because its name conflicted with the style sheet name of the core `Switch` component - we should rename all components in `vulcan:ui-material` that start with Mui


eric-burel:
Thanks! Agreed for renaming, but the exported name should be kept the same to avoid breaking change, we still have a lot of `replaceComponent` at Live for Good for instance

PR #2658, opened 2020-11-16
React Node Support in getString()

ErikDakoda:
* In the values object you pass to `getString()` you can now use React nodes, in addition to simple strings, enabling the embedding of HTML tags, images, etc.


eric-burel:
Sounds cool. Do we have some existing unit test for this feature?

ErikDakoda:
Let me see if I can write some. I may need your help.

PR #2662, opened 2020-11-19
add to storybook plugin-proposal-optional-chaining to accept component.jsx

EloyID:
The use of the optional chaining `?.` in `component.jsx` stops storybook working, I added the babel plugin to make it work

https://github.com/VulcanJS/Vulcan/blob/devel/packages/vulcan-ui-material/lib/components/forms/base-controls/mixins/component.jsx

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -128,7 +128,8 @@ module.exports = ({ config }) => {
           presets: ['@babel/react', {
             plugins: [
               '@babel/plugin-proposal-class-properties',
-              '@babel/plugin-syntax-dynamic-import'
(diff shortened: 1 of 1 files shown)


eric-burel:
Awesome! Did you add it in package.json too?

EloyID:
The other plugin @babel/plugin-proposal-class-properties which was already present in the file is not so I did not add it. Do you want me to add to the pull request package.json with both dependencies ?

eric-burel:
Ha that's weird. I just want to make sure that there is no missing dependency, if it works ok to you when you reinstall `node_moduels` then that's fine and I'll merge

EloyID:
I have tested to delete node modules and reinstall and there was no problem :)

PR #2670, opened 2020-12-13
Feature RTL support in locale

fghamsary:
This is for the RTL Support the ticket I have created https://github.com/VulcanJS/Vulcan/issues/2669
Could you please check it carefully, as I don't have slightest idea about the impacts, but it's working fine on my local project for translations of StateOfJS ;) 
https://github.com/StateOfJS/state-of-js-graphql-results-api/issues/50
After checking the code, I had come to the conclusion that without this, the RTL support was not possible at all.
This is just the beginning and give the developer the opportunity by usage of RTL class in HTML to do custom CSS if needed ;)


SachaG:
Thanks, this looks good!

PR #2675, opened 2021-01-02
Random

ErikDakoda:
* Random is meant to be a drop-in replacement for Meteor's `random` package
 * https://docs.meteor.com/packages/random.html
 * So far only Random.id is implemented


eric-burel:
@ErikDakoda You'll need to add the package in Vulcan Starter as well and merge both in sync

PR #2676, opened 2021-01-02
Reactive State

ErikDakoda:
* Added a new mechanism for storing global state in Apollo Client 3 reactive variables
 * Implemented supporting functions in `reactive-state.js`
 * There is a pull request for the documentation: https://github.com/VulcanJS/vulcan-docs/pull/163
 * Another PR forthcoming, "Flash messages in reactive state" depends on this


eric-burel:
Sounds awesome ! I'll review more deeply later on but I already see unit tests. @Apollinaire @SachaG what's your opinion on this?

packages/vulcan-lib/lib/modules/reactive-state.js
```diff
@@ -0,0 +1,215 @@
...
+
+
+/**
+ * React HoC for using reactive state in a Class component
+ * @param {Object} options Hook options
+ *        {string} options.stateKey The key of the desired reactive state
+ *        {string} options.[propName] Name of the props injected by the hook; omit to use `stateKey`
+ * @returns {function} A wrapped React component that includes the props returned by `useReactiveState`
+ */
+export const withReactiveState = (options) => C => {
```
eric-burel, line 198: Do you use the HOC in your own app? Otherwise it can quickly become more maintenance, I'd expect peoeple to write their own HOC if they really need them.

packages/vulcan-lib/test/mongoParams.test.js
```diff
@@ -7,7 +7,7 @@ const test = it;
 const mayTheFourth = new Date('1977-05-04T22:00:00');
```
eric-burel, line 7: Unit tests 🥇

packages/vulcan-lib/lib/modules/reactive-state.js
```diff
@@ -0,0 +1,215 @@
...
+    if (skipDuplicate) return;
+    throw new Error(`There is already a reactive state object named ${stateKey}`);
+  }
+
+  if (schema) {
+    schema = createSchema(schema);
+    defaultValue = cleanReactiveStateValue(defaultValue || {}, schema);
+  }
+
+  reactiveStates[stateKey] = {
```
eric-burel, line 54: I am not fond of this register pattern anymore and would tend to deprecate them in existing application, even though those patterns are common in Vulcan. 
Is there a good reason to store all the vars in a global object? If they need to carry metadata with the object, maybe we can add those data directly in the created object? Like `const cartItemsVar = makeVar([]); carItemsVar._vulcan = { schema }; return carItemsVar;`.
The point being that you avoid to use a global register, which in turns avoid any risk of conflict and forces end user to clarify their dependencies with actual ES6 `import/export`. 

That's just a proposal though, if you feel good using it as is then I am fine with this implementation.

packages/vulcan-lib/test/reactive-state.test.js
```diff
@@ -0,0 +1,257 @@
+import expect from 'expect';
```
eric-burel, line 1: 💯

packages/vulcan-lib/lib/modules/reactive-state.js
ErikDakoda: Really I did it this way to follow the Vulcan pattern. I will try it without the global state and see how that feels.

ErikDakoda:
@eric-burel now that I have been using reactive state I am considering ditching `immutability-helper` for updating the state object and instead use a simple Object.assign() like React's `setState()` does. I could accept a stateChange object or a function that returns a stateChange object. Any thoughts?

packages/vulcan-lib/lib/modules/reactive-state.js
ErikDakoda: I do use HoC's primarily because most of my code was written before hooks existed. It's my opinion that for now we should provide both a hook and an HoC wrapper for all core functionality.

eric-burel:
@ErikDakoda what does it change in terms of code for the end usage ? Whatever you feel the most intuitive when coding is fine with me, and ditching a dependency is always good

ErikDakoda:
I will be replacing this PR with an updated one.

PR #2677, opened 2021-01-02
Flash messages in reactive state

ErikDakoda:
* An adaptation of Vulcan's flash message system, but using reactive state instead of React context
 * Updated `withMessages` to use reactive state
 * Added new `useMessages` hook
 * Added new `useIntl` hook
 * Separated `FlashMessages` component into its own file
 * Updated `Flash` component to use reactive state
 * Moved `getProperties()` from the Flash component to `normalizeMessage()` in withMessages
 * Added new Material UI implementation of the `Flash` component
 * Removed `messages` from the state of the `App` component, `MessageContext` from the component tree, and the methods `UNSAFE_componentWillMount`, `componentWillUnmount`, `flash`, and `clear`
 * Removed `messages.js`, which only defined `MessageContext`
 * Updated `error.js` to report the http status of network errors thrown by Apollo

NOTE: This PR depends on #2675 and #2676

### DIFF
--- a/packages/vulcan-core/lib/modules/components/AccessControl.jsx
+++ b/packages/vulcan-core/lib/modules/components/AccessControl.jsx
@@ -3,7 +3,7 @@ import { Components, registerComponent } from 'meteor/vulcan:lib';
 import { useCurrentUser } from '../containers/currentUser';
 import Users from 'meteor/vulcan:users';
 import { useHistory } from 'react-router-dom';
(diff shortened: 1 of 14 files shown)


ErikDakoda:
I will be replacing this pull request with an updated one.

eric-burel:
Eager to see that, thank for your work on this I didn't know this feature of Apollo and it's very interesting

PR #2682, opened 2021-01-20
Reactive State

ErikDakoda:
* Added a new mechanism for storing global state in Apollo Client 3 reactive variables
 * Implemented supporting functions in `reactive-state.js`
 * There is a pull request for the documentation: https://github.com/VulcanJS/vulcan-docs/pull/163
 * Another PR forthcoming, "Flash messages in reactive state" depends on this


ErikDakoda:
@eric-burel Thank you for your feedback on my last pull request. Based on your suggestions I made a few changes that I think improved and simplified the implementation of this feature:

 * I decided to keep the global state registry - primarily because I wanted to keep the ability to reset all reactive state when the user logs out (or any time the Apollo client store is reset)
 * I updated the documentation to encourage the use of import statements instead of relying on the global state registry
 * I decided to jettison `immutability-helper` for a more intuitive way to mutate object state - it now works very similarly to React's `setState()`
 * I decided to also jettison `useReactiveState` and `withReactiveState` and the documentation shows how to import and use reactive state in a component

eric-burel:
Sounds great to me! The code feels simplified indeed that's great.

eric-burel:
@ErikDakoda I'll let you merge yourself if you want to do a last check

PR #2687, opened 2021-01-23
SmartForm minor bug fixes Jan 2020

ErikDakoda:
* Fixed bug: `SmartForm.initField()` set each field's `document` property to the initial document, instead of the current document, resulting in bugs like [Nested Schema document update not working](https://github.com/VulcanJS/Vulcan/issues/2686)
 * Fixed bug: `FormComponent.getDerivedStateFromProps()` would fail when `intlOrRegularValue` was null (because `typeof null === 'object'`); it's possible that this should not normally occur, but it has occurred in my codebase because of Invariant Violation errors thrown by Apollo, likely because we have not upgraded to Apollo Client 3 cache
 * Fixed bug: Material UI Number input would not behave correctly when the user entered "0." quickly (the period would disappear)
 * Fixed bug: Creating a new document with SmartForm would fail when no Apollo errors occurred (`meta` prop was missing)
 * Fixed bug: `getString()` in `intl.js` would fail if the defaultMessage passed was not a string

### DIFF
--- a/packages/vulcan-forms/lib/components/Form.jsx
+++ b/packages/vulcan-forms/lib/components/Form.jsx
@@ -375,7 +375,6 @@ class SmartForm extends Component {
     // intialize properties
     let field = {
       ..._.pick(fieldSchema, formProperties),
-      document: this.state.initialDocument,
       name: fieldName,
       datatype: fieldSchema.type,
       layout: this.props.layout,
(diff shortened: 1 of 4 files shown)


ErikDakoda:
I would like someone (@eric-burel) to quickly review this, primarily because I am changing `SmartForm.initField()` to change `field.document` from the initial document to the current document. I believe this is a bug, but it's possible that some users of the code are expecting `field.document` to contain the initial document.

PR #2689, opened 2021-01-26
add graphql.context callback

Apollinaire:
Simply add a server-side callback called `graphql.context` that lets us iterate and modify the context passed to resolvers. It will run once per request, at graphql context initialization.


eric-burel:
Best one-liner PR of 2021 yet

eric-burel:
Don't you need the request as parameter somewhere?

Apollinaire:
request is already in context.req, so i thought it was useless to add it. However, now that you mention it, it would be a good idea to add it in, simply as a guarantee to pass it to every callback.

PR #2690, opened 2021-01-26
Connector filter calbacks

Apollinaire:
Adds a callback to Connector.filter, which allows to iterate on the `selector`, `options` and `filteredFields` used in a collection's multi query.

### DIFF
--- a/packages/vulcan-lib/lib/server/connectors/mongo.js
+++ b/packages/vulcan-lib/lib/server/connectors/mongo.js
@@ -36,11 +36,7 @@ DatabaseConnectors.mongo = {
     values that should always apply. 
 
     */
-    const defaultInputObject = await filterFunction(
-      collection,
-      collection.options.defaultInput,
-      context
-    );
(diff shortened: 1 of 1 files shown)


eric-burel:
I think for the documentation part, the best way at this point is to register the callback with a description

Apollinaire:
I can't register these callbacks here because the connectors are not initialized on the collection creation, they take the collection as a parameter: 
```js
DatabaseConnectors.mongo = {
  filter: async (collection, input, context) => {...}
}
```
We could register them inside `createCollection`, but it seems weird to register a callback so far from where it's called, that could lead to confusion. Also, the previous equivalent to these callbacks (`` `${typeName.toLowerCase()}.parameters` ``) were not registered. So I guess it's ok to leave it as it is. This is a pretty deep feature, so I guess it's fine to go without the documentation for this one.

eric-burel:
Yep agreed, feel free to merge

Issue #2664, opened 2020-11-22
Smartform shows delete button regardless of collection.permissions.canDelete


Apollinaire:
**Describe the bug**

SmartForm doesn't check the new API collection..permissions.canDelete to tell if it's possible to delete a document, and shows the delete button. 

**To Reproduce**
create a collection with `permissions: { canDelete: [] }`, or without it altogether. Use a smartform to edit a document of that collection, it will show the delete button. If you click it, the server will reject the mutation (thank god) , but the error is not handled. 

**Expected behavior**
The SmartForm checks that the user can delete the document and shows the delete button only if he can.

**Additional context**
v1.16.0 ; current master, commit 27d6f014d3c6c6c7ca756c91566c8452bc50c4a7

kevinashworth:
In [Form.jsx](https://github.com/VulcanJS/Vulcan/blob/c1bfcfc779cf1b1aa0c26d0476b4a1b759df7906/packages/vulcan-forms/lib/components/Form.jsx#L1162), there is this conditional:

`this.props.showRemove && this.props.showDelete`

Why does Vulcan have both `showRemove` and `showDelete`, and what is the difference between them?

Apollinaire:
There has probably been some prop renaming, and the old prop was kept for backwards compatibility.
The docs only mention showRemove, so I guess it's the one to use now.

PR #2694, opened 2021-02-24
Fixed issue #2664 + Fixed canUpdate on Card.jsx

fmiglianico:
#2664 
Fixed permission check to show/hide the delete button on SmartForms.
Also fixed the permission check for the edit button on Cards (was missing the document to be able to check ownership).

### DIFF
--- a/packages/vulcan-forms/lib/components/FormSubmit.jsx
+++ b/packages/vulcan-forms/lib/components/FormSubmit.jsx
@@ -2,6 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Components } from 'meteor/vulcan:core';
 import { registerComponent } from 'meteor/vulcan:core';
+import Users from 'meteor/vulcan:users';
+import get from 'lodash/get';
 
 const FormSubmit = ({
   submitForm,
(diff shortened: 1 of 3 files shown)


packages/vulcan-forms/lib/components/FormSubmit.jsx
```diff
@@ -12,12 +14,35 @@ const FormSubmit = ({
...
 }, {
   isChanged,
   clearForm,
-}) => (
+}) => {
+
+  // Delete permission check
+  let canDelete = false;
+
+  if (Users.isAdmin(currentUser)) {
```
eric-burel, line 29: Seems overkill, don't we have a function to do this check already (Users.can)?

fmiglianico: You're right, i'll simplify this. Should I keed the admin's bypass ?

eric-burel: I think it is already included in the existing Users functions (and it should be if it's not the case), so no you don't need to make it explicit normally

Apollinaire: You should check at the source code for the permissions in the users package, there are functions that should do the job in one line

packages/vulcan-forms/lib/components/FormSubmit.jsx
```diff
@@ -48,7 +50,12 @@ const FormSubmit = ({
       </a>
     ) : null}
   
-    {deleteDocument ? (
+    { deleteDocument && Users.canDelete({
+        context: { Users },
```
eric-burel, line 54: Thanks for the update! Are you sure you need to pass the context here?

fmiglianico: I wasn't sure, I copied it from Card.jsx where canUpdate was called with this parameter, but I guess it was for backwards compatibility which we don't need in this case. I fixed it.

eric-burel:
Thanks for this PR!

eric-burel:
For the record you can add "closes #PR" number so the ticket is closed automatically when the PR is merged

PR #2696, opened 2021-03-05
Feature/clean old api

eric-burel:
@ErikDakoda  this answers the issue of the old resolvers and new resolvers amiguously coexisting into the code.

I hope there are no regression due to updating the deprecated default resolvers. The old ones are still there anyway if we want to compare. Old API is still used in some parts of the code, it will be easier to spot them.

I couldn't test much because I encountered a weird build issue: after an install and `meteor npm run test`, I get `TypeError: registerCallback is not a function` (it's undefined), which doesn't make much since this function is well defined... No idea why, a reinstall did not change it.


ErikDakoda:
@eric-burel Wow, this is big! I'll have a look soon. Is this a breaking change, or is it supposed to be totally backwards compatible?

eric-burel:
Thanks! It will break existing applications yes, I've slightly updated the migrating doc and Sacha had published articles about the new API on vulcan blog. But this PR specifically is just renaming stuffs, this code already existed in Vulcan

Neobii:
If I read it correctly, it looks like you moved off the `2` type API to
without the 2 ie `useMulti2` --> `useMulti` and then `useMulti` to
`useMultiOld`.

ErikDakoda:
Wow, this is too soon for me. I have dozens of components and 100+ server-side queries that I have to still migrate from `terms` to `input`. I am using the new API when I create something new or when I need to touch/change something existing. It's been hard to keep up with the constant refactoring.

When I asked that question about default_resolvers vs default_resolvers2 - I had just realized that I am using the new API on collections that still use the old default_resolvers without any problems - so I was wondering what had changed.

eric-burel:
I think the default_resolvers basically does the same as the new, except they are also accepting the old API.
This PR does not remove the "old" resolvers hopefully but only renames them. But that also means a lot of renaming for your app, and I am not yet able to setup update scripts (some framework provides such "codemods", that would be worth digging).

I will focus on updating Meteor first because that's a less breaking change, and see if it solves my build issues.

ErikDakoda:
@eric-burel Are you able to build and run the unit tests after upgrading to Meteor 2?

eric-burel:
Just tried and yeah, it doesn't solve my build issues sadly

PR #2701, opened 2021-03-22
Update fixes

Neobii:
Various fixes for the meteor upgrade!

Also Changes from [this pull request](https://github.com/VulcanJS/Vulcan/pull/2697/commits/21ce25f3287e0eca12cd684a698835df87717046) without the lint changes.


eric-burel:
Nice thank you, what is Service Configuration btw?

Neobii:
That's to config the various Oauth things like facebook and twitter... I think you have to include it now...

If you use the exampleSettings.json 
```
"oAuth": {
    "facebook": {
      "appId": "foo",
      "secret": "bar"
    }
  },  ```
  this will give you an error now "configuration" is not defined
  
[Service Configuration](https://docs.meteor.com/api/accounts.html#service-configuration)

PR #2718, opened 2021-06-03
Bugfix/duplicate item with limit

IBRAHIMDANS:
# Description
BUG with `useMulti2` when I use limit, document is duplicated 
I add `uniqBy` to `lodash` to remove documents duplicated by `_id`

image


eric-burel:
Hi @IBRAHIMDANS , the PR is not minimal sorry so it would be difficult to merge. Also, this is definitely unexpected, the bug is higher up. Why does the resolver sends duplicates in the first place?

PR #2721, opened 2021-06-14
fix bug duplicate item with  usemulti2 

IBRAHIMDANS:
**Quick Fix**

currently, when I call "useMulti2" giving it a **limit** we have elements that are returned several times so I use "**uniq**" from '**lodash**' to solve it


eric-burel:
Closed in favour of https://github.com/VulcanJS/Vulcan/pull/2722

PR #2724, opened 2021-07-15
update material ui imports

EloyID:
CF https://material-ui.com/guides/minimizing-bundle-size/#option-1

The third level imports from Material UI are not accepted in the latest release of v4.
 I have just updated these imports. It works in my local machine, if someone else can verify it works fine


eric-burel:
Not tested directly but sounds good to me thanks Eloy!

